PR #4, opened 2015-02-14
initial draft of log rfc

taotetek:
Peter - here's an initial rfc draft for log for discussion.  Cheers!

### DIFF
--- a/rfc/rfc004-package-log.md
+++ b/rfc/rfc004-package-log.md
@@ -33,11 +33,11 @@ http://peter.bourgon.org/go-kit/#package-log
 *	Log MAY provide some built in formatters such as RFC3164, JSON, etc.
 
 ### Levels
-*	Log SHALL provide a set of defined severity levels.
+*	Log MAY provide a set of defined severity levels that can be used (perhaps as a wrapper).
 *	Log SHALL include severity as a k/v pair and allow setting it through the same mechanism as any other k/v pair
 *	Log MAY provide wrapper types as a convenience for setting the severity level
 *	Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
-*	Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
+*	Log MAY allow tying program actions such as panic to a log level.
 
 ## Implementation
 
--- a/rfc/rfc004-package-log.md
+++ b/rfc/rfc004-package-log.md
@@ -1,6 +1,6 @@
 ---
 RFC: 004
-Author: Peter Bourgon <peter@bourgon.org>
+Author: Peter Bougon <peter@bourgon.org> & Brian Knox <bknox@digitalocean.com>
 Status: Draft
 ---
 
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
 
 ## Scope
 
-To be defined.
+* Log SHALL be drop in compatible with the standard library logger.
+* Log SHALL use key / value pairs for structured data.
+* Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
+* Log SHALL be transport agnostic by accepting and io.Writer as output target.
+* Log SHALL provide a default stdout io.Writer.
+* Log MAY implement io.MultiWriters allowing broadcast of logs over multiple transports.
+* Log MAY provide configurable back pressure handling strategies in the case of blocked io.Writers
+* Log MAY provide some built in transports such as syslog and logstash.
+* Log SHALL be format agnostic by providing an interface for log formatting.
+* Log MAY provide some built in formatters such as RFC3164, JSON, etc.
+* Log SHALL provide a set of defined severity levels.
+* Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
+* Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
 
 ## Implementation
 
-To be defined.
+* The initial implementation should be a minimal feature set focused only on the scope of the RFC.
+* Additional features and niceties may be added as there is demonstrable proof that the features solve real world problems.
--- a/rfc/rfc004-package-log.md
+++ b/rfc/rfc004-package-log.md
@@ -13,23 +13,23 @@ http://peter.bourgon.org/go-kit/#package-log
 ## Scope
 
 * Log SHALL use key / value pairs for structured data.
-** Log MAY preserve type safety when expressing logs in a structured format that supports types
-** Log SHALL provide a way to set a default k/v set per log context
-** Log MAY provide some pre-canned default keys (level, time, etc) for convenience
+  Log MAY preserve type safety when expressing logs in a structured format that supports types
+  Log SHALL provide a way to set a default k/v set per log context
+  Log MAY provide some pre-canned default keys (level, time, etc) for convenience
 * Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
-** Log MAY allow per log call adhoc k/v pairs (see Logrus as an example)
+  Log MAY allow per log call adhoc k/v pairs (see Logrus as an example)
 * Log SHALL be transport agnostic with pluggable transports.
 * Log SHALL provide a default stdout io.Writer.
 * Log MAY implement io.MultiWriters allowing broadcast of logs over multiple transports.
-** Log MAY use channels instead
+  Log MAY use channels instead
 * Log MAY provide configurable back pressure handling strategies in the case of blocked io.Writers
 * Log MAY provide some built in transports such as syslog and logstash.
-** Log MAY use encoding.* above the transport level
+  Log MAY use encoding.* above the transport level
 * Log SHALL be format agnostic by providing an interface for log formatting.
 * Log MAY provide some built in formatters such as RFC3164, JSON, etc.
 * Log SHALL provide a set of defined severity levels.
-** Log SHALL include severity as a k/v pair and allow setting it through the same mechanism as any other k/v pair
-** Log MAY provide wrapper types as a convenience for setting the severity level
+  Log SHALL include severity as a k/v pair and allow setting it through the same mechanism as any other k/v pair
+  Log MAY provide wrapper types as a convenience for setting the severity level
 * Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
 * Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
 
--- a/rfc/rfc004-package-log.md
+++ b/rfc/rfc004-package-log.md
@@ -1,6 +1,6 @@
 ---
 RFC: 004
-Author: Peter Bougon <peter@bourgon.org> & Brian Knox <bknox@digitalocean.com>
+Authors: Brian Knox <bknox@digitalocean.com>
 Status: Draft
 ---
 
@@ -12,17 +12,24 @@ http://peter.bourgon.org/go-kit/#package-log
 
 ## Scope
 
-* Log SHALL be drop in compatible with the standard library logger.
 * Log SHALL use key / value pairs for structured data.
+** Log MAY preserve type safety when expressing logs in a structured format that supports types
+** Log SHALL provide a way to set a default k/v set per log context
+** Log MAY provide some pre-canned default keys (level, time, etc) for convenience
 * Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
+** Log MAY allow per log call adhoc k/v pairs (see Logrus as an example)
 * Log SHALL be transport agnostic by accepting and io.Writer as output target.
 * Log SHALL provide a default stdout io.Writer.
 * Log MAY implement io.MultiWriters allowing broadcast of logs over multiple transports.
+** Log MAY use channels instead
 * Log MAY provide configurable back pressure handling strategies in the case of blocked io.Writers
 * Log MAY provide some built in transports such as syslog and logstash.
+** Log MAY use encoding.* above the transport level
 * Log SHALL be format agnostic by providing an interface for log formatting.
 * Log MAY provide some built in formatters such as RFC3164, JSON, etc.
 * Log SHALL provide a set of defined severity levels.
+** Log SHALL include severity as a k/v pair and allow setting it through the same mechanism as any other k/v pair
+** Log MAY provide wrapper types as a convenience for setting the severity level
 * Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
 * Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
 
--- a/rfc/rfc004-package-log.md
+++ b/rfc/rfc004-package-log.md
@@ -12,28 +12,34 @@ http://peter.bourgon.org/go-kit/#package-log
 
 ## Scope
 
-* Log SHALL use key / value pairs for structured data.
-  Log MAY preserve type safety when expressing logs in a structured format that supports types
-  Log SHALL provide a way to set a default k/v set per log context
-  Log MAY provide some pre-canned default keys (level, time, etc) for convenience
-* Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
-  Log MAY allow per log call adhoc k/v pairs (see Logrus as an example)
-* Log SHALL be transport agnostic with pluggable transports.
-* Log SHALL provide a default stdout io.Writer.
-* Log MAY implement io.MultiWriters allowing broadcast of logs over multiple transports.
-  Log MAY use channels instead
-* Log MAY provide configurable back pressure handling strategies in the case of blocked io.Writers
-* Log MAY provide some built in transports such as syslog and logstash.
-  Log MAY use encoding.* above the transport level
-* Log SHALL be format agnostic by providing an interface for log formatting.
-* Log MAY provide some built in formatters such as RFC3164, JSON, etc.
-* Log SHALL provide a set of defined severity levels.
-  Log SHALL include severity as a k/v pair and allow setting it through the same mechanism as any other k/v pair
-  Log MAY provide wrapper types as a convenience for setting the severity level
-* Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
-* Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
+### Key / Value Pairs
+*	Log SHALL use key / value pairs for structured data.
+*	Log MAY preserve type safety when expressing logs in a structured format that supports types
+*	Log SHALL provide a way to set a default k/v set per log context
+*	Log MAY provide some pre-canned default keys (level, time, etc) for convenience
+*	Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
+*	Log MAY allow per log call adhoc k/v pairs (see Logrus as an example)
+
+### Transport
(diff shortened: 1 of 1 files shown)


rfc/rfc004-package-log.md
```diff
@@ -1,6 +1,6 @@
 ---
 RFC: 004
-Author: Peter Bourgon <peter@bourgon.org>
+Author: Peter Bougon <peter@bourgon.org> & Brian Knox <bknox@digitalocean.com>
```
peterbourgon: You can claim exclusive authorship :)

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
 
 ## Scope
 
-To be defined.
+* Log SHALL be drop in compatible with the standard library logger.
```
peterbourgon: What do you mean by this? I'm not sure it's a hard requirement.

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
...
+* Log SHALL use key / value pairs for structured data.
```
peterbourgon: Agreed. Can you think of a use case for allowing per-log-call key/value pairs? e.g. `log.InfoWith("something happened: %d", count, "specialkey", "specialvalue")`

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
...
+* Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
+* Log SHALL be transport agnostic by accepting and io.Writer as output target.
+* Log SHALL provide a default stdout io.Writer.
+* Log MAY implement io.MultiWriters allowing broadcast of logs over multiple transports.
+* Log MAY provide configurable back pressure handling strategies in the case of blocked io.Writers
```
peterbourgon: Stretch goal :)

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
...
+* Log MAY provide some built in transports such as syslog and logstash.
+* Log SHALL be format agnostic by providing an interface for log formatting.
+* Log MAY provide some built in formatters such as RFC3164, JSON, etc.
+* Log SHALL provide a set of defined severity levels.
```
peterbourgon: I'd like these to be configurable, ideally at the type level. That is, some services will require only INFO + ERROR, and I'd like to be able to implement something that uses the gokit logger and only exposes Info and Error methods.

tsenart:
Yay. First RFC! :-)

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
...
+* Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
+* Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
 
 ## Implementation
 
-To be defined.
\ No newline at end of file
+* The initial implementation should be a minimal feature set focused only on the scope of the RFC.
```
peterbourgon: Yep, I'm very interested in getting something usable for the first cut.

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
 
 ## Scope
 
-To be defined.
+* Log SHALL be drop in compatible with the standard library logger.
+* Log SHALL use key / value pairs for structured data.
+* Log MAY allow multiple stored logging contexts each with a different set of k/v pairs.
+* Log SHALL be transport agnostic by accepting and io.Writer as output target.
+* Log SHALL provide a default stdout io.Writer.
+* Log MAY implement io.MultiWriters allowing broadcast of logs over multiple transports.
```
tsenart: I think there isn't such an interface. The io.MultiWriter function simply returns a Writer.

tsenart:
Is the scope of this package limited to logging for humans? Or could we be thinking of structured "event" logging (e.g. Kafka)?

peterbourgon:
@tsenart I absolutely want `package log` to be a viable choice for structured event logging in the long term.

rfc/rfc004-package-log.md
tsenart: If one thinks of a log entry as a record of structured data (like key value pairs), wouldn't severity be just another attribute of this entry? Why give special treatment to this one attribute?

tsenart:
@peterbourgon If so, I would imagine this package to work with something like encoding.TextMarshaler and encoding.BinaryMarshaler, providing ergonomic layers above these lower level.

rfc/rfc004-package-log.md
tj: We (ab)use ioMultiWriter for this, but it's sort of a hack since writes != messages, it works since there's no partial writes going. Using io.Writers is probably not the best interface, it requires that you're fully aware of _which_ io.Writers are being used, special-cased for messages

rfc/rfc004-package-log.md
rafroehlich2: I see the severity of the entry just being a key in the overall object.

rfc/rfc004-package-log.md
tsenart: I'd love type safety to be preserved in the structured logging front. For the lack of better mechanisms, code generation + go generate could come to the rescue, receiving users' types as input. Just an idea :-)

rfc/rfc004-package-log.md
andyleap: Would it be worth it to use channels instead of writers, and just have log transports give a channel to the log package, and have the log package multiplex out to all the log transport channels?  Then the transport could decide buffer on the channel and such.  Maybe if a channel would block the send, skip that channel?

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,21 @@ http://peter.bourgon.org/go-kit/#package-log
...
+* Log MAY provide configurable back pressure handling strategies in the case of blocked io.Writers
+* Log MAY provide some built in transports such as syslog and logstash.
+* Log SHALL be format agnostic by providing an interface for log formatting.
+* Log MAY provide some built in formatters such as RFC3164, JSON, etc.
+* Log SHALL provide a set of defined severity levels.
+* Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
```
aybabtme: curious for the justification on this one

rfc/rfc004-package-log.md
taotetek: The more I think about it the more I'm not sure it's a hard requirement either.

rfc/rfc004-package-log.md
taotetek: Definitely a use case for per log call k/v pairs in an ad-hoc fashion.  Something like .WithField / .WithFields from logrus would be nice.

rfc/rfc004-package-log.md
taotetek: Agreed.  Not needed in the initial implementation, but something that will probably end up being needed eventually.

rfc/rfc004-package-log.md
aybabtme: In logrus, we found that having the logger be exportable as an `io.Writer`, and thus pluggable in the stdlib `log.Logger` was a good way to remain compatible with things that expect a stdlib logger.  But otherwise, the stdlib logger is very limited for structured logging.

rfc/rfc004-package-log.md
taotetek: * I was thinking along the lines of the convenience methods some logging packages allow.. log.Error, log.Info, etc.  
* In the case of making the log level just another key - transport implementations like rfc3164 syslog would then need a way to map a user defined severity to a syslog level (which is certainly doable).

rfc/rfc004-package-log.md
taotetek: To me, "handling an error" and "logging about an error" are two distinct concerns.  I'm not a fan of things like log.Fatal and log.Panic, as I think terminating execution of a program is something that should be outside of the domain of the logger, which should just care about logging.

I'm open to discussion, just my opinion.

aybabtme: Makes sense to me, although I make use of `log.Fatal` in simple CLI app to exit with a message and non-0 exit code. But that's just lazyness and having it be hooks would work fine too. I agree with your opinion.

rfc/rfc004-package-log.md
aybabtme: the spelling of `Bougon` seems wrong =P

aybabtme:
Would the log package be concerned with basic, default keys such as:

1. `level` for the log level
2. `time` for a timestamp of the log event.
3. Maybe: `id` for a total ordering of the logs within a process (assuming that `time` can clash depending on resolution)
4. Maybe: filename + line #?
5. Maybe: goroutine #? That would have to be a hack given that goroutine IDs aren't exposed by the runtime.

Or would those be provided by the user as regular key-values? My opinion would be to provide some basic keys, like 1 to 3, otherwise the process of entering the values at each log entry would be very cumbersome and lead to inconsistency across and within programs.

taotetek:
@aybabtme I like the idea of some basic default keys - however in practice I often find my idea of what keys should be basic and default don't match someone else's.  

We could have our cake and eat it too though, by providing a way to set basic default keys during logger creation - and providing a set that we think is appropriate for people who want to use them.

We do similar things in rsyslog - e.g., there are standard default templates for various things, that can be overridden if one wishes to.

rfc/rfc004-package-log.md
peterbourgon: @aybabtme you're referring to [this](https://github.com/Sirupsen/logrus/blob/38a6ff9215d472ce953bb9aa37b5327715c0f940/writer.go#L9-16)? I guess it ends up logging with an empty fieldset?

rfc/rfc004-package-log.md
peterbourgon: What do you think of a base logger interface, with only a single method for logging with fields, and then implementing different Warn/Error/etc. method-sets as wrapper types? Just a quick sketch, borrowing heavily from logrus:

```go
func (l *Logger) WithField(k string, v interface{}) *Entry
func (l *Logger) WithFields(Fields) *Entry

func (e *Entry) Printf(format string, args ...interface{})
```

```go
type BasicLogger struct{ *Logger }

func (l *BasicLogger) Printf(format string, args ...interface{}) {
    l.WithField("level", "default").Printf(format, args...)
}

func (l *BasicLogger) Errorf(format string, args ...interface{}) {
    l.WithField("level", "error").Printf(format, args...)
}
```

```go
type MyOrgLogger struct{ *Logger }

func (l *MyOrgLogger) Debugf(format string, args ...interface{}) {
    l.WithField("level", "debug").Printf(format, args...)
}

func (l *MyOrgLogger) Infof(format string, args ...interface{}) {
    l.WithField("level", "info").Printf(format, args...)
}

func (l *MyOrgLogger) Warnf(format string, args ...interface{}) {
    l.WithField("level", "warn").Printf(format, args...)
}

func (l *MyOrgLogger) Errorf(format string, args ...interface{}) {
    l.WithField("level", "error").Printf(format, args...)
}

func (l *MyOrgLogger) Fatalf(format string, args ...interface{}) {
    l.WithField("level", "fatal").Printf(format, args...)
    panic("kaboom")
}
```

peterbourgon:
@tsenart —

> I would imagine this package to work with something like encoding.TextMarshaler and encoding.BinaryMarshaler, providing ergonomic layers above these lower level.

Would wiring the `package log` output to a TeeWriter that split to a {Text,Binary}Marshaler work?

@aybabtme —

> My opinion would be to provide some basic keys, like 1 to 3, otherwise the process of entering the values at each log entry would be very cumbersome and lead to inconsistency across and within programs.

What makes sense to me is that the base logger would provide no fields by default, and that we could ship wrappers that provided different fieldsets. We could even anoint one of those wrappers as Default, if we wanted.

rfc/rfc004-package-log.md
tsenart: +1

rfc/rfc004-package-log.md
aybabtme: correct

tsenart:
I was trying to think about how to decouple the specific Logger logic from
the output serialization format. Given we want a common interface for text
and binary logging. But if the Logger takes objects that know how to
Marshal themselves, then the scope of its responsibilities is greatly
reduced. What would be left out for a Logger to do? :-)

peterbourgon:
> But if the Logger takes objects that know how to Marshal themselves, then the scope of its responsibilities is greatly reduced. What would be left out for a Logger to do? :-)

Just thinking out loud — the [logrus.Logger](http://godoc.org/github.com/Sirupsen/logrus#Logger) emits [Entry](http://godoc.org/github.com/Sirupsen/logrus#Entry) objects, which represent an individual log entry, inherit their parent's fields, can take on their own. If we did something like that, but defined Entry as an interface, we could try to perform an upgrade at serialization time...

```go
// binaryLogger implements kitlog.Logger
type binaryLogger struct { dst io.Writer }

// binaryEntry implements kitlog.Entry
type binaryEntry struct { /* ... */ }

// MarshalBinary allows binaryEntry to also satisfy encoding.BinaryMarshaler
func (e *binaryEntry) MarshalBinary() ([]byte, error) { /* ... */ }

// WithField implements kitlog.Logger by yielding a binaryEntry
func (l *binaryLogger) WithField(k string, v interface{}) Entry {
    return &binaryEntry{ /* ... */ }
}

// Write implements kitlog.Logger...
func (l *binaryLogger) Write(e Entry) error {
    // ...but only works with entries that are BinaryMarshalers
    marshaler, ok := e.(encoding.BinaryMarshaler)
    if !ok {
        return ErrBadEntry
    }
    buf, err := marshaler.MarshalBinary()
    if err != nil {
        return err
    }
    return l.dst.Write(buf)
}
```

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,34 @@ http://peter.bourgon.org/go-kit/#package-log
...
+*	Log MAY use encoding.* above the transport level
+*	Log MAY provide configurable back pressure handling strategies in the case of blocked Writers
+
+### Formats
+*	Log SHALL be format agnostic by providing an interface for log formatting.
+*	Log MAY provide some built in formatters such as RFC3164, JSON, etc.
+
+### Levels
+*	Log SHALL provide a set of defined severity levels.
+*	Log SHALL include severity as a k/v pair and allow setting it through the same mechanism as any other k/v pair
```
peterbourgon: I find these two points contradictory. If severity is just a k/v pair, same as any other, then shouldn't the package be agnostic to which ones are defined?

rfc/rfc004-package-log.md
```diff
@@ -12,8 +12,34 @@ http://peter.bourgon.org/go-kit/#package-log
...
+*	Log MAY provide wrapper types as a convenience for setting the severity level
+*	Log MUST NOT intrinsicly tie severity levels to program actions - e.g., a call to a specific log level should not call a panic.
+*	Log MAY allow tying program actions such as panic to a log level as a configurable option via hooks.
```
peterbourgon: I would strike `via hooks` as I guess that's an implementation detail. It could also be satisfied with wrappers.

rfc/rfc004-package-log.md
taotetek: @peterbourgon it's worded poorly - my thought was including something like a wrapper, with some standard ones ( info, debug, error, etc ) for convenience - but that you could certainly not use the wrapper and just implement your own.

rfc/rfc004-package-log.md
taotetek: fair point, I'll remove the reference to "hooks"

peterbourgon:
Otherwise, looks good! I think this is enough to set the tone of the package and get started. With the above minor changes, I'll merge and we can start looking at code :)

aybabtme:
@peterbourgon : with regards to `logrus.Entry`

One of the most useful feature of the `logrus.Entry` is that it serves as a scope and can be used to `Print` multiple time. We leverage this a lot at `{{.Work}}`:

```go
ll := log.WithFields(log.Fields{
	"server.id":  srv.ID,
	"user.id":    user.ID,
	"region.id":  region.ID,
})

ll.Info("creating a cataclismic event")  // {"server.id":..., "user.id":..., "region.id":..., "msg": ...}
ev := &Event{ /*...*/ }

err := db.Events.Create(ev)
if err != nil {
	ll.WithError(err).Error("failed to start cataclism") // {"server.id":..., "user.id":..., "region.id":..., "err": err, "msg": ...}
	return
}

ll = ll.WithField("event.id", ev.ID) 
// ...
ll.Info("searching for dinosaurs") // {"server.id":..., "user.id":..., "region.id":..., "event.id":..., "msg":...}
ll.Error("some mammal survived!")   // {"server.id":..., "user.id":..., "region.id":..., "event.id":..., "msg":...}
```

In HTTP requests and RPC, we set a `request.id` field of a `log.Entry` at the entry point and pass the `*log.Entry` to every handlers, so that all their logs include the `request.id` field that we can then trace in Logstash/Kibana.

```go
ll := log.WithFields(log.Fields{
	"http.method":    r.Method,
	"http.path":      r.URL.Path,
	"http.query":     r.URL.RawQuery,
	"http.useragent": r.UserAgent(),
	"http.remote_ip": remoteIP,
})
if uuid, ok := dctx.UUIDFromContext(ctx); ok {
	ll = ll.WithField("request.id", uuid)
}

switch r.Method {
case "GET", "HEAD":
	h.serveFile(ctx, ll, rw, r)
case "POST":
	h.saveFile(ctx, ll, rw, r)
	...
}
```

peterbourgon:
@aybabtme ah, so you consider the logger and/or entry a first-order parameter to a lot of (all?) biz-level functions, then... ?

aybabtme:
Yes, depending on the component. We pass it either as a function argument, or sometime attach to a type: say an `http.Handler` that will always log using its given `*log.Entry`.  

```go
type FileHandler struct {
   log *log.Entry
}

func NewFileHandler(l *log.Entry) *FileHandler {
    return &FileHandler{log: l.WithField("handler", "files") }
}

func (f *FileHandler) ServeHTTP(...) {
    f.log.Info("serving something")
}
```

At this point, our usage of it becomes somehow redundant with `context.Context` in certain places, I haven't figured what to do about it.

peterbourgon:
@aybabtme Interesting. Still thinking out loud: a context.Context seems like the right place to hang things like contextual loggers or log entries. If you need to introduce a new log context, that should map to deriving a new context.Context, so you can extract+extend+reinject the logging object as appropriate. Does it make sense to you?

peterbourgon:
We hashed the context stuff out on IRC. Forward! :)

timblair:
@peterbourgon @aybabtme: could you summarise the IRC discussion on contexts here?  Would be good to have to wrap up the comments in this PR.

joho:
:+1: to getting a TL;DR on irc discussions, with timezones it's very hard to keep up.

peterbourgon:
[The IRC log in question](https://gist.github.com/peterbourgon/9aa3b32ecb6b4f78c4b4). Just some discussion about what's appropriate to let leak into method arguments, and alternate strategies.

joho:
Thanks!

tsenart:
+1 for keeping things pure with http://play.golang.org/p/Yqqr_DH8F7

PR #5, opened 2015-02-15
Initial draft of server rfc

mattheath:
Very incomplete, initial draft for a `server` package for discussion.

### DIFF
--- a/rfc/rfc000-package-server.md
+++ b/rfc/rfc000-package-server.md
@@ -37,7 +37,7 @@ Package server is probably the biggest and most important component of the toolk
 *   A server MAY limit the total number of concurrent requests it can serve.
 *   A server MAY impose rate limits on specific clients.
 *   Rate limit behaviour MAY range from minimum request intervals, to time based, or leaky bucket algorithms.
-*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service, using some system such as memcached.
+*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service
 
 ### Backpressure
 
--- a/rfc/rfc000-package-server.md
+++ b/rfc/rfc000-package-server.md
@@ -1,19 +1,91 @@
----
-RFC: 000
-Author: Peter Bourgon <peter@bourgon.org>
-Status: Draft
+# Package Server
+
 ---
 
-# package server
+RFC: 000
+
+Author: Matt Heath <matt@mattheath.com>
+
+Status: Draft
+
+---
 
 ## Motivation
 
-http://peter.bourgon.org/go-kit/#package-server
+[http://peter.bourgon.org/go-kit/#package-server](http://peter.bourgon.org/go-kit/#package-server)
+
+Package server is probably the biggest and most important component of the toolkit. Ideally, we should be able to write our services as implementations of normal, nominal Go interfaces, and delegate integration with the environment to the server package. The package should encode and enforce conventions for server-side concerns, like health checks, system-wide request tracing, connection management, backpressure and throttling, and so on. For each of those topics, it should provide interfaces for different, pluggable strategies. It should integrate with service discovery, and work equally well over multiple transports. Considerable prior art exists in the form of Finagle, Karyon (Netflixs application service library), and likely many more.
 
 ## Scope
 
-To be defined.
+### Endpoints
+
+*   An endpoint is defined as a handler interface which receives a request and returns a response or an error to the client.
+*   A server SHALL have one or more endpoints.
+*   An endpoint SHALL accept a request context and propagate this - allowing access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline.
+*   An endpoint MAY respect request cancellation if or when the request's deadline expires.
+*   A server MAY expose information about its endpoints, to allow integration with additional tools.
+
+### Contexts
+
+*   Requests SHALL be executed within a request [context](https://blog.golang.org/context), which the server will pass through the request chain.
+
+### Throttling
+
+*   A server MAY throttle inbound requests and reject requests from clients based on a number of factors.
+*   A server MAY respond with either an Out Of Capacity error, or a Rate Limit Exceeded error when rejecting requests.
+*   A server MAY limit the total number of concurrent requests it can serve.
+*   A server MAY impose rate limits on specific clients.
+*   Rate limit behaviour MAY range from minimum request intervals, to time based, or leaky bucket algorithms.
+*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service, using some system such as memcached.
+
+### Backpressure
+
+*   A server MAY return back-pressure information to clients, via the transport.
+
+### SLAs
+
+*   A server MAY report its SLA per endpoint to a discovery system, allowing clients to estimate response time.
+
+### Healthchecks
+
+*   A server SHALL accept registratration of healthchecks with a defined interface.
+*   A server MAY register default healthchecks to report the health of built in components of the server.
+*   A server SHALL register an endpoint which can be queried to obtain healthcheck information and status.
+*   A server MAY publish these healthcheck statuses via a pluggable transport.
+
+### Service Discovery
+
+*   A server SHALL register itself with a service discovery mechanism on startup.
+*   A server SHALL attempt to deregister itself with a service discovery mechanism on shutdown.
+*   The discovery mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism itself is beyond the scope of this RFC.
+
+### Request Tracing
+
+*   Requests received by the server which are accompanied with tracing information SHALL respect this information and pass this information onto other sub-requests initiated by clients within the server.
+*   The request tracing mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism is beyond the scope of this RFC.
+
+### Transport
+
+*   A server SHALL receive and respond to requests via a Transport.
+*   The Transport mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism of the transport is beyond the scope of this RFC.
+
+### Codec
+
+*   A server SHALL encode and decode requests and responses via an interchangeable Codec.
+*   A server MAY support multiple encodings, and use the appropriate Codec as indicated by the transport.
+*   A server MAY indicate to the transport the encoding used, allowing clients to easily decode the response.
 
 ## Implementation
 
 To be defined.
+
+## Further Reading
+
+[Your Server as a Function](http://monkey.org/~marius/funsrv.pdf) - Marius Eriksen
+
+[Finagle](https://twitter.github.io/finagle/) - Twitter
+
+[Karyon](https://github.com/Netflix/karyon) - Netflix
+
+[State of the Art in Microservices](https://www.slideshare.net/adriancockcroft/dockercon-state-of-the-art-in-microservices) - Adrian Cockcroft
--- a/rfc/rfc000-package-server.md
+++ b/rfc/rfc000-package-server.md
@@ -23,7 +23,7 @@ Package server is probably the biggest and most important component of the toolk
 *   An endpoint is defined as a handler interface which receives a request and returns a response or an error to the client.
 *   A server SHALL have one or more endpoints.
 *   An endpoint SHALL accept a request context and propagate this - allowing access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline.
-*   An endpoint MAY respect request cancellation if or when the request's deadline expires.
+*   An endpoint SHALL respect request cancellation if or when the request's deadline expires.
 *   A server MAY expose information about its endpoints, to allow integration with additional tools.
 
 ### Contexts
--- a/rfc/rfc000-package-server.md
+++ b/rfc/rfc000-package-server.md
@@ -30,18 +30,14 @@ Package server is probably the biggest and most important component of the toolk
 
 *   Requests SHALL be executed within a request [context](https://blog.golang.org/context), which the server will pass through the request chain.
 
-### Throttling
+### Throttling & Backpressure
 
 *   A server MAY throttle inbound requests and reject requests from clients based on a number of factors.
 *   A server MAY respond with either an Out Of Capacity error, or a Rate Limit Exceeded error when rejecting requests.
 *   A server MAY limit the total number of concurrent requests it can serve.
 *   A server MAY impose rate limits on specific clients.
 *   Rate limit behaviour MAY range from minimum request intervals, to time based, or leaky bucket algorithms.
-*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service
-
-### Backpressure
-
-*   A server MAY return back-pressure information to clients, via the transport.
+*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service.
 
 ### SLAs
 
--- a/rfc/rfc000-package-server.md
+++ b/rfc/rfc000-package-server.md
@@ -49,7 +49,7 @@ Package server is probably the biggest and most important component of the toolk
 
 ### Healthchecks
 
-*   A server SHALL accept registratration of healthchecks with a defined interface.
+*   A server SHALL accept registration of healthchecks with a defined interface.
 *   A server MAY register default healthchecks to report the health of built in components of the server.
 *   A server SHALL register an endpoint which can be queried to obtain healthcheck information and status.
 *   A server MAY publish these healthcheck statuses via a pluggable transport.


rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+*   A server MAY throttle inbound requests and reject requests from clients based on a number of factors.
+*   A server MAY respond with either an Out Of Capacity error, or a Rate Limit Exceeded error when rejecting requests.
+*   A server MAY limit the total number of concurrent requests it can serve.
+*   A server MAY impose rate limits on specific clients.
+*   Rate limit behaviour MAY range from minimum request intervals, to time based, or leaky bucket algorithms.
+*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service, using some system such as memcached.
+
+### Backpressure
+
+*   A server MAY return back-pressure information to clients, via the transport.
```
mreiferson: do you see this as something fundamentally different from throttling - i.e. does it need to be stated separately?

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+*   Requests SHALL be executed within a request [context](https://blog.golang.org/context), which the server will pass through the request chain.
+
+### Throttling
+
+*   A server MAY throttle inbound requests and reject requests from clients based on a number of factors.
+*   A server MAY respond with either an Out Of Capacity error, or a Rate Limit Exceeded error when rejecting requests.
+*   A server MAY limit the total number of concurrent requests it can serve.
+*   A server MAY impose rate limits on specific clients.
+*   Rate limit behaviour MAY range from minimum request intervals, to time based, or leaky bucket algorithms.
+*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service, using some system such as memcached.
```
mreiferson: doesn't seem strictly necessary to add `using some system such as memcached` - I think the point is communicated effectively without it.

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+
+*   A server MAY return back-pressure information to clients, via the transport.
+
+### SLAs
+
+*   A server MAY report its SLA per endpoint to a discovery system, allowing clients to estimate response time.
+
+### Healthchecks
+
+*   A server SHALL accept registratration of healthchecks with a defined interface.
```
mreiferson: `s/registratration/registration`

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
 ## Scope
 
-To be defined.
+### Endpoints
+
+*   An endpoint is defined as a handler interface which receives a request and returns a response or an error to the client.
+*   A server SHALL have one or more endpoints.
+*   An endpoint SHALL accept a request context and propagate this - allowing access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline.
+*   An endpoint MAY respect request cancellation if or when the request's deadline expires.
+*   A server MAY expose information about its endpoints, to allow integration with additional tools.
```
mreiferson: Are you talking about (potentially) registering *endpoints* for discovery (in addition to registering the *service itself*)?

mreiferson:
@mattheath nice job getting this one started :sparkles:

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+[http://peter.bourgon.org/go-kit/#package-server](http://peter.bourgon.org/go-kit/#package-server)
+
+Package server is probably the biggest and most important component of the toolkit. Ideally, we should be able to write our services as implementations of normal, nominal Go interfaces, and delegate integration with the environment to the server package. The package should encode and enforce conventions for server-side concerns, like health checks, system-wide request tracing, connection management, backpressure and throttling, and so on. For each of those topics, it should provide interfaces for different, pluggable strategies. It should integrate with service discovery, and work equally well over multiple transports. Considerable prior art exists in the form of Finagle, Karyon (Netflixs application service library), and likely many more.
 
 ## Scope
 
-To be defined.
+### Endpoints
+
+*   An endpoint is defined as a handler interface which receives a request and returns a response or an error to the client.
```
peterbourgon: Can you expand on this? Do you mean all server endpoints to implement literally the same interface, e.g.

```go
type Handler interface {
    func Serve(Request) (Response, error)
}
```

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+*   A server SHALL have one or more endpoints.
+*   An endpoint SHALL accept a request context and propagate this - allowing access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline.
+*   An endpoint MAY respect request cancellation if or when the request's deadline expires.
```
peterbourgon: I'd like to upgrade this to a SHALL, unless you can think of a reason otherwise :)

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+*   A server MAY expose information about its endpoints, to allow integration with additional tools.
+
+### Contexts
+
+*   Requests SHALL be executed within a request [context](https://blog.golang.org/context), which the server will pass through the request chain.
+
+### Throttling
```
peterbourgon: Do you imagine throttling to be expressed through errors returned by handlers?

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+*   A server MAY return back-pressure information to clients, via the transport.
+
+### SLAs
+
+*   A server MAY report its SLA per endpoint to a discovery system, allowing clients to estimate response time.
+
+### Healthchecks
+
+*   A server SHALL accept registratration of healthchecks with a defined interface.
+*   A server MAY register default healthchecks to report the health of built in components of the server.
```
peterbourgon: Register where?

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+
+### Request Tracing
+
+*   Requests received by the server which are accompanied with tracing information SHALL respect this information and pass this information onto other sub-requests initiated by clients within the server.
+*   The request tracing mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism is beyond the scope of this RFC.
+
+### Transport
+
+*   A server SHALL receive and respond to requests via a Transport.
+*   The Transport mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism of the transport is beyond the scope of this RFC.
```
peterbourgon: True, but this is the interesting bit :) I think we probably need to sketch out the code for server + transport in conjunction with each other.

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,91 @@
...
+
+### Codec
```
peterbourgon: Can you enumerate some example codecs, and how one might be used?

peterbourgon:
:100:

rfc/rfc000-package-server.md
mreiferson: :+1:

rfc/rfc000-package-server.md
mreiferson: My assumption is that this is talking about the ability to register more intelligent domain-specific health checks to be used when the service is probed

rfc/rfc000-package-server.md
mreiferson: Relatedly, how do you envision errors interacting between the server/transport pkgs?  It seems like theres a bit of overlap/potential for leaky interfaces (e.g. HTTP codes)

rfc/rfc000-package-server.md
mattheath: agreed :+1:

rfc/rfc000-package-server.md
peterbourgon: A naïve first-cut? — http://play.golang.org/p/3-9V4w8vo6

rfc/rfc000-package-server.md
mattheath: :+1:

rfc/rfc000-package-server.md
mattheath: Yes. Services themselves need to be registered into a service discovery system, but ideally the endpoints, and potentially the message formats they expect/return would be available and queryable - so tools can be built around this?

rfc/rfc000-package-server.md
mreiferson: Yea, that makes sense when you implement (control) both `Handler` and `Transport` - I was imagining scenarios where that might not be the case (transports provided out-of-box)?

rfc/rfc000-package-server.md
mattheath: Yes, pretty much this ^^

rfc/rfc000-package-server.md
mattheath: Yes, using a database as an example, you may want a connectivity healthcheck so your server identifies when it can't connect to the database. Alternatively something much more complicated based on numbers of failed attempts like a circuit breaker, but domain specific. Will knock up an example :smile:

rfc/rfc000-package-server.md
mattheath: Fair point. They could and probably should be implemented as the same thing. One advantage of the throttling applying backpressure (eg. through errors) would be not needing an additional way to pass this information back to clients.

rfc/rfc000-package-server.md
mattheath: I'm thinking of this as inter-service serialisation formats - I may be using incorrect terminology. So something like JSON, Protobuf or Avro. This doesn't necessarily apply when using something like Thrift though as in that case the codec and transport are coupled together?

rfc/rfc000-package-server.md
mattheath: Yes this does get leaky if we're talking the full range of HTTP codes.

I'd envisage a limited number of actually error types, eg. `BadRequest`, `InternalServerError`, `NotFound` (not necessarily an error?) etc which could be directly translated to a limited set of HTTP error codes, if using HTTP as the transport layer.

rfc/rfc000-package-server.md
mattheath: Indeed. There are a few obvious ones, like `net/rpc` (though [expanded with contexts](https://godoc.org/code.google.com/p/vitess/go/rpcplus)), HTTP & JSON, RabbitMQ & Protobuf etc

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+Package server is probably the biggest and most important component of the toolkit. Ideally, we should be able to write our services as implementations of normal, nominal Go interfaces, and delegate integration with the environment to the server package. The package should encode and enforce conventions for server-side concerns, like health checks, system-wide request tracing, connection management, backpressure and throttling, and so on. For each of those topics, it should provide interfaces for different, pluggable strategies. It should integrate with service discovery, and work equally well over multiple transports. Considerable prior art exists in the form of Finagle, Karyon (Netflixs application service library), and likely many more.
 
 ## Scope
 
-To be defined.
+### Endpoints
+
+*   An endpoint is defined as a handler interface which receives a request and returns a response or an error to the client.
+*   A server SHALL have one or more endpoints.
+*   An endpoint SHALL accept a request context and propagate this - allowing access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline.
```
jonomacd: I'm a bit wooly on how context are passed between services.  For example service A -> B -> C.  It seems clear that the request specific values will be passed down the chain, but what about the deadline? Does each service define it's own when a request comes in or is this passed from the top (service A in this example)?  If it is passed from the top do we have to worry about clock skew, particularly for short deadlines?

peterbourgon:
I'll try to summarize [the chat](https://gist.github.com/peterbourgon/4d817dc1acc08e469052)  between myself, @mattheath, and Vendan last night...

At the moment, we imagine the core to be:

```go
package server

type Request struct { // or interface?
	context.Context
	// more to be defined
}

type Response struct { 
	context.Context
	// more to be defined
}

type Handler interface {
	Serve(*Request) (*Response, error) // RPC semantics
}
```

Concrete server implementations could have one Handler per RPC method, or a single Handler for all methods + demuxing in the method, or anything in between.

A Codec would know how to decode requests and encode responses from something concrete:

```go
type HTTPCodec interface {
	Decode(*http.Request) (Request, error)
	Encode(http.ResponseWriter, Response) error
}
```

- There may be a way to supply a general Codec interface, to be determined.
- It should be possible to provide different implementations of e.g. HTTPCodec, that e.g. encode parameters differently.
- All Codecs would be responsible for transparently and losslessly forwarding request context, probably through the Request object.

Combining a Handler with a specific Codec could yield a specific Transport:

```go
type HTTPServiceTransport struct {
	service Handler 
	codec   HTTPCodec
}

func (t *HTTPServiceTransport) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	req, err := t.codec.Decode(r)
	if err != nil {
		http.Error(w, err.Error(), http.StatusBadRequest)
		return
	}

	// Mapping Handler errors to HTTP codes should be pluggable,
	// so that each service can define its own bindings.
	resp, err := t.service.Serve(req)
	switch err {
	case nil:
		break
	case ErrSomething:
		http.Error(w, err.Error(), http.StatusPaymentRequired)
		return
	default:
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}

	if err = t.codec.Encode(w, resp); err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
}
```

**Importantly**, service authors ideally shouldn't need to implement the Handler interface themselves. Every service should be defined as a pure Go interface:

```go
type FooService interface {
	Bar(value int) error
	Baz(key string) (string, error)
}
```

And we will have some process to generate appropriate stubs:

```go
type Foo_Bar_Adapter struct{ FooService }

func (x *Foo_Bar_Adapter) Serve(req *Request) (*Response, error) {
	err := x.FooService.Bar(req.map["value"].(int))
	return &Response{ // Totally speculating about this bit
		Error: err,
		// ...
	}, nil
}

type Foo_Baz_Adapter struct{ FooService }

func (x *Foo_Baz_Adapter) Serve(req *Request) (*Response, error) {
	val, err := x.FooService.Bar(req.map["key"].(string))
	return &Response{
		Out:   val, // Totally speculating here, too
		Error: err,
		// ...
	}, nil
}
```

Those stubs could be generated at runtime via reflection, or during compile time via `go generate`, or (if necessary for the MVP) manually written by the service author.

rfc/rfc000-package-server.md
peterbourgon: There's a lot baked into this question...

- SLAs like this are typically per-service, so a service with a maximum per-request duration would encode that into its per-request context via the existing Deadline mechanism
- SLAs are also best-effort — it's possible for a service to violate a per-request duration SLA by taking longer — and that's operationally OK
- It's up to parent services (the A in your A->B->C) to decide what to do when downstream dependencies take too long
- If you want to encode semantics like "complete this request in X milliseconds, or return an error" into your service, my intuition is that X should be a parameter on that request method.

That's just my thoughts. Does it help? Does anyone agree or disagree?

mattheath: Largely agree.

1) I'd assume SLAs are typically per endpoint on a service. As an example - given a service managing a resource, a `create` operation may have very different performance characteristics to a `read` operation
2) Ineviatable that services will exceed their SLA, but the client will likely have given up by this point and may ignore the response.
3) Yes, i'd see this logic going into the `client` package
4) We could allow a `timeout` or `deadline` parameter to be passed through the transport from `client` to `server`, and the server could automatically set the deadline in its request context to this duration?

mattheath: Also, using a duration rather than an explicit timestamp for deadline expiration would avoid clock skew at the cost of inaccuracy as the transport time wouldn't be taken into account.

dgryski: Note that net/context has both deadline and timeout: https://godoc.org/golang.org/x/net/context

peterbourgon: 1. Agree, thanks for the clarification. I've updated my comment.
2. :+1:
3. :+1:
4. Calculating and forwarding a max_duration or deadline parameter via a context gets us into a tricky place, where information that materially affects service behavior is passed through an implicit interface. And the behavior we'd be encoding — that requests should timeout after some duration — actually belongs to `package client`. This makes me uncomfortable enough that I'd prefer to leave it out of scope for now.

mreiferson:
@peterbourgon transport/codec example looks good

> Importantly, service authors ideally shouldn't need to implement the Handler interface themselves. Every service should be defined as a pure Go interface:

Can you elaborate on the motivation here?

peterbourgon:
@mreiferson 

> Can you elaborate on the motivation here?

A service, most abstractly, is nothing more than a collection of behaviors, which is perfectly modeled by Go interfaces. Integrating on the interface makes intuitive sense and lines up all the incentives. Authors write their business logic exactly once, totally in the business domain. We have a well-defined surface area to attach transports. We can do well-defined unit and integration testing.

I see enough advantages that I'd like to try to make it work. Of course, it's totally possible I'm overlooking an important detail that makes it intractable.

mreiferson:
I'm a bit skeptical on the need for this level of abstraction (it feels a borderline "magical").  But, as you pointed out, it's something that will shake out when we begin to implement this pkg.

rfc/rfc000-package-server.md
jonomacd: I'm inclined to agree about the deadline parameter.  I puts a lot of onus on people writing endpoints to check for this deadline throughout the code.  I think it is a relatively rare case when an endpoint will actually abide by it.  That said, I suppose it doesn't hurt very much to include it.

It looks to me like the request context is basically just headers when on the wire and those are marshalled in to the net/context type (or some equivalent) when the request is received in a service. Is this the correct way to view it?

peterbourgon:
Point well taken. Fundamentally I want service authors to be able to write and maintain

```go
func Add(i, j int) int {
    return i + j
}
```

and not

```
func Add(req *server.Request) (*server.Response, error) {
    i, err := req.Get("i")
    if err != nil {
        return nil, fmt.Errorf("i: %v", err)
    }

    j, err := req.Get("j")
    if err != nil {
        return nil, fmt.Errorf("j: %v", err)
    }
    
    value := i + j

    return server.NewResponseTo(req).With("value", value), nil
}
```

rfc/rfc000-package-server.md
peterbourgon: >  I suppose it doesn't hurt very much to include it.

I'm afraid it might, actually. I'm afraid it conflates the responsibilities of the caller vs. the callee.

> It looks to me like the request context is basically just headers when on the wire and those are marshalled in to the net/context type (or some equivalent) when the request is received in a service. Is this the correct way to view it?

More or less. For me, the canonical example of what goes into the context is the [request trace ID](https://github.com/peterbourgon/gokit/blob/master/rfc/rfc006-request-tracing.md). Stuff that isn't part of the business-level contract of the service, but is nevertheless important to thread through a request tree.

aybabtme: If clients disconnect after a deadline, it would be good if the transport provided something like http://golang.org/pkg/net/http/#CloseNotifier.

tsenart:
What about trying some code generation approach with a `gokit` tool that could be used along with `go generate`?
```shell
$ gokit service "http:protobuf" adder.go
generating adder_codec.go
generating adder_transport.go
generating adder_service.go
```

Where `adder.go` would contain a service interface definition:
```go
type Adder interface {
    Add(...int64) int64
}
```

peterbourgon:
@tsenart that's exactly what I hope will work!

tsenart:
Would it make sense to create an separate RFC for the gokit tool?

mreiferson:
> Point well taken. Fundamentally, and maybe naïvely, I want service authors to be able to write and maintain ...

Yes, I think it's worth at least exploring what an implementation that supports this would look like :+1:

rafroehlich2:
Will the server implement it's own routing system or will this be plug able as well? This being if the server goes down the road of the minimalistic, abstracted handler interface.

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
 
-# package server
+---
 
 ## Motivation
 
-http://peter.bourgon.org/go-kit/#package-server
+[http://peter.bourgon.org/go-kit/#package-server](http://peter.bourgon.org/go-kit/#package-server)
+
+Package server is probably the biggest and most important component of the toolkit. Ideally, we should be able to write our services as implementations of normal, nominal Go interfaces, and delegate integration with the environment to the server package. The package should encode and enforce conventions for server-side concerns, like health checks, system-wide request tracing, connection management, backpressure and throttling, and so on. For each of those topics, it should provide interfaces for different, pluggable strategies. It should integrate with service discovery, and work equally well over multiple transports. Considerable prior art exists in the form of Finagle, Karyon (Netflixs application service library), and likely many more.
```
dsymonds: A bunch of these concerns aren't unique to servers, but also relevant to clients.

rfc/rfc000-package-server.md
dsymonds: I think deadlines and their close relative cancelations, need to stay with the context, wherever it goes. The reasoning is that deadlines/cancelations are tightly linked to a request (e.g. a client only wants to wait 50ms for something), which is precisely what a context is meant to be for.

It's not necessarily server-specific, but it is definitely relevant to servers *and* clients.

rfc/rfc000-package-server.md
dsymonds: except probably with a context argument too

peterbourgon: I'd expect the context to be part of the Request object, decoded by the codec in the transport layer. Any reason it should be a separate parameter?

rfc/rfc000-package-server.md
peterbourgon: Indeed; I hope I've covered all of the overlapping concerns in [the motivation for package client](http://peter.bourgon.org/go-kit/#package-client). And I expect shared code to be extracted to top-level pacakges. Would you propose a different approach?

peterbourgon:
@rafroehlich2 

> Will the server implement it's own routing system or will this be plug able as well? This being if the server goes down the road of the minimalistic, abstracted handler interface.

I'm not sure what you mean by routing system, can you expand a bit? I don't think the server will implement one, as the binding of service method [stub] to transport endpoint will be explicit and one-to-one.

rfc/rfc000-package-server.md
aybabtme: agreed, context seems to be the best mechanism for that

rfc/rfc000-package-server.md
dsymonds: Well, a bunch are to do with the RPC subsystem (e.g. health checking, tracing, conn management; actually, it seems like *most* of them are transport-related), so the transport package seems to be the right place to put them.

rfc/rfc000-package-server.md
dsymonds: You want to be able to change/replace the context (e.g. wrapping with a tighter timeout, or adding more context keys) without having to make a whole new request. In my experience, `func Serve(context.Context, *Req) (*Res, error)` works well.

rafroehlich2:
@peterbourgon 
What I intended to ask (poorly worded) was if the handler functions would be explicitly registered by the end user or implicitly during the code generation?

peterbourgon:
@rafroehlich2 I guess it's not really set in stone. I imagine that a given transport will have a specific method of registering methods to endpoints "baked-in". But we have to try it and see.

mattheath:
I would see the `Transport` interface having methods to register endpoints, allowing similar behaviour across a number of transports.

As an example:
 * a HTTP transport would be similar behaviour to registering these endpoints directly as HTTP handlers
 * an AMQP transport would receive AMQP deliveries, match these to the correct endpoint, execute this, and then deal with returning the response to the original client

mattheath:
@peterbourgon could we split the discussion of code generation for handlers into a separate RFC? With the assumption that we agree a handler interface which the `gokit` generator would output that matches the interface the server provides?

peterbourgon:
@mattheath Absolutely. Let's settle on this for now, and see how it feels?

```go
type Request struct { /* ... */ }

type Response struct { /* ... */ }

type Server interface {
    Serve(*context.Context, *Request) (*Response, error)
}
```

tsenart:
I'm not convinced that the context.Context can't be easily manipulated
while embedded in a Request.

rfc/rfc000-package-server.md
bensigelman: @dsymonds I disagree that most of these are transport-related. While the transport subsystem should indeed know about connection peer health checks, throttling, etc, the server itself needs to provide the underlying information (about health, queue depth, and so on).

rfc/rfc000-package-server.md
bensigelman: I certainly agree that there needs to be a `Context` somewhere in here. I also agree with @dsymonds that it ought to be its own parameter... Ideally the `Request` type will be pure structural payload data that can be serialized and deserialized without losing information, and I wouldn't want the Context to be part of that serialized format. (I'm imagining that `Request` could (or even should) be a protobuf or thrift type here)

I might be misunderstanding what you mean by `Request` here, though... is that supposed to be a user-defined type? If so, clearer to call it `MyRequest`. If it is in fact a general-purpose Request wrapper that has the app-specific payload data along with other tidbits, then perhaps the Context could go in there, too. Though I would prefer to see that parameter represent the user payload directly if possible.

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
 
 ## Scope
 
-To be defined.
+### Endpoints
+
+*   An endpoint is defined as a handler interface which receives a request and returns a response or an error to the client.
+*   A server SHALL have one or more endpoints.
```
bensigelman: Is there at most one "server" per process? Is there at most one "server" per open port/socket/etc? (Unfortunately the word has been used in so many different contexts over the years that I don't think there is any conventional answer to these questions! Might be good to spell it out, even if the answer is ambiguous.)

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+*   An endpoint SHALL accept a request context and propagate this - allowing access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline.
+*   An endpoint SHALL respect request cancellation if or when the request's deadline expires.
+*   A server MAY expose information about its endpoints, to allow integration with additional tools.
```
bensigelman: there's a lot to be gained be standardizing early on an interface for monitoring introspection (of request counters, latency distributions, error rates, etc).

rfc/rfc000-package-server.md
peterbourgon: As we describe it currently, `Request` is a common, Transport-agnostic type. Kind of a bridge type between the business domain, and whatever's serialized to the wire.

bensigelman: Unrelated to the context stuff: I often wish that we could support schemes other than a single Request/Response pair. I.e., it would be great if we had built-in support from day one for channels; certainly for the responses, but perhaps also for the requests, thus allowing for generic message-passing between clients and servers.

To motivate this, consider a SaaS system that has an embedded client library. Due to the way that various VPCs/NATs are configured, it is of course not generally possible for the SaaS system to open a connection to its clients directly. So, instead, the clients do a "long-poll" pattern to gather information from the server. It would be cleaner in my mind if the client could initiate the connection (required due to the aforementioned networking issues) but, after the connection is initiated, it's just a bidirectional channel between client and server.

(If anyone is swayed by this, we could discuss exactly how the interface would look, how Context would be treated, etc)

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+### Contexts
+
+*   Requests SHALL be executed within a request [context](https://blog.golang.org/context), which the server will pass through the request chain.
+
+### Throttling & Backpressure
+
+*   A server MAY throttle inbound requests and reject requests from clients based on a number of factors.
+*   A server MAY respond with either an Out Of Capacity error, or a Rate Limit Exceeded error when rejecting requests.
+*   A server MAY limit the total number of concurrent requests it can serve.
+*   A server MAY impose rate limits on specific clients.
```
bensigelman: the server might not be the right layer for some of these DOS-ish things (i.e., perhaps that belongs at the transport auth layer?)

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+*   Rate limit behaviour MAY range from minimum request intervals, to time based, or leaky bucket algorithms.
+*   A server MAY implement a pluggable throttle interface, allowing richer implementations - such as an implementation which shares information across instances of the service.
+
+### SLAs
+
+*   A server MAY report its SLA per endpoint to a discovery system, allowing clients to estimate response time.
```
bensigelman: what do you mean by "SLA" here? (I know what the acronym literally means, but can you be more specific about your use of it here? Theoretically an SLA is established up-front, almost as a contractual thing... I think you mean something more like "SLI" (Service Level Indicator), the current state of things, regardless of what was "Agreed" to?)

rfc/rfc000-package-server.md
ShawnMilo: Related to wishing for early support for channels/other request types: What about using the ssh protocol, which has all that built in?

http://blog.gopheracademy.com/go-and-ssh/

rfc/rfc000-package-server.md
bensigelman: oh, also: I could imagine this SLI information showing up in the same interface that allows for general-purpose endpoint introspection and interrogation. Or, more generally, shouldn't this data and that general endpoint data be stored in the same place?

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+### Service Discovery
+
+*   A server SHALL register itself with a service discovery mechanism on startup.
+*   A server SHALL attempt to deregister itself with a service discovery mechanism on shutdown.
+*   The discovery mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism itself is beyond the scope of this RFC.
+
+### Request Tracing
+
+*   Requests received by the server which are accompanied with tracing information SHALL respect this information and pass this information onto other sub-requests initiated by clients within the server.
+*   The request tracing mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism is beyond the scope of this RFC.
```
bensigelman: I'd love to be involved in this when the time comes. I agree it can be deferred as long as you have some of the basics in place (e.g., Context).

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+
+*   A server SHALL accept registration of healthchecks with a defined interface.
+*   A server MAY register default healthchecks to report the health of built in components of the server.
+*   A server SHALL register an endpoint which can be queried to obtain healthcheck information and status.
+*   A server MAY publish these healthcheck statuses via a pluggable transport.
+
+### Service Discovery
+
+*   A server SHALL register itself with a service discovery mechanism on startup.
+*   A server SHALL attempt to deregister itself with a service discovery mechanism on shutdown.
```
bensigelman: I kind of prefer systems that don't even pretend to rely on graceful shutdown (since ungraceful shutdown and/or network partitions are inevitable and it's better to be prepared!).

rfc/rfc000-package-server.md
bensigelman: let's decouple transport from the encoding of the payload data... i.e., let's talk about `net/rpc`, HTTP, and RabbitMQ, but *not* JSON, Protobuf, Thrift, etc.

rfc/rfc000-package-server.md
bensigelman: Thrift does not tightly couple the transport and serialization format (i.e., "codec"). It in fact supports multiple transport mechanisms and multiple serialization formats.

Side note: Thrift makes me a little sad... so many of the pieces are the right basic shape, but there are critical gaps in the implementation for most languages, tons of legit/serious bugs, and – IMO – too many features in the underlying IDL.

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+*   The request tracing mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism is beyond the scope of this RFC.
+
+### Transport
+
+*   A server SHALL receive and respond to requests via a Transport.
+*   The Transport mechanism SHALL be interchangeable, and satisfy a defined interface, however the mechanism of the transport is beyond the scope of this RFC.
+
+### Codec
+
+*   A server SHALL encode and decode requests and responses via an interchangeable Codec.
```
bensigelman: Some very rough proof-of-concept code might be worth pursuing here... I'm concerned that there might be a lot of boilerplate, user-visible casting, or both. Would be nice to see how far we can get without generating code.

rfc/rfc000-package-server.md
```diff
@@ -1,19 +1,87 @@
...
+*   A server MAY support multiple encodings, and use the appropriate Codec as indicated by the transport.
+*   A server MAY indicate to the transport the encoding used, allowing clients to easily decode the response.
```
bensigelman: not a fan of this or the previous item, but I guess that's what "MAY" is for :)

rfc/rfc000-package-server.md
bensigelman: @ShawnMilo I'm no expert on ssh, but wouldn't that involve getting the public keys of all possible clients into the server somehow? (Which, to my mind, seems inconvenient, perhaps prohibitively so) Also, my sense is that the SSH handshake is expensive, though perhaps that would be fine for long-lived channels as you mention.

ShawnMilo: From the article linked above, it seems that the benefits of full-duplex communication and multiple channels might make up for the extra configuration burden. Also, I suspect that some sort of authentication is ideal, assuming that different components of the distributed application could be on the public Internet or on shared hosting. And yes, I was assuming longer-lived connections.

There are ways of reducing the public/private key sharing burden. One that comes to mind is using a single key per application. All components of the system can connect (in either direction) to any other component securely. Given that, at some point, some configuration file and/or embedded info is going to be required to discover and connect to some kind of coordinator service, this seems to be a reasonable approach.

tsenart: @bensigelman, @ShawnMilo: One of the defined goals of the project is RPC as the messaging pattern.

bensigelman: @tsenart that's a bit of a tautology (esp since the project is truly in its infancy and we are literally discussing the defined goals as we speak!), but so be it. RPC is certainly the most important thing to support. But I can think of numerous examples where streaming responses were absolutely essential to proper system behavior. Happy to elaborate if that's necessary.

peterbourgon: @bensigelman I'm on board, but I'm also absolutely terrified of scope creep and analysis paralysis :) Can we settle on RPC for the MVP, and table other messaging patterns for a v2?

bensigelman: @peterbourgon as you know via twitter, scope creep for things like this is terrifying!

I don't really care about supporting other message-passing paradigms from day one. The trouble is that it can be really difficult to retrofit them onto a scheme that assumes RPC for everything, mainly since RPC looks so much like a function call and thus implies all sorts of API decisions that simply don't work for other paradigms. What excites me about Go in this instance is that the channel primitive is such a good fit for a few of the other big/important paradigms.

Of course this is your baby, so do what you will. If it was me, though, I would at least support a mode where the server can return N values as a "streaming" response. This is critically important for microservices that need to return objects that are as large or larger than their own memory footprint. You *need* to break the objects up into pieces, and with streaming responses you are saved a lot of trouble and complexity. (I.e., stick with only one request object to keep the interface closer to RPC)

Another possibility would be to make sure that all of the auth stuff is well-encapsulated in the transport layer and tell people who need streaming (etc) to build it themselves on top of that encapsulated transport layer.

peterbourgon: You're right that it's difficult to retrofit fundamental changes like that, but the good news is I'm not afraid of breaking the API :) This task is gargantuan enough even if we constrain it to a single messaging pattern, and I think we'll learn a lot in the process which should help inform other pattern implementations in a later iteration.

bensigelman: Fair enough! :)

peterbourgon:
There's a quick sketch of some server code over in PR #8 if anyone wants to talk specifics over there.

rfc/rfc000-package-server.md
mattheath: appreciating its more complex, and likely not the default, any reason you wouldn't want to support multiple encodings? eg. json _and_ protobuf?

rfc/rfc000-package-server.md
mattheath: Agree 100%. Merely a suggestion for cleaning up registrations that the service discovery would _need_ to clean up itself, slightly in advance of the service discovery system's timeout. Could change this to MAY?

rfc/rfc000-package-server.md
mattheath: I'd envisage this as an expected deadline that the server thinks its endpoint should respond within, which if it was published would allow clients to better gauge the amount of time they should wait before timing out - eg. an upper 90th which the client treats as a cutoff.

rfc/rfc000-package-server.md
mattheath: Yes, I was assuming one 'server' in a process, with multiple 'endpoints' ie. Handlers/HandleFuncs with some additional semantics. That make sense?

rfc/rfc000-package-server.md
dsymonds: The server would merely plug in that information to the transport subsystem. The client would also do that. The transport package is the common layer.

rfc/rfc000-package-server.md
dsymonds: Each encoding scheme has its wrinkles and warts (I'm looking at you, JSON!). Supporting multiple encoding formats means you multiply those weird bits and end up fighting against them all.

I'd strongly suggest that a transport should be the one to decide what the encoding is. Many transports will pick exactly one (particularly RPC transports).

peterbourgon: I'm into binding a transport+encoding together. I'd like to make sure that one process can expose multiple transport+encodings for the same business logic on separate ports.

rfc/rfc000-package-server.md
bensigelman: @dsymonds I don't see it that way; the data originates in the server, so it should be a part of the server interface. What you're suggesting implies that the transport layer is the "proper" way to determine server health... that seems ill-advised, esp in light of other aspects of this proposal that would have that same sort of health information published to discovery systems and so forth.

I do agree that the transport layer needs access to this data, but that doesn't mean it should be considered the source of it.

rfc/rfc000-package-server.md
bensigelman: Somewhat, yes... it just gets confusing because, e.g., the health for one endpoint may have nothing to do with the health of another (I can give a specific example if this doesn't make sense), so there's something to be said for allowing multiple Servers per process in order to avoid these sorts of shared-fate issues.

rfc/rfc000-package-server.md
bensigelman: I think that'd be advisable, but it sounds like we are in agreement about the spirit of the thing (not relying on graceful shutdown).

I like graceful shutdown for hermetic automated testing, but will argue strenuously against depending on it for anything production-related.

rfc/rfc000-package-server.md
dsymonds: Who's going to use the server health information if not the clients? A server itself doesn't care if it's unhealthy.

I'm not saying the transport package should be the source of the information, but it should "own" it in a sense, since how else will clients find out if they are connected to an unhealthy backend server? I picture the client and server packages as independent things both depending on the transport package (among others).

dsymonds:
FYI, commenting on a GitHub pull request like this is an atrocious way to have this sort of discussion. Can we switch to using the mailing list instead?

rfc/rfc000-package-server.md
bensigelman: @dsymonds whitebox monitoring systems immediately come to mind. I don't see why this is so controversial: the server health is a property of the server, and thus is a natural fit for the server interface. Transport peers are not the only parties that care about server health.

peterbourgon:
> Can we switch to using the mailing list instead?

Yes, I think at this point (almost 100 comments!) we're circling on points that won't necessarily make an impact on the RFC. In the interest of forward progress, I'll put a star next to the particularly controversial points and merge the RFC as Accepted. Thanks very much @mattheath! We can continue the discussion on the mailing list. In particular I'm very interested in the responsibilities of package server vs. package transport.

PR #6, opened 2015-02-16
First cut of package metrics

peterbourgon:
EDIT // Ready to review

### DIFF
--- a/metrics/expvar/expvar.go
+++ b/metrics/expvar/expvar.go
@@ -1,0 +1,113 @@
+// Package expvar implements an expvar backend for package metrics.
+//
+// The current implementation ignores fields. In the future, it would be good
+// to have an implementation that accepted a set of predeclared field names at
+// construction time, and used field values to produce delimiter-separated
+// bucket (key) names. That is,
+//
+//    c := NewFieldedCounter(..., "path", "status")
+//    c.Add(1) // "myprefix_unknown_unknown" += 1
+//    c2 := c.With("path", "foo").With("status": "200")
+//    c2.Add(1) // "myprefix_foo_status" += 1
+//
+// It would also be possible to have an implementation that generated more
+// sophisticated expvar.Values. For example, a Counter could be implemented as
+// a map, representing a tree of key/value pairs whose leaves were the actual
+// expvar.Ints.
+package expvar
+
+import (
+	"expvar"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/peterbourgon/gokit/metrics"
+
+	"github.com/codahale/hdrhistogram"
+)
+
+type counter struct {
+	v *expvar.Int
+}
+
+// NewCounter returns a new Counter backed by an expvar with the given name.
+// Fields are ignored.
+func NewCounter(name string) metrics.Counter {
+	return &counter{expvar.NewInt(name)}
+}
+
+func (c *counter) With(metrics.Field) metrics.Counter { return c }
+
+func (c *counter) Add(delta uint64) { c.v.Add(int64(delta)) }
+
+type gauge struct {
+	v *expvar.Int
+}
+
+// NewGauge returns a new Gauge backed by an expvar with the given name.
+// Fields are ignored.
+func NewGauge(name string) metrics.Gauge {
+	return &gauge{expvar.NewInt(name)}
+}
+
+func (g *gauge) With(metrics.Field) metrics.Gauge { return g }
+
+func (g *gauge) Add(delta int64) { g.v.Add(delta) }
+
+func (g *gauge) Set(value int64) { g.v.Set(value) }
+
+type histogram struct {
+	mu   sync.Mutex
+	hist *hdrhistogram.WindowedHistogram
+
+	name   string
+	gauges map[int]metrics.Gauge
+}
+
+// NewHistogram is taken from http://github.com/codahale/metrics. It returns a
+// windowed HDR histogram which drops data older than five minutes.
+//
+// The histogram exposes metrics for each passed quantile as gauges. Quantiles
+// should be integers in the range 1..99. The gauge names are assigned by
+// using the passed name as a prefix and appending "_pNN" e.g. "_p50".
+func NewHistogram(name string, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.Histogram {
+	gauges := map[int]metrics.Gauge{}
+	for _, quantile := range quantiles {
+		if quantile <= 0 || quantile >= 100 {
+			panic(fmt.Sprintf("invalid quantile %d", quantile))
+		}
+		gauges[quantile] = NewGauge(fmt.Sprintf("%s_p%02d", name, quantile))
+	}
+	h := &histogram{
+		hist:   hdrhistogram.NewWindowed(5, minValue, maxValue, sigfigs),
+		name:   name,
+		gauges: gauges,
+	}
+	go h.rotateLoop(1 * time.Minute)
+	return h
+}
+
+func (h *histogram) With(metrics.Field) metrics.Histogram { return h }
+
+func (h *histogram) Observe(value int64) {
+	h.mu.Lock()
+	err := h.hist.Current.RecordValue(value)
+	h.mu.Unlock()
+
+	if err != nil {
+		panic(err.Error())
+	}
+
+	for q, gauge := range h.gauges {
+		gauge.Set(h.hist.Current.ValueAtQuantile(float64(q)))
+	}
+}
+
+func (h *histogram) rotateLoop(d time.Duration) {
+	for _ = range time.Tick(d) {
+		h.mu.Lock()
+		h.hist.Rotate()
+		h.mu.Unlock()
+	}
+}
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,0 +1,37 @@
+// Package metrics provides an extensible framework to instrument your
+// application. All metrics are safe for concurrent use. Considerable design
+// influence has been taken from https://github.com/codahale/metrics and
+// https://prometheus.io.
+package metrics
+
+// Counter is a monotonically-increasing, unsigned, 64-bit integer used to
+// capture the number of times an event has occurred. By tracking the deltas
+// between measurements of a counter over intervals of time, an aggregation
+// layer can derive rates, acceleration, etc.
+type Counter interface {
+	With(Field) Counter
+	Add(delta uint64)
+}
+
+// Gauge captures instantaneous measurements of something using signed, 64-bit
+// integers. The value does not need to be monotonic.
+type Gauge interface {
+	With(Field) Gauge
+	Set(value int64)
+	Add(delta int64)
+}
+
+// Histogram tracks the distribution of a stream of values (e.g. the number of
+// milliseconds it takes to handle requests). Implementations may choose to
+// add gauges for values at meaningful quantiles.
+type Histogram interface {
+	With(Field) Histogram
+	Observe(value int64)
+}
+
+// Field is a key/value pair associated with an observation for a specific
+// metric. Fields may be ignored by implementations.
+type Field struct {
+	Key   string
+	Value string
+}
--- a/rfc/rfc003-package-metrics.md
+++ b/rfc/rfc003-package-metrics.md
@@ -12,8 +12,55 @@ http://peter.bourgon.org/go-kit/#package-metrics
 
 ## Scope
 
-To be defined.
+- Package metrics SHALL implement Gauges, Counters, and Histograms.
+
+- Each metric type SHALL allow observations with an unlimited number of key/value field pairs,
+  similar to [package log](https://github.com/peterbourgon/gokit/blob/master/rfc/rfc004-package-log.md).
+
+- Counter SHALL be an increment-only counter of type uint64.
+
+- Gauge SHALL be an arbitrarily-settable register of type int64.
+
+- Histogram SHALL collect observations of type int64.
+
+- These interfaces SHALL be the primary and exclusive API for metrics.
+
+- We SHALL provide a variety of implementations of each interface that act as a
+  bridge to different backends: expvar, Graphite, statsd, Prometheus, etc.
+
+- Each metric backend MAY provide additional value-add behaviors. For example,
+  a backend for Histogram may bucket observations according to quantile and
+  calculate additional, derived statistics.
+
 
 ## Implementation
 
-To be defined.
+https://github.com/peterbourgon/gokit/tree/master/metrics
+
+### Gauge
+
+```go
+type Gauge interface {
+	With(Field) Gauge
+	Set(value int64)
+	Add(delta int64)
+}
+```
+
+### Counter
+
+```go
+type Counter interface {
+	With(Field) Counter
+	Add(delta uint64)
+}
+```
+
+### Histogram
+
+```go
+type Histogram interface {
+	With(Field) Histogram
+	Observe(int64)
+}
+```
--- a/README.md
+++ b/README.md
@@ -24,6 +24,28 @@ See [the motivating blog post](http://peter.bourgon.org/go-kit) and, eventually,
 - Having opinions on configuration passing -- flags vs. env vars vs. files vs. ...
 - _more TODO_
 
+## Contributing
+
+At this stage, we're still developing the initial drafts of all of the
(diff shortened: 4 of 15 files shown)


rfc/rfc003-package-metrics.md
```diff
@@ -12,8 +12,54 @@ http://peter.bourgon.org/go-kit/#package-metrics
...
 
-To be defined.
\ No newline at end of file
+### Gauge
+
+```go
+type Gauge interface {
+	With(fields ...Fields) Gauge
+	Set(value float64)
+	Inc(delta float64)
```
tsenart: Why provide an Inc method on a Gauge?

rfc/rfc003-package-metrics.md
```diff
@@ -12,8 +12,54 @@ http://peter.bourgon.org/go-kit/#package-metrics
...
+	Inc(delta uint64)
+}
+```
+
+### Histogram
+
+```go
+type Histogram interface {
+	With(fields ...Fields) Histogram
+	Observe(time.Duration)
```
tsenart: s/Fields/Field/? Also where is Field defined?

tsenart:
Have you considered parametrizing Fields at construction time?

rfc/rfc003-package-metrics.md
peterbourgon: Field is just map[string]interface{}, same as package log. Implementation detail.

peterbourgon:
@tsenart

> Have you considered parametrizing Fields at construction time?

Both are possible with this API, I think...

```go
g0 := NewGauge(...) // all fields at observation time
g1 := NewGauge(...).With("f", "v") // f=v is by default at construction time
```

peterbourgon:
@tsenart 

> Why provide an Inc method on a Gauge?

There are use cases for both Set (absolute) and Inc (delta). Also it copies [expvar.Float](http://golang.org/pkg/expvar/#Float) or Int or whatever.

tsenart:
+1

aybabtme:
lgtm

metrics/histogram_test.go
```diff
@@ -0,0 +1,58 @@
...
+		if y < y0 {
+			z = y * y
+			x = y * (((a[3]*z+a[2])*z+a[1])*z + a[0]) / ((((b[3]*z+b[3])*z+b[1])*z+b[0])*z + 1.0)
+		} else {
+			z = math.Sqrt(-math.Log((1.0 - y) / 2.0))
+			x = (((c[3]*z+c[2])*z+c[1])*z + c[0]) / ((d[1]*z+d[0])*z + 1.0)
+		}
+		x = x - (math.Erf(x)-y)/(2.0/math.SqrtPi*math.Exp(-x*x))
+		x = x - (math.Erf(x)-y)/(2.0/math.SqrtPi*math.Exp(-x*x))
+	}
```
mreiferson: 1. this block of code is amazing (doubly so that it's coming from stackoverflow)
2. bravo @peterbourgon 
3. I have no clue what it does

:fire:

jfbus:
Why vendorize prometheus & other providers ? I would have put each provider in a separate module, to avoid importing unused code.

This :
```
import "github.com/peterbourgon/gokit/metrics"
m := metrics.NewPrometheusCounter("test", "multi_with", "bar", "Bar counter.", []string{"a"}),
```
would be replaced by :
```
import "github.com/peterbourgon/gokit/metrics/prometheus"
m := prometheus.NewCounter("test", "multi_with", "bar", "Bar counter.", []string{"a"}),
```

peterbourgon:
Yes, migrating each provider to its own package could make sense. I'd still need to vendorize e.g. prometheus/client_golang in the prometheus package, though.

rogpeppe:
I thought that vendoring imports was conventional for repos with main packages, but not for library repos. Is that not the case?

peterbourgon:
If I want to provide a consistent user experience, I need to vendor and rewrite import paths. Otherwise I risk making `package metrics` incompatible with users' incidental Prometheus versions. [See here](http://peter.bourgon.org/go-in-production/#dependency-management).

rogpeppe:
From that link:

"If you’re shipping a library, create a vendor subdirectory in the root of your repository. Treat it just like a prefix in package paths; for example, copy the dependency github.com/user/dep to vendor/user/dep. Then, rewrite all of your imports, transitively."

I'm afraid I fundamentally disagree with that approach. If everyone does this, it means we end up with n copies of every package that's used by more than one repo.

For example, if I happen to import both github.com/peterbourgon/gokit/metrics and github.com/codahale/hdrhistogram, then I'm going to get two copies of that hdrhistogram package, and woe betide me if it's a package that has some kind of global registry.

Surely that can't be right!

I believe it's the responsibility of the maintainers of the *roots* of the dependency trees to organise their dependencies to ensure reproducibility by locking to a particular set of dependencies and using those.

This is where stable APIs are important - if the API for a given import path is maintained backward compatibly then it should always be safe to update to the latest published version of a given import path. Of course there is always the possibility that someone breaks something, and that's when you might need to pin a dependency to an earlier version, but ideally all your tests will pass against the latest version of all the packages you depend on.

If they don't, then you'll need to communicate with upstream to get them to fix the issue and/or downstream to ask them to use a different version of the dependency. If you can't do either of those, *then* is the time to fork the package, but hopefully that will be the exception rather than the rule.

All IMHO.

peterbourgon:
@rogpeppe 

> For example, if I happen to import both github.com/peterbourgon/gokit/metrics and github.com/codahale/hdrhistogram, then I'm going to get two copies of that hdrhistogram package, and woe betide me if it's a package that has some kind of global registry. Surely that can't be right!

One of them will have the canonical import path, and the other will have an import path with 'gokit' in it somewhere. But they're distinct packages. As far as I can see, as long as you don't import both in the same import scope (i.e. the same file) then there's no conflict; any package-global stuff, like a metrics.Registry, would be duplicated. If the package had an init() that mutated e.g. http.DefaultRegistry or something, that could yield unpredictable behavior... but that would be bad package behavior indeed.

Did I make an incorrect assumption in there?

> I believe it's the responsibility of the maintainers of the roots of the dependency trees to organise their dependencies to ensure reproducibility by locking to a particular set of dependencies and using those.

I'm sympathetic to this view. Hmm. Sleeping on this.

rogpeppe:
As an example of the kind of things that can go wrong, suppose that one of the gokit dependencies is gopkg.in/mgo.v2 and it provides a method that accepted a value into which it stores a value retrieved from mongoDB.

A caller might well pass a value that implements bson.Setter (from gopkg.in/mgo.v2/bson) and expect that method to be called when the value is unmarshaled from mongodb. But they'd be mistaken, because their bson package is not yours and the method signatures are thus not compatible.

Another issue: these vendored packages will bleed through into the external APIs presented by gokit, forcing other packages to import them directly (I've seen this happen for real), and then making them incompatible with other third party repos that have vendored the same package.

If everyone ends up doing this, we'll end up with bloated binaries and larger working sets. I *really* hope we can avoid ubiquitous duplication of packages.

joho:
Things get gnarly again when your vendored libraries have `init` functions that do the same thing. I'd been bitten once when the postgres library moved repos and I had it imported twice under two paths - in init they both register themselves in a global map somewhere under the same key and whichever loaded second would error out.

Should we be mirroring the approach `database/sql` takes where you import the generic package, then the database specific one as a "side effects import". I know I just outlined a case above where it bit me, but it seems be an established idiom that works most of the time.

adg:
Perhaps an explicit goal of the gokit project should be to depend only on
libraries that have some kind of policy for maintaining API compatibility.
We could even write the (simple) policy and ask the projects we like to
adhere to it.

I'm wary of scope creep for the project, but one of its stated goals is to
address package versioning issues. Codifying and advertising compatibility
expectations seems like a step in the right direction. We could then look
at tooling for checking/enforcing API compatibility.

mreiferson:
Since we're having this discussion here, I lean towards gokit not vendoring deps and that being the responsibility of *applications* that import gokit.

rogpeppe:
> Perhaps an explicit goal of the gokit project should be to depend only on
> libraries that have some kind of policy for maintaining API compatibility.
> We could even write the (simple) policy and ask the projects we like to
> adhere to it.

+1. That was my thought too.

peterbourgon:
I'm all for it. Maybe @rogpeppe or @adg if you felt up to it you could draft the wording in a new RFC, and once approved I'll put it in the appropriate place on the project page?

metrics/expvar/expvar.go
```diff
@@ -0,0 +1,113 @@
...
+			panic(fmt.Sprintf("invalid quantile %d", quantile))
+		}
+		gauges[quantile] = NewGauge(fmt.Sprintf("%s_p%02d", name, quantile))
+	}
+	h := &histogram{
+		hist:   hdrhistogram.NewWindowed(5, minValue, maxValue, sigfigs),
+		name:   name,
+		gauges: gauges,
+	}
+	go h.rotateLoop(1 * time.Minute)
```
mreiferson: What do you think about more explicitly accounting for these various behind-the-scenes goroutines we're launching (and ensuring their clean shutdown)?

Specifically, let's use a `sync.WaitGroup`?

mreiferson: There's probably a larger discussion here around how we would know it's time to cleanly shutdown - the current API doesn't seem to support this...

peterbourgon: If you can think of a clean way to provide a stop mechanism without polluting the metric interfaces, I'm all ears. But it would be an optimization: since metric lifetime is exactly equal to process lifetime, I'm not too concerned about cleaning these up.

adg:
On 26 February 2015 at 04:26, Peter Bourgon <notifications@github.com>
wrote:

> Maybe @rogpeppe <https://github.com/rogpeppe> or @adg
> <https://github.com/adg> if you felt up to it you could draft the wording
> in a new RFC, and once approved I'll put it in the appropriate place on the
> project page?
>

OK, I'll write something up today.

peterbourgon:
I've adopted all the ideas re: dependencies, so if there aren't any objections I'll merge this tomorrow AM CET.

metrics/expvar/expvar.go
mreiferson: Ok, I've been noodling on whether or not there could be reasonable use cases for metrics that don't have the same lifetime as the process but haven't thought of anything.

So I'm :+1: for now.

mreiferson:
you want to squash down these commits a bit or are we throwing repo history to the wind :grinning:?

peterbourgon:
I can squish 'em.

mreiferson:
LGTM

metrics/expvar/expvar.go
```diff
@@ -0,0 +1,113 @@
...
+	gauges map[int]metrics.Gauge
+}
+
+// NewHistogram is taken from http://github.com/codahale/metrics. It returns a
+// windowed HDR histogram which drops data older than five minutes.
+//
+// The histogram exposes metrics for each passed quantile as gauges. Quantiles
+// should be integers in the range 1..99. The gauge names are assigned by
+// using the passed name as a prefix and appending "_pNN" e.g. "_p50".
+func NewHistogram(name string, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.Histogram {
```
bensigelman: the way that the codahale/metrics package does sliding-window histograms is pretty lousy IMO... as long as the underlying histogram (HDR or otherwise) is only added-to over time (certainly the case for cumulative latency distributions), it makes more sense IMO to have the export mechanism – goroutine or otherwise, doesn't matter – take occasional snapshots and just send the "diffs" of the underlying accumulated histogram. This totally solves the precision issues associated with the codahale window segments, and it's more efficient, too. Let me know if that doesn't make sense. This is mostly an implementation detail, but not completely since you can expose more flexibility to the user of this lib about the window width.

metrics/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+// between measurements of a counter over intervals of time, an aggregation
+// layer can derive rates, acceleration, etc.
+type Counter interface {
+	With(Field) Counter
+	Add(delta uint64)
+}
+
+// Gauge captures instantaneous measurements of something using signed, 64-bit
+// integers. The value does not need to be monotonic.
+type Gauge interface {
```
bensigelman: my experience with Gauges suggests that they are better implemented by callbacks than by Set() sorts of interfaces... I actually think there's something to be said for allowing Counters to be [optionally] implemented that way, but it's less important in that case.

metrics/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+// application. All metrics are safe for concurrent use. Considerable design
+// influence has been taken from https://github.com/codahale/metrics and
+// https://prometheus.io.
+package metrics
+
+// Counter is a monotonically-increasing, unsigned, 64-bit integer used to
+// capture the number of times an event has occurred. By tracking the deltas
+// between measurements of a counter over intervals of time, an aggregation
+// layer can derive rates, acceleration, etc.
+type Counter interface {
```
bensigelman: I'm surprised that none of these metric interfaces has a 'name' field or similar...

peterbourgon: The name is really only relevant at export-time, and so can be built in to the construction process. The "name" at observation time is the variable name.

metrics/metrics.go
peterbourgon: It should be straightforward to wrap this API with a callback mechanism. Personally, I've only ever used Gauges via Set.

bensigelman: At google I had precisely the opposite experience! Weird.

I agree that it's not difficult to wrap this, but again I've found that the common case for gauges is that there's already a source of ground truth in your app code somewhere, and you just need to expose it whenever the monitoring heartbeat happens. So a callback is a better fit. Compare this with a counter, of course, where you mainly interact with the counter when the counted thing has just occurred.

Could you humor me with an example or two of your gauge use case?

peterbourgon: Allow me to walk my statement back a bit :) — I actually rarely use gauges. When I do, they typically reflect allocation of resources.

```go
func process() {
    // ...
    something.Allocate(n)
    myGauge.Add(n)
    defer myGauge.Add(-n)
    // ...
}
```

In general the API borrows heavily from [Prometheus](http://godoc.org/github.com/prometheus/client_golang/prometheus#Gauge), and so to adapt their example:

```go
ops := getWork()
opsQueued.Add(len(ops))
for _, op := range ops {
    c <- op
}

// ...

for op := range c {
    opsQueued.Add(-1)
    // ...
}
```

metrics/expvar/expvar.go
peterbourgon: I'd be very interested to see — and merge ;) — expvar Histograms implemented with another backend. Hint, hint ;)

bensigelman: FYI, I would love to do this but just need to find the time. Sorry for the
radio silence... I'll let you know if I come up with the cycles!

peterbourgon: No worries at all! :+1:

metrics/metrics.go
bensigelman: Sorry for the insanely long delay, I am just catching up on things now.

The prometheus example actually is a perfect way to show how much easier a callback-based approach would be! Compare what you have there with something like this:

```
// (Obviously I don't feel strongly about the details here... emphasis on the callback aspect)
type GaugeValue interface {
    With(Field) GaugeValue
    SetInt64(value int64)
    SetFloat64(value float64)  // (per the comment about support temperatures, load avg, etc)
}

...

// Then, to define your gauge:
myGauge := newGaugeWithCallback("ops_queued", valCallback func(gv GaugeValue) {
    // (NOTE: the API user can Set() for multiple With() values within this callback)
    gv.SetInt64(len(ops))
})
```

Basically, the value of a metric only matters when it's being sampled by the "monitoring heartbeat", whatever that is in a given monitoring system. Counters are an odd case, because they are actually stateful... but when you want to monitor a gauge, the actual ground truth of the gauge value almost always lives somewhere else in your application, and it makes more sense to just sample that value during the monitoring heartbeat than to Set() and Set() and Set() again every time it changes. Not only is the latter strategy more expensive, but it also leads to incorrect gauge values when a future maintainer doesn't realize that the gauge code needs to be updated every time the thing it's measuring is updated.

Hope this makes sense. Of course a "stateful" gauge is trivial to implement on top of the callback-based approach.

PR #8, opened 2015-02-18
Server WIP — codec, service, context...

peterbourgon:
Just something to spark some discussions...


server/add_service.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Resource TODO
+type Resource struct{}
+
+// GetA TODO
+func (r Resource) GetA(ctx context.Context, a int) int {
+	return a
+}
+
+// AddRequest can be autogenerated.
```
adg: Are you proposing a mechanism to generate these?

peterbourgon: I think they could be generated, but I haven't yet sussed out the details. I'm hopeful that `go generate` is up to the task.

PR #12, opened 2015-03-05
add RFC007: API Stability Policy

adg:


peterbourgon:
:+1:. Clear, concise, and leaves plenty of room for implementation details. I'll leave it for a couple days for feedback.

tsenart:
LGTM

bketelsen:
LGTM

aybabtme:
:ship:

wongak:
:+1:

rfc/rfc007-api-stability.md
```diff
@@ -0,0 +1,93 @@
...
+
+# API Stability Policy
+
+## Motivation
+
+The gokit project depends on code maintained by others.
+This includes the Go standard library and sub-repositories
+and other external libraries.
+
+The Go project provides stability guarantees, but the other external libraries
```
dsymonds: "The Go project" should probably be replaced with something more specific, like "The Go language and standard library", since the subrepos (x/tools and friends) don't have any stability guarantees (yet).

rfc/rfc007-api-stability.md
```diff
@@ -0,0 +1,93 @@
...
+> We will not change the package's exported API in backward incompatible ways.
+> Future changes to this package will not break dependent code.
+
+### Coverage
+
+The promise of stability includes:
+
+* The package name,
+* Exported type declarations and struct fields (names and types),
+* Exported function and method names, parameters, and return values,
```
dsymonds: Could you clarify this a bit? If there's an exported function that returns an unexported type (as bad a style as that is), is that covered here? To what extent? It probably doesn't have to cover the unexported name, but does the type identity need to remain stable? (e.g. two exported functions, both returning the same unexported type; do they always have to return the same type in the future?)

Maybe this should just rule out functions and methods that operate on unexported types, whether they be arguments or return values.

rfc/rfc007-api-stability.md
```diff
@@ -0,0 +1,93 @@
...
+* Exported constant names (TODO: and values?),
```
dsymonds: I think we concluded that it has to include their values in the standard library, so it'd have to here.

(That reminds me that it'd be nice to generalise the api tool for this kind of purpose; perhaps a "stable" package in the sense of this RFC should check in an api.txt file or something).

rfc/rfc007-api-stability.md
```diff
@@ -0,0 +1,93 @@
...
+Should the author wish to break compatibility by redesigning the API the author
+should create a new package with a new import path. 
+
+### Awareness
+
+This text should be present in a file named STABILITY in the repository root.
+
+### Enforcement
+
+Tooling may be devised to check the stability of an API as a package evolves,
```
dsymonds: Ah, right. Yes, +1 on this.

rfc/rfc007-api-stability.md
```diff
@@ -0,0 +1,93 @@
...
+
+* Struct literals. For the addition of features it may be necessary to add
+  fields to exported structs in the package API. Code that uses unkeyed struct
+  literals (such as pkg.T{3, "x"}) to create values of these types would fail
+  to compile after such a change. However, code that uses keyed literals
+  (pkg.T{A: 3, B: "x"}) will continue to compile after such a change. We will
+  update such data structures in a way that allows keyed struct literals to
+  remain compatible, although unkeyed literals may fail to compile. (There are
+  also more intricate cases involving nested data structures or interfaces, but
+  they have the same resolution.) We therefore recommend that composite
```
dsymonds: worth mentioning that govet already flags unkeyed struct literals?

dsymonds:
LGTM in broad strokes. I've added some comments on parts that might need further clarification.

rfc/rfc007-api-stability.md
adg: On 9 March 2015 at 09:50, David Symonds <notifications@github.com> wrote:

> Could you clarify this a bit? If there's an exported function that returns
> an unexported type (as bad a style as that is), is that covered here? To
> what extent? It probably doesn't have to cover the unexported name, but
> does the type identity need to remain stable? (e.g. two exported functions,
> both returning the same unexported type; do they always have to return the
> same type in the future?)
>
> Maybe this should just rule out functions and methods that operate on
> unexported types, whether they be arguments or return values.
>
Just functions and methods? What about unexported names used in type
declarations?

I'm tempted to say "the type identity must remain the same", which seems
like the loosest guarantee of compatibility we can provide. Maybe @robpike
or @griesemer can offer a suggestion.

Also, do you have an opinion on the TODOs about global variables and
constants? Should changes to them break compatibility? I suppose it depends
how opaque the values are.

dsymonds: Well, this bullet point talks about functions and methods, which is why I was only focused on that. But yes, exported vars and typed consts should probably have to have an exported or builtin type too.

Just ruling out mixing exported and unexported names in function/method/type/var/const declarations might be sufficient; it can be weakened in the future.

adg: On 9 March 2015 at 10:09, David Symonds <notifications@github.com> wrote:

> Well, this bullet point talks about functions and methods, which is why I
> was only focused on that. But yes, exported vars and typed consts should
> probably have to have an exported or builtin type too.
>
> Just ruling out mixing exported and unexported names in
> function/method/type/var/const declarations might be sufficient; it can be
> weakened in the future.
>

This doesn't seem right. It's pretty common to use unexported functions,
types, or values to initialise variables or constants, and I don't think we
should deter people from doing that.

Also the common "options function" pattern typically uses an unexported
type in the function's type signature.

Outright forbidding mixing seems like a misstep.

dsymonds: Yeah, it seems fine to ignore the initialiser value for vars and consts.

The options pattern doesn't need to use a wholly unexported type. It should normally return something exported that may have unexported methods, since you want programs to be able to, say, pass around a `...pkg.WhateverOption` from outside the package.

peterbourgon:
Any outstanding objections on this?

adg:
Let me take one more pass over it.

apsanz:
I am still trying to get my mind around the final product and workflow of using gokit but is it important to say if binary compatibility is not required? 

Also, is there any limitations on using the latest features in a new version of Go? Is compatibility only guaranteed if compiling with the latest version of Go? Is so that should be added to the RFC.

peterbourgon:
@apsanz neither are pressing issues, IMO. Binary compatibility isn't a problem in the Go universe because we don't distribute binary artifacts, and thanks to the Go authors' strict compatibility promises on point updates everyone typically (and safely) assumes latest stable Go release. I'm happy to add the latter point to the README and/or RFC.

apsanz:
Not really concerned about binary distribution, more about not needing to stop all everything and rebuild at the same time. Would we need to rebuild and deploy the server with every go release? Wouldn't this require downtime? Or would every service use it own version of the server package and still communicate between each other correctly?

peterbourgon:
Ah, so under [Non-goals](https://github.com/peterbourgon/gokit#non-goals) we have "Having opinions about deployment" which I think this falls under. Also providing context is

> Users who import gokit into their package main are responsible to organize and maintain all of their dependencies to ensure code compatibility and build reproducibility.

PR #17, opened 2015-03-09
add floating point gauge

daniel-garcia:
The current spec only contains an integer based gauge. Many metrics of interest are sometimes expressed as floating point numbers; load average being an example. This proposal extends the metric interface to include a floating point gauge.

I have not extended the statsd interface because I'm unfamiliar with with its protocol.

### DIFF
--- a/metrics/expvar/expvar.go
+++ b/metrics/expvar/expvar.go
@@ -57,6 +57,22 @@ func (g *gauge) Add(delta int64) { g.v.Add(delta) }
 
 func (g *gauge) Set(value int64) { g.v.Set(value) }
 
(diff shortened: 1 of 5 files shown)


peterbourgon:
Thanks for the PR. If we change package metrics to support a float64 Gauge type, I think it would replace the existing Gauge, rather than complement it. That would have knock-on consequences in (at least) the expvar Histogram that would need to be worked through.

I'll probably attack this in the next week, if you can stand to wait. If you want to amend your PR to change the existing Gauge type to work with float64s, please feel free.

peterbourgon:
That was straightforward. Please see #18

PR #18, opened 2015-03-10
Change metrics.Gauge data type to float64

peterbourgon:
Does what it says on the tin :)

### DIFF
--- a/metrics/expvar/expvar.go
+++ b/metrics/expvar/expvar.go
@@ -42,20 +42,20 @@ func (c *counter) With(metrics.Field) metrics.Counter { return c }
 func (c *counter) Add(delta uint64) { c.v.Add(int64(delta)) }
 
 type gauge struct {
-	v *expvar.Int
+	v *expvar.Float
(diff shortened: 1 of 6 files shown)


daniel-garcia:
:+1:

metrics/expvar/expvar.go
```diff
@@ -100,7 +100,7 @@ func (h *histogram) Observe(value int64) {
 	}
 
 	for q, gauge := range h.gauges {
-		gauge.Set(h.hist.Current.ValueAtQuantile(float64(q)))
+		gauge.Set(float64(h.hist.Current.ValueAtQuantile(float64(q))))
```
tsenart: The inner conversion of `q` to `float64` becomes unnecessary.

tsenart: Oh, never mind. You kept the key type in the `gauges` map.

mreiferson:
remember when fixed point was cool?

(also LGTM)

PR #19, opened 2015-03-12
metrics: add callback-style constructors for Gauges

peterbourgon:
In package metrics, we define the interfaces for use by the producer (client app) and leave the details of exposition (to statsd, to Prometheus, etc.) to the specific implementation.

Some implementations will statelessly send every observation directly to the aggregation layer (e.g. statsd); some will statefully collect observations and make them available for scrape heartbeats (e.g. Prometheus).

The semantics that @bensigelman wants [here](https://github.com/peterbourgon/gokit/pull/6/files#r26010599) are actually coupled to implementations that _have_ monitoring heartbeats. We want to support those implementations without necessarily relying on them.

This PR introduces new callback-based Gauge constructors for app code that's built that way. Since the callback is the sole authority for the value of the Gauge, those constructors simply Register or Publish the metric, and return nothing. That's fine: callbacks can be shared.

We also include an adapter for statsd, emitting callback-based Gauges to a statsd aggregator on some interval.

Eager for comments.

### DIFF
--- a/metrics/prometheus/prometheus.go
+++ b/metrics/prometheus/prometheus.go
@@ -111,6 +111,29 @@ func (g prometheusGauge) Add(delta float64) {
 	g.GaugeVec.With(prometheus.Labels(g.Pairs)).Add(delta)
 }
 
+// RegisterCallbackGauge registers a Gauge with Prometheus whose value is
+// determined at collect time by the passed callback function. The callback
+// determines the value, and fields are ignored, so RegisterCallbackGauge
+// returns nothing.
+func RegisterCallbackGauge(namespace, subsystem, name, help string, callback func() float64) {
+	RegisterCallbackGaugeWithLabels(namespace, subsystem, name, help, prometheus.Labels{}, callback)
+}
+
+// RegisterCallbackGaugeWithLabels is the same as RegisterCallbackGauge, but
+// attaches a set of const label pairs to the metric.
+func RegisterCallbackGaugeWithLabels(namespace, subsystem, name, help string, constLabels prometheus.Labels, callback func() float64) {
+	prometheus.MustRegister(prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Namespace:   namespace,
+			Subsystem:   subsystem,
+			Name:        name,
+			Help:        help,
+			ConstLabels: constLabels,
+		},
+		callback,
+	))
+}
+
 type prometheusHistogram struct {
 	*prometheus.SummaryVec
 	Pairs map[string]string
--- a/metrics/expvar/expvar.go
+++ b/metrics/expvar/expvar.go
@@ -19,6 +19,7 @@ package expvar
 import (
 	"expvar"
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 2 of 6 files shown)


metrics/prometheus/prometheus.go
```diff
@@ -111,6 +111,29 @@ func (g prometheusGauge) Add(delta float64) {
...
+// determined at collect time by the passed callback function. The callback
+// determines the value, and fields are ignored, so RegisterCallbackGauge
+// returns nothing.
+func RegisterCallbackGauge(namespace, subsystem, name, help string, callback func() float64) {
+	RegisterCallbackGaugeWithLabels(namespace, subsystem, name, help, prometheus.Labels{}, callback)
+}
+
+// RegisterCallbackGaugeWithLabels is the same as RegisterCallbackGauge, but
+// attaches a set of const label pairs to the metric.
+func RegisterCallbackGaugeWithLabels(namespace, subsystem, name, help string, constLabels prometheus.Labels, callback func() float64) {
```
bensigelman: 1) thanks for sending out this PR!

2) I would suggest having the callback do the label->value mapping... more efficient that way, and also easier to hold a read lock (or similar) while reading values out of a big map.

peterbourgon: If I'm understanding you correctly, I don't think that's compatible with [the Prometheus GaugeFunc API](http://godoc.org/github.com/prometheus/client_golang/prometheus#GaugeFunc), which effectively fixes the label dimensions at construction time.

If that's not what you mean, can you give an example?

bensigelman: That's correct. The prometheus API is ill-suited to the situations that multi-dimensional callback-based metrics are used (or at least the situations where they would have been used at Google!). The API they have is a hassle if the label values aren't known in advance (common in practice -- happy to provide examples if you want), and a downright dealbreaker if it's important to export the values atomically.

peterbourgon: I'll point the Prometheus guys to this discussion. But I guess it's out-of-scope for the PR.

bensigelman: I'm happy to illustrate a way to get the prometheus stuff to coexist with an api that provides labels at heartbeat-time if you want... but no pressure if you want to move on!

juliusv: @bensigelman Just to clarify, with the API, label *values* don't need to be known in advance, but label *names* do (and @brian-brazil and some others have argued that you should always have known label *names* on a given metric - examples to the contrary are welcome). Or are you encountering a case where the API also doesn't deal well with label values not being known during registration time?

The `constLabels` in the line above are just a concept in Prometheus (rarely used!) for setting labels which always have constant values (like the process revision or something like that).

bensigelman: Oh, @juliusv, that's very good to know. I agree that knowing the names is a great (important?) thing to know in advance, so I rescind part of my critique here.

But now the obvious follow-up question: how does `callback` set label name+value pairs?

juliusv: Actually, in this specific example, a `GaugeFunc` is a gauge without any variable labels, totally one-dimensional. For doing callback-based multidimensional metrics with label values determined at callback time, implement the `prometheus.Collector` interface.

juliusv: And just for completeness, a `GaugeVec` (http://godoc.org/github.com/prometheus/client_golang/prometheus#GaugeVec) is a multi-dimensional gauge metric (allowing arbitrary label values when setting the gauge value), but it's not callback-based.

juliusv: @bensigelman Ah sorry, didn't see that you already replied. The callback (`Collect` method) returns a bunch of [prometheus.Metric](http://godoc.org/github.com/prometheus/client_golang/prometheus#Metric) objects which in these kinds of collector callbacks are frequently constructed via the `MustNewConstMetric()` function, like in this example collector: https://github.com/prometheus/client_golang/blob/master/prometheus/expvar.go#L76

So the label values are returned as a string slice whose order has to match the label names that were specified at registration time.

PR #20, opened 2015-03-18
Add go-metrics to related projects

grobie:
This seems trying to solve the same goal of `package metrics`.


juliusv:
:+1: And it supports Prometheus ;)

PR #23, opened 2015-04-15
Improve package log documentation.

ChrisHines:
My take on improving the docs on package log's core pieces with style cues taken from Go's standard library.


peterbourgon:
I like it! Thanks!

Issue #24, opened 2015-04-15
Document concurrency properties of package log.


ChrisHines:
I propose that gokit/log.Logger should specify that implementations must be safe for concurrent use. Logger clients should not need to explicitly coordinate with each other.

peterbourgon:
Totally agree. I'll put that wording into the package documentation.

PR #25, opened 2015-04-15
log: notes and tests re: concurrent access

peterbourgon:
Addresses #24.

@ChrisHines anything else you'd put in here?


ChrisHines:
LGTM.

PR #27, opened 2015-04-16
Fix With to return a concurrent safe Logger.

ChrisHines:
The current version of `log.With` does not return a Logger safe for concurrent use if `keyvals` has excess capacity. This PR provides a test for that bug and a fix for `With`.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -24,7 +24,12 @@ func With(logger Logger, keyvals ...interface{}) Logger {
 		return w.With(keyvals...)
 	}
 	return LoggerFunc(func(kvs ...interface{}) error {
-		return logger.Log(append(keyvals, kvs...)...)
+		// Limiting the capacity of the first argument to append ensures that
+		// a new backing array is created if the slice must grow. Using the
+		// extra capacity without copying risks a data race that would violate
+		// the Logger interface contract.
+		n := len(keyvals)
+		return logger.Log(append(keyvals[:n:n], kvs...)...)
 	})
 }
 
--- a/log/log_test.go
+++ b/log/log_test.go
@@ -61,19 +61,20 @@ func TestWithConcurrent(t *testing.T) {
 	// can bucket the event counts.
 	var wg sync.WaitGroup
 	wg.Add(goroutines)
+	const n = 10000
 	for i := 0; i < goroutines; i++ {
 		go func(idx int) {
 			defer wg.Done()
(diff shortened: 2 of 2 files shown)


log/log_test.go
```diff
@@ -33,3 +34,46 @@ type mylogger struct{ withs int }
...
+	l := log.With(logger, make([]interface{}, 0, 2)...)
+
+	// Start logging concurrently. Each goroutine logs its id so the logger
+	// can bucket the event counts.
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for i := 0; i < goroutines; i++ {
+		go func(idx int) {
+			defer wg.Done()
+			for j := 0; j < 10000; j++ {
```
peterbourgon: Can we parameterize this?

```go
const n = 10000
for i := 0; i < goroutines; i++ {
    go func(idx int) {
        defer wg.Done()
        for j := 0; j < n; j++ {
            l.Log("goroutineIdx", idx)
        }
    }(i)
}
```

Then,

```go
for bucket, have := range counts {
    if want := n; want != have {
        t.Errorf("bucket %d: want %d, have %d", bucket, want, have)
    }
}
```

peterbourgon:
Thanks very much!

log/log.go
```diff
@@ -24,7 +24,12 @@ func With(logger Logger, keyvals ...interface{}) Logger {
 		return w.With(keyvals...)
 	}
 	return LoggerFunc(func(kvs ...interface{}) error {
-		return logger.Log(append(keyvals, kvs...)...)
+		// Limiting the capacity of the first argument to append ensures that
+		// a new backing array is created if the slice must grow. Using the
+		// extra capacity without copying risks a data race that would violate
+		// the Logger interface contract.
+		n := len(keyvals)
+		return logger.Log(append(keyvals[:n:n], kvs...)...)
```
tsenart: Is the order of keyvals relevant? If not, would it be simpler to append(kvs, keyvals...)?

ChrisHines:
@peterbourgon 

>Can we parameterize this?

Done.

@tsenart 

>Is the order of keyvals relevant? If not, would it be simpler to append(kvs, keyvals...)?

That doesn't fix the problem. kvs could just as easily have excess capacity.

tsenart:
@ChrisHines: You're right. I wasn't seeing the problem clearly.

PR #31, opened 2015-04-20
log: Reuse code from withLogger in With

tsenart:
A small code reuse change set following up the work from @ChrisHines in #28.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -20,18 +20,11 @@ type Logger interface {
 // If logger implements the Wither interface, the result of
 // logger.With(keyvals...) is returned.
 func With(logger Logger, keyvals ...interface{}) Logger {
-	if w, ok := logger.(Wither); ok {
-		return w.With(keyvals...)
+	w, ok := logger.(Wither)
+	if !ok {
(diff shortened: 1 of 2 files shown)


ChrisHines:
LGTM.

Code readability is improved by centralizing the thread safety code and the explanatory comment.

Also, this change subtly improves correctness—even if that was not intended—by copying the backing array of the `keyvals` passed to `log.With`. Although I don't expect this to come up very often, the following code works after this change but would have produced unexpected results with the old code.

```go
keyvals := []interface{}{"key", "v1"}
logger = log.With(logger, keyvals...)
keyvals[1] = "v2"
logger.Log(keyvals...)
```

This introduces an additional allocation when creating the first `withLogger`, but the result is correct in the general case.

Can we add a test case for the above scenario as part of this PR?

tsenart:
@ChrisHines: Added the test.

ChrisHines:
LGTM (in case it wasn't obvious)

peterbourgon:
> +12 -13

:sunglasses:

PR #32, opened 2015-04-22
Support dynamic values in log contexts.

ChrisHines:
#### Goal

It is common to include a time-stamp and call-site or stack trace information in every log event in diagnostic logs. Package log should make it easy to add these fields when appropriate.

#### The Problem

Time-stamp and call-site values are dynamic, meaning the logged value changes with each call to `Log`. At the same time we usually want to make a global choice about whether to include them in our logs or not. These fields rapidly lose their value if they are not consistently available and having to specify them each time we call `Log` would be both tedious and error prone.

We could provide simple wrappers for `Logger` that injects the additional fields. This works nicely for time-stamps:

```go
func NewTimestampLogger(logger Logger, key interface{}) Logger {
    return LoggerFunc(func(keyvals ...interface{}) error {
        return logger.Log(append(keyvals, key, time.Now())
    })
}
```

But when we try the same for call-sites the details get ugly:

```go
func NewCallerLogger(logger Logger, key interface{}) Logger {
    return LoggerFunc(func(keyvals ...interface{}) error {
        // how many stack frames should we skip?
        _, file, line, ok := runtime.Caller(?skip?)
        return logger.Log(append(keyvals, key, fmt.Sprint(file, ":", line))
    })
}
```

Finding the call-site for `Log` depends on knowing how many stack frames to ascend? But that can change depending on the order we wrap the base `Logger`. Even if we are careful to always call `NewCallerLogger` last so that it is at the top of our logging stack it will likely get wrapped later by calls to `With`.

`NewCallerLogger` could implement the `Wither` interface to avoid getting wrapped by a `withLogger`. But this approach ends up essentially re-implementing `withLogger`.

#### The Solution

This PR recognizes that the above approaches are essentially specialized versions of `With`. They create a new logging context with a key and value. The essential difference is that the value is not static.

The proposed solution takes inspiration from package database/sql/driver. I propose a new `Valuer` interface similar to [`driver.Valuer`](http://golang.org/pkg/database/sql/driver/#Valuer). I also add a `BindValues` function to scan the values in a keyvals slice for `Valuers` and replace them with the result of their `Value` method. I modify `withLogger` to call `BindValues` when logging.

The result is an *extensible* mechanism for storing dynamic values in a logging context. Implementing `Timestamp` and `Caller` as commonly needed `Valuers` is trivial. Applications or other packages could use this capability to log other types of dynamic data. It seems like there could be some an interesting composition of this capability with package metrics for example.

### DIFF
--- a/log/value.go
+++ b/log/value.go
@@ -1,0 +1,68 @@
+package log
+
+import (
+	"time"
+
+	"gopkg.in/stack.v1"
+)
+
+// BindValues returns a slice with all value elements (odd indexes) that
+// implement Valuer replaced with the result of calling their Value method. If
+// no value elements implement Valuer, the original slice is returned.
+func BindValues(keyvals ...interface{}) []interface{} {
+	if !containsValuer(keyvals) {
+		return keyvals
+	}
+
+	bound := make([]interface{}, len(keyvals))
+	copy(bound, keyvals)
+	for i := 1; i < len(bound); i += 2 {
+		if v, ok := bound[i].(Valuer); ok {
+			bound[i] = v.Value()
+		}
+	}
+
+	return bound
+}
+
+func containsValuer(keyvals []interface{}) bool {
+	for i := 1; i < len(keyvals); i += 2 {
+		if _, ok := keyvals[i].(Valuer); ok {
+			return true
+		}
+	}
+	return false
+}
+
+// Value is the return type of the Value method of the Valuer interface.
+type Value interface{}
+
+// A Valuer is able to convert itself into log Value.
+//
+// A Valuer passed to With stores a dynamic value which is reevaluated on each
+// log event.
+type Valuer interface {
+	// Value returns a log Value instead of an interface{} to avoid
+	// inadvertently matching types from packages not intended for use with
+	// the gokit/log package.
+	Value() Value
+}
+
+type timeStamp struct{}
+
+func (*timeStamp) Value() Value {
+	return time.Now()
+}
+
+// Timestamp is a Valuer that returns the result of time.Now() from its Value method.
+var Timestamp = &timeStamp{}
+
+type caller struct{}
+
+func (*caller) Value() Value {
+	return stack.Caller(3)
+}
+
+// Caller is a Valuer that returns a "gopkg.in/stack.v1".Call from its Value
+// method.
+var Caller = &caller{}
--- a/log/value.go
+++ b/log/value.go
@@ -10,9 +10,9 @@ import (
 // dynamic value which is re-evaluated with each log event.
 type Valuer func() interface{}
 
-// BindValues returns a slice with all value elements (odd indexes) that
-// implement Valuer replaced with the result of calling their Value method. If
-// no value elements implement Valuer, the original slice is returned.
+// BindValues returns a slice with all value elements (odd indexes) containing
+// a Valuer replaced by their generated value. If no Valuers are found, the
+// original slice is returned.
 func BindValues(keyvals ...interface{}) []interface{} {
 	if !containsValuer(keyvals) {
 		return keyvals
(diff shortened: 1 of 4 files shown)


log/value.go
```diff
@@ -0,0 +1,68 @@
...
+	"time"
+
+	"gopkg.in/stack.v1"
+)
+
+// BindValues returns a slice with all value elements (odd indexes) that
+// implement Valuer replaced with the result of calling their Value method. If
+// no value elements implement Valuer, the original slice is returned.
+func BindValues(keyvals ...interface{}) []interface{} {
+	if !containsValuer(keyvals) {
```
tsenart: Is this a fast-path to avoid the allocation of the `bound` slice?

log/value.go
```diff
@@ -0,0 +1,68 @@
...
+	bound := make([]interface{}, len(keyvals))
+	copy(bound, keyvals)
+	for i := 1; i < len(bound); i += 2 {
+		if v, ok := bound[i].(Valuer); ok {
+			bound[i] = v.Value()
+		}
+	}
+
+	return bound
+}
```
tsenart: What do you think of this?
```go
bound := make([]interface{}, len(keyvals))
for i, val := range keyvals {
    if v, ok := val.(Valuer); ok {
        bound[i] = v.Value()
    } else {
        bound[i] = val
    }
}
```

log/value.go
```diff
@@ -0,0 +1,68 @@
...
+// A Valuer passed to With stores a dynamic value which is reevaluated on each
+// log event.
+type Valuer interface {
+	// Value returns a log Value instead of an interface{} to avoid
+	// inadvertently matching types from packages not intended for use with
+	// the gokit/log package.
+	Value() Value
+}
+
+type timeStamp struct{}
```
tsenart: Could we have `ValuerFunc`s? I think that gives users more flexibility.
```go
var (
	// Timestamp is a Valuer that returns the result of time.Now() from its Value method.
	Timestamp = ValuerFunc(func() Value { return time.Now() })
	// Caller is a Valuer that returns a "gopkg.in/stack.v1".Call from its Value
	// method
	Caller = ValuerFunc(func() Value { return stack.Caller(3) })
)

// ValuerFunc is an adapter func type that implements the Valuer interface.
type ValuerFunc func() Value

func (f ValuerFunc) Value() Value { return f() }
```

tsenart: For testing purposes, injecting the time function could be a good idea.
```go
// Timestamp returns a Valuer that returns the time returned by t from its Value method.
func Timestamp(t func() time.Now) Valuer {
  return ValuerFunc(func() Value { return t() })
}
// logger = log.With(logger, "ts", log.Timestamp(time.Now))
```

log/value.go
ChrisHines: Yes.

log/value.go
ChrisHines: That will also inspect keys for Valuers, which wasn't the intention (but open for debate if you think we should do that).

Also, I tried something like that during development. If memory serves is was a little slower, perhaps due to the extra bounds checking in the else clause.

log/value.go
ChrisHines: I don't have a strong opinion about `ValuerFunc`.

For testing purposes would it be sufficient to just assign a new value to log.Timestamp?

tsenart: Mutating global state has its drawbacks. For instance, running parallel
tests that need Timestamp would be harder.

On Friday, April 24, 2015, Chris Hines <notifications@github.com> wrote:

> In log/value.go
> <https://github.com/peterbourgon/gokit/pull/32#discussion_r29016380>:
>
> > +
> > +// Value is the return type of the Value method of the Valuer interface.
> > +type Value interface{}
> > +
> > +// A Valuer is able to convert itself into log Value.
> > +//
> > +// A Valuer passed to With stores a dynamic value which is reevaluated on each
> > +// log event.
> > +type Valuer interface {
> > +	// Value returns a log Value instead of an interface{} to avoid
> > +	// inadvertently matching types from packages not intended for use with
> > +	// the gokit/log package.
> > +	Value() Value
> > +}
> > +
> > +type timeStamp struct{}
>
> I don't have a strong opinion about ValuerFunc.
>
> For testing purposes would it be sufficient to just assign a new value to
> log.Timestamp?
>
> —
> <https://github.com/peterbourgon/gokit/pull/32/files#r29016380>.
>


-- 
Sent from Gmail Mobile

peterbourgon:
Hey! Thanks a bunch. I have some specific comments that might be best expressed in the form of another PR ;) Shall I just make those on your fork, @ChrisHines?

ChrisHines:
@peterbourgon Yes, go ahead with a PR on my fork.

peterbourgon:
[Here you go](https://github.com/ChrisHines/gokit/pull/1) — these PRs be crazy.

ChrisHines:
Updated with results of the PR at ChrisHines/gokit#1.

@tsenart One thing still puzzles me, why did you define log.Timestamp like this:
```go
Timestamp = func(t func() time.Time) Valuer {
	return func() interface{} { return t() }
}
```
Instead of like this:
```go
func Timestamp(t func() time.Time) Valuer {
	return func() interface{} { return t() }
}
```
I don't see the scenario where we would want to change the Timestamp function. Isn't accepting a function argument flexible enough?

tsenart:
I was just trying to squeeze lines of code to convince @peterbourgon that
it wasn't that much longer! :-)

No reason to define it that way but now that we're on it, using
const instead of var would intuitively have the same effect. I don't think
this is part of the language though.

On Saturday, April 25, 2015, Chris Hines <notifications@github.com> wrote:

> Updated with results of the PR at ChrisHines/gokit#1
> <https://github.com/ChrisHines/gokit/pull/1>.
>
> @tsenart <https://github.com/tsenart> One thing still puzzles me, why did
> you define log.Timestamp like this:
>
> Timestamp = func(t func() time.Time) Valuer {
>     return func() interface{} { return t() }
> }
>
> Instead of like this:
>
> func Timestamp(t func() time.Time) Valuer {
>     return func() interface{} { return t() }
> }
>
> I don't see the scenario where we would want to change the Timestamp
> function. Isn't accepting a function argument flexible enough?
>
> —
> <https://github.com/peterbourgon/gokit/pull/32#issuecomment-96128146>.
>


-- 
Sent from Gmail Mobile

ChrisHines:
I'm happy with this PR now. @peterbourgon & @tsenart please take another look.

tsenart:
@ChrisHines: Two simplifications: https://github.com/ChrisHines/gokit/pull/2

ChrisHines:
@peterbourgon & @tsenart: All changes we discussed on my fork are now included in this PR.

peterbourgon:
The Travis failure was some spurious thing. LGTM! Thanks a ton!

PR #38, opened 2015-04-29
Move HTTP bindings to package transport

peterbourgon:
- Use reflection to generate new Request types for codecs
- Codec.Decode takes Request as out parameter, like json.Decode
- Relevant changes to addsvc

### DIFF
--- a/transport/codec/json/json_test.go
+++ b/transport/codec/json/json_test.go
@@ -1,0 +1,41 @@
+package json_test
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+
+	"golang.org/x/net/context"
+
+	jsoncodec "github.com/peterbourgon/gokit/transport/codec/json"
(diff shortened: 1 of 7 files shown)


transport/codec/json/json_test.go
```diff
@@ -0,0 +1,41 @@
...
+	Values []string `json:"values"`
+}
+
+func TestDecode(t *testing.T) {
+	buf := bytes.NewBufferString(`{"a":1,"b":"2"}`)
+	var req request
+	if _, err := jsoncodec.New().Decode(context.Background(), buf, &req); err != nil {
+		t.Fatal(err)
+	}
+	if want, have := (request{A: 1, B: "2"}), req; !reflect.DeepEqual(want, have) {
```
tsenart: `want != have` should be enough.

PR #39, opened 2015-05-04
metrics: prometheus: allow Summary and Histogram

peterbourgon:
Prometheus summaries are easy-to-use, but because they dynamically compute bucket sizes to establish quantiles, they cannot be meaningfully aggregated. So, we introduce a way to create a gokit Histogram via a Prometheus histogram, which carries the upfront cost of declaring your buckets, with the benefit of being able to aggregate across multiple timeseries. 

/cc @beorn7


metrics/prometheus/prometheus.go
```diff
@@ -134,21 +134,21 @@ func RegisterCallbackGaugeWithLabels(namespace, subsystem, name, help string, co
...
 	Pairs map[string]string
 }
 
-// NewHistogram returns a new Histogram backed by a Prometheus summary. It
-// uses a 10-second max age for bucketing. The histogram is automatically
-// registered via prometheus.Register.
-func NewHistogram(namespace, subsystem, name, help string, fieldKeys []string) metrics.Histogram {
-	return NewHistogramWithLabels(namespace, subsystem, name, help, fieldKeys, prometheus.Labels{})
+// NewSummary returns a new Histogram backed by a Prometheus summary. It uses
+// a 10-second max age for bucketing, emulating statsd. The histogram is
```
beorn7: Not related to this PR, but:

10s is probably fine for high-qps environment, but is it really intended to disregard any observation older than 10s, no matter what? (This is not configurable in gokit, as I understand...) Note that the default is 10m, two orders of magnitude more...

beorn7:
Code-wise, this looks all good to me.

Perhaps add a link somewhere to http://prometheus.io/docs/practices/histograms/ so that users can decide what type of histogram they want.

metrics/prometheus/prometheus.go
peterbourgon: I fixed it there because, to the best of my knowledge, it replicates the behavior of statsd as it exports to graphite, and I defer to statsd's opinions. If you think it's worthwhile, I'll add a new constructor that parameterizes the max age.

beorn7: Users at SC rarely needed to change the 10m max age. That doesn't prove that they wouldn't have been happier with 10s, though... It's a tough call how many of the config options you want to let leak into an abstracted/unified API like gokit metrics. Going with statsd sounds reasonable to me for the sake of unification. Given that I have created all the options in prometheus.SummaryOptions, I think they are all worthwhile. ;) I have to leave it to your judgement what is desired for gokit. I believe you have the understanding now to come to an informed decision.

peterbourgon:
Just for the record, that last commit had the effect of making an unspecified max age act as the Prometheus DefMaxAge i.e. 10 minutes.

PR #41, opened 2015-05-15
Provide coherent READMEs for each package.

peterbourgon:
RFCs were a good way to kickstart discussions, but I think we have enough well-formed opinions, and we're far enough along in implementation, that they're no longer necessary to corral discussion. This PR will give each package a succinct README that describes its _raison d'être_, rationale (drawn from RFCs where appropriate) and several usage examples. After that's done, I'll probably archive the RFCs and move the API stability policy to a docs/ folder. Comments and feedback welcome.

### DIFF
--- a/log/README.md
+++ b/log/README.md
@@ -2,7 +2,7 @@
 
 `package log` is an opinionated package for logging in your service.
 It provides a minimal interface for structured logging, which may be wrapped to encode conventions, enforce type-safety, etc.
-It can used for both typical application log events, and log-structured data streams.
+It can be used for both typical application log events, and log-structured data streams.
 
 ## Rationale
 
--- a/log/README.md
+++ b/log/README.md
@@ -1,7 +1,7 @@
 # package log
 
-`package log` is an opinionated package for logging in your service.
-It provides a minimal interface for structured logging, which may be wrapped to encode conventions, enforce type-safety, etc.
+`package log` provides a minimal interface for structured logging in services.
+It may be wrapped to encode conventions, enforce type-safety, etc.
 It can be used for both typical application log events, and log-structured data streams.
 
 ## Rationale
--- a/log/README.md
+++ b/log/README.md
@@ -1,0 +1,54 @@
+# package log
+
+`package log` is an opinionated package for logging in your service.
+It provides a minimal interface for structured logging, which may be wrapped to encode conventions, enforce type-safety, etc.
+It can used for both typical application log events, and log-structured data streams.
+
+## Rationale
+
+TODO
+
+## Usage
+
+Typical application logging.
+
+```go
+import "github.com/go-kit/kit/log"
+
+func main() {
+	logger := log.NewPrefixLogger(os.Stderr)
+	logger.Log("question", "what is the meaning of life?", "answer", 42)
+}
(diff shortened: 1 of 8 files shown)


log/README.md
```diff
@@ -0,0 +1,54 @@
+# package log
+
+`package log` is an opinionated package for logging in your service.
+It provides a minimal interface for structured logging, which may be wrapped to encode conventions, enforce type-safety, etc.
+It can used for both typical application log events, and log-structured data streams.
```
ChrisHines: It can ^be^ used ...

ChrisHines:
LGTM. We should work to put the code examples in the godoc of each package also.

peterbourgon:
Yeah, I like the idea of proper examples. I'll roll that into this PR.

log/README.md
```diff
@@ -0,0 +1,54 @@
+# package log
+
+`package log` is an opinionated package for logging in your service.
```
harlow: Thoughts on removing:
> is an opinionated package for logging in your service.

Consider:

`package log` provides a minimal interface for structured logging, which may be wrapped to encode conventions, enforce type-safety, etc.

harlow:
> Yeah, I like the idea of proper examples. I'll roll that into this PR.

Agreed :+1: 

Though I'd consider removing all the references to "it's an opinionated..." I feel like it's distracting to have in each of the intro paragraphs.

peterbourgon:
@harlow if I merge this to master, would you be OK with rebasing your branch and submitting a new PR? Is that the easiest way?

harlow:
@peterbourgon sure thing!

peterbourgon:
Gonna continue with the rationales and proper examples in a subsequent PR by @harlow

eduncan911:
Just a comment... What about auto generating the READMEs based on the go documentation of code?

http://godoc.org/github.com/davecheney/godoc2md

Then write an accompanying doc.go for each package. 

Output looks like this:

https://github.com/eduncan911/es
^- and peak through the importable packages (folders) for additional READMEs.

By having the `doc.go` file for each package and sub-package, not only is the github's `README.md` check, but `godoc.or/` is also kept in check with the exact same contents:

http://godoc.org/github.com/eduncan911/es

eduncan911:
As for Travis build badges and other misc badge-links, put them in a BADGES.md file and write a small script that:

Loops through each directory.  If there is `*.go` code:

* 1) execute `godoc2md github.com/eduncan911/es/bus > README.md`
* 2) prepend: `cat BADGES.md README.md > README.md~; mv README.md~ README.md`

ChrisHines:
I like the idea in general, but I'm not a fan of having the full API in the README. Is there any way to only include the package summary that godoc puts before the index, including package scoped examples?  That feels right to me. If someone wants to see all the docs a link to godoc.org is sufficient IMHO.

eduncan911:
I believe that can be done with the templates, yes.  His needs work anyhow.

peterbourgon:
@ChrisHines echoes my opinion exactly. I think the workflow is a good idea for the various packages. We can keep the badges exclusively in the top-level README, which I guess will always be handcrafted. I've created #52.

PR #43, opened 2015-05-23
Fix bugs in StdlibWriter and tests.

ChrisHines:
Two fixes in this PR.

1. We were passing data strings to the format argument of Printf and Fprintf. Not good if the data contains any %'s.
2. go test -cpu 2,4 was failing because `TestStdlibAdapterUsage` was calling `log.SetFlags`. On the second pass `TestStdlibWriter` would fail because the logger flags were not set to the defaults.

### DIFF
--- a/log/stdlib.go
+++ b/log/stdlib.go
@@ -7,14 +7,14 @@ import (
 	"strings"
 )
 
-// StdlibWriter implements io.Writer by invoking the stdlib log.Printf. It's
+// StdlibWriter implements io.Writer by invoking the stdlib log.Print. It's
 // designed to be passed to a gokit logger as the writer, for cases where it's
 // desirable to pipe all log output to the same, canonical destination.
 type StdlibWriter struct{}
 
 // Write implements io.Writer.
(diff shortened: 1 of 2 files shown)


peterbourgon:
Yep, I follow. :dango:!

ChrisHines:
@peterbourgon  said:

>Yep, I follow. :dango:!

That's the second time I've seen you "sign off" on a PR with :dango: and I'm trying to understand the reference. From [Wikipedia](http://en.wikipedia.org/wiki/Dango):

>A common Japanese proverb “Hana yori dango” (花より団子?, literally, “dumplings rather than flowers”) refers to a preference for practical things rather than aesthetics.

Or from the [disambiguation page](http://en.wikipedia.org/wiki/Dango_(disambiguation)):

>Dango (go game), a derogatory term used by go players internationally referring to an inefficient, dumpling-like cluster of stones in a go game

Is there some other interpretation that's relevant? Please enlighten me.

peterbourgon:
Haha. It's just my :+1: or LGTM.

Issue #42, opened 2015-05-22
Proposal: Remove log.DefaultLogger


ChrisHines:
### Rationale

Logging is an application concern and should be controlled by applications. Each application should decide how log events are filtered and routed. Widely reusable (open source) libraries shouldn't do any logging, preferring to return errors or provide some other mechanism for event notification and leave it to the application to hook that into its choice for logging if desired. Organizations may enforce a logging standard for all of their internal packages. In that case internal libraries can reasonably assume a standard approach to logging and produce log events in accordance, but the application should still control filtering and routing.

### Recommended Best Practice

Each source of log events should provide a mechanism for accepting a `Logger` implementation from the main application. The scope of the accepted `Logger` determines the granularity of control over filtering and routing available.  Some possibilities:

- Function arguments
- Context value or constructor argument
- Package scoped function or variable

Packages that do not create log events do not need any of these. Thus, **`log.DefaultLogger` is not needed as package log is not a source of log events**.

Of the above choices, a package scoped logger is attractive. Ignoring concurrency issues for the time being, each package that creates log events could include a declaration such as:
```go
var Log log.Logger = log.NewDiscardLogger()
```
Now the application can control logging for each package by changing its `Log` variable. If an application wanted to identify the source package for each log event it could do so by supplying a different contextual logger to each package. For example:
```go
packageA.Log = log.With(logger, "module", "packageA")
```

ChrisHines:
One more thought. Organizations that wish to standardize their logging across projects will likely create an internal logging package built from gokit/log pieces and other packages (e.g. [lumberjack](gopkg.in/natefinch/lumberjack.v2)). Such a package would define local logging policies. It would construct a base `Logger` from configuration data. It may choose to export a `DefaultLogger` if desired. In short, it encodes the local policy decisions that are out of scope for a general purpose library such as go-kit.

harlow:
I like this a lot `var Log log.Logger = log.NewDiscardLogger()`. I'm curious what the concurrency issues are here?

ChrisHines:
The concurrency issue arises if one goroutine assigns a new value to `Log` while another goroutine is logging with it. There could be a data race accessing the `Log` variable. I have an idea to handle this but didn't want to complicate the discussion here.

harlow:
Ok I see. That makes sense :thumbsup:

peterbourgon:
[...]

PR #44, opened 2015-05-23
Add SwapLogger.

ChrisHines:
Solution to the concurrency problem mentioned in https://github.com/go-kit/kit/issues/42#issuecomment-104809294.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -62,8 +62,11 @@ func (f LoggerFunc) Log(keyvals ...interface{}) error {
 }
 
 // SwapLogger wraps another logger that may be safely replaced while other
-// goroutines use the SwapLogger concurrently. An uninitialized SwapLogger
+// goroutines use the SwapLogger concurrently. The zero value for a SwapLogger
 // will discard all log events without error.
+//
+// SwapLogger serves well as a package global logger that can be changed by
+// importers.
 type SwapLogger struct {
 	logger atomic.Value
 }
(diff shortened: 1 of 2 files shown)


log/log.go
```diff
@@ -58,3 +60,32 @@ type LoggerFunc func(...interface{}) error
 func (f LoggerFunc) Log(keyvals ...interface{}) error {
 	return f(keyvals...)
 }
+
+// SwapLogger wraps another logger that may be safely replaced while other
+// goroutines use the SwapLogger concurrently.
```
peterbourgon: Would you mind adding a comment about intended usage? e.g. "One application of a SwapLogger would be a package global logger that can be changed by importers."

peterbourgon:
:+1:

peterbourgon:
Hmm. What do you think about `logger.Swap(nil)` as a shortcut for `logger.Swap(log.NewDiscardLogger())`?

ChrisHines:
Wrapping a `nil` logger could also be special cased in Log.

```go
func (l *SwapLogger) Log(keyvals ...interface{}) error {
    if l.logger == nil {
        return nil
    }
    return l.logger.Load().(loggerStruct).Log(keyvals...)
}
```

This would guard against a runtime panic the current code doesn't prevent, and also further delay the need to export a `NewDiscardLogger`.

log/log.go
ChrisHines: I'm not opposed to adding that, but I wonder if it would be better to put it in the package docs if we include other advice on how to best use this package. Maybe it belongs in both. Thoughts?

peterbourgon: That's also a fine place for it. I don't have a strong opinion.

peterbourgon:
Sounds reasonable to me.

ChrisHines:
@peterbourgon I've updated the PR. With the latest change the zero value has some meaning, which raises the question of whether we should still have `NewSwapLogger`. What do you think?

peterbourgon:
Interesting. If this works without panicing

```go
var logger log.SwapLogger
logger.Log("going", "nowhere")
```

then I agree we don't need and shouldn't have the constructor.

ChrisHines:
Hopefully the last update. @peterbourgon please take one more look.

peterbourgon:
:+1:!

PR #45, opened 2015-05-23
Fix Zipkin tracing, and introduce package client

peterbourgon:
—


peterbourgon:
The main thing with this PR is several fixes to package tracing/zipkin, so it actually works now :)




In order to test that, I wanted one addsvc to call another. So, it was a good time to introduce the concept  of a client, concretely [transport/http.Client](https://sourcegraph.com/github.com/go-kit/kit@663cea4656e10ae53d6211cfe0645e65acd9c2c5/.tree/transport/http/client.go#startline=25&endline=25). Working through that made it clear we'd want a client-side Endpoint abstraction for things like circuit breakers. And it turned out that had the exact same signature as the server-side Endpoint. So we break that out to [endpoint.Endpoint](https://sourcegraph.com/github.com/go-kit/kit@663cea4656e10ae53d6211cfe0645e65acd9c2c5/.tree/endpoint/endpoint.go#startline=11&endline=11). Packages server and client will re-appear in the future, for server- and client-side value-add components.

Future work:

- Thrift and gRPC clients
- package endpoint/ratelimit
- package client/cb
- package server/healthcheck

/cc @ChrisHines @tsenart

endpoint/endpoint.go
```diff
@@ -0,0 +1,21 @@
+package endpoint
```
tsenart: I secretly wished this package would be the `end` and have a type called `Point`.

tsenart:
It's really great to see Zipkin working! Good work!

addsvc/grpc_binding.go
```diff
@@ -20,15 +20,15 @@ type grpcBinding struct{ server.Endpoint }
 // way to manipulate the RPC context, like headers for HTTP. So we don't have
 // a way to transport e.g. Zipkin IDs with the request. TODO.
 func (b grpcBinding) Add(ctx context.Context, req *pb.AddRequest) (*pb.AddReply, error) {
```
harlow: @peterbourgon where do you envision using the `TraceID` w/ Zipkin? We have the ability to add metadata to the context of gRPC library like this:

```go
import "google.golang.org/grpc/metadata"
// ...

md := metadata.Pairs("TraceID", token)
ctx := context.Background()
ctx = metadata.NewContext(ctx, md)
```

Then we can read it back out like this:

```go
func (b grpcBinding) Add(ctx context.Context, req *pb.AddRequest) (*pb.AddReply, error) {
  md, _ := metadata.FromContext(ctx)
  // access to the TraceID: md["TraceID"]
}
```

https://godoc.org/google.golang.org/grpc/metadata

I can add a PR for this if you like. I'm having a little trouble getting the forked version of the repo running as all the include paths point the mainline. Not sure if there is an easier way to deal with that.

peterbourgon: Yes, we'll definitely add the traceID/spanID/parentSpanIDs to the context. But we'll probably need to make the keys configurable, as only Zipkin-over-HTTP has well-defined keys ([these](https://sourcegraph.com/github.com/go-kit/kit@8662a12630268cc431ff3825eb46f49a0719a894/.tree/tracing/zipkin/zipkin.go#startline=23&endline=26)), to the best of my knowledge.

> 'm having a little trouble getting the forked version of the repo running as all the include paths point the mainline.

The way to fork in Go is to add your fork as a git remote to the canonical checkout. That is,

```
$ cd $GOPATH/src/github.com/go-kit/kit
$ git remote add harlow git@github.com:harlow/kit
$ git checkout -b my-feature-branch
$ # commit commit commit
$ git push harlow my-feature-branch
```

harlow: Thanks for this :+1: I was doing the opposite and using main repo as `upstream`. This works much better.

Issue #34, opened 2015-04-29
log.PrefixLogger should become logfmt compliant


peterbourgon:
[logfmt](https://godoc.org/github.com/kr/logfmt)

ChrisHines:
+1

peterbourgon:
Might also involve a rename.

ChrisHines:
I agree a rename is warranted. I was going to raise this exact issue with PrefixLogger today if you hadn't. My apps currently use logfmt via the log15 implementation, and I've used the package linked above for parsing the logs.

I've been considering whether it would be worth it to write an encoding/json or encoding/xml style package that supports marshaling and unmarshaling logfmt data. Thoughts on that idea?

peterbourgon:
Does that imply logfmt would become the common denominator format/encoding for gokit/log? If so, do you believe it's mature enough for that?

ChrisHines:
The RFC says:
- Log SHALL be format agnostic by providing an interface for log formatting.
- Log MAY provide some built in formatters such as RFC3164, JSON, etc.

So I don't think it has to take on that role but I would like it to be one of the choices.

The logfmt package I was postulating would be standalone and gokit could chose to import it or not. It would be awesome if someone has already written such a package, but I haven't found one yet.

Regarding maturity, the best link I can find that speaks to that is: https://brandur.org/logfmt

peterbourgon:
Ah, so your "an encoding/json or encoding/xml style package that supports marshaling and unmarshaling logfmt data" would be external to gokit/log?

ChrisHines:
That was the idea, yes. It would hopefully be general enough for use in other logging packages or standalone, as well as providing the parsing side. It should also support [encoding.TextMarshaler](http://golang.org/pkg/encoding/#TextMarshaler) and [encoding.TextUnmarshaler](http://golang.org/pkg/encoding/#TextUnmarshaler) interfaces.

peterbourgon:
:+1: to the idea!

ChrisHines:
Status update: I [checked](https://github.com/kr/logfmt/issues/5) that the authors of https://github.com/kr/logfmt are OK with me creating a new Go project for logfmt. I have created [the project](https://github.com/go-logfmt/logfmt) and started implementing a `MarshalKeyvals` function built on top of a lower level `Encoder`.

The package is still unstable, tests are incomplete, and some tests still fail, but for logging basic data types from well behaved sources the `MarshalKeyvals` function will get the job done.

[...]

PR #47, opened 2015-05-25
Replace PrefixLogger with LogfmtLogger (fix for #34).

ChrisHines:
- Fix for #34.
- Adds dependency on gopkg.in/logfmt.v0.
- Breaks logging of compound data types until the logfmt package has support for that.
- Improves performance:

Before:
```
BenchmarkPrefixLoggerSimple       500000      2916 ns/op     322 B/op          8 allocs/op
BenchmarkPrefixLoggerContextual   300000      4723 ns/op     531 B/op         13 allocs/op
```
After:
```
BenchmarkLogfmtLoggerSimple      1000000      2155 ns/op     272 B/op          6 allocs/op
BenchmarkLogfmtLoggerContextual   500000      3502 ns/op     480 B/op         10 allocs/op
```


ChrisHines:
See https://github.com/go-logfmt/logfmt/issues/1 for discussion about compound data types.

log/logfmt_logger_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	buf.Reset()
+	if err := logger.Log("a", 1, "err", errors.New("error")); err != nil {
+		t.Fatal(err)
+	}
+	if want, have := "a=1 err=error\n", buf.String(); want != have {
+		t.Errorf("want %#v, have %#v", want, have)
+	}
+
+	// Logfmt does not have a handle compound types at this time.
```
harlow: how do you feel about remove the comments and creating an issue?

log/logfmt_logger.go
```diff
@@ -0,0 +1,29 @@
...
+type logfmtLogger struct {
+	w io.Writer
+}
+
+// NewLogfmtLogger returns a basic logger that encodes keyvals as simple "k=v"
+// pairs to the Writer.
+func NewLogfmtLogger(w io.Writer) Logger {
+	return &logfmtLogger{w}
+}
+
```
harlow: Can we add a GoDoc style comment for Exported func `Log`?

log/logfmt_logger_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func BenchmarkLogfmtLoggerContextual(b *testing.B) {
+	benchmarkRunner(b, log.NewLogfmtLogger(ioutil.Discard), withMessage)
+}
+
+func TestLogfmtLoggerConcurrency(t *testing.T) {
+	testConcurrency(t, log.NewLogfmtLogger(ioutil.Discard))
+}
+
+type mymap map[int]int
```
harlow: doesn't look like these are used. Can we remove?

harlow:
Looking good @ChrisHines :+1:

log/logfmt_logger.go
ChrisHines: golint doesn't complain about it. Curious why this is, I looked at the lint code and learned that it doesn't complain because the receiver type (`logfmtLogger`) is not exported. Because the receiver type is not exported, the `Log` method does not appear in the godoc.

We could add a comment such as

    // Log implements the Logger interface.

But that seems like noise in this case.

log/logfmt_logger_test.go
ChrisHines: They are "used" by the commented out code in TestLogfmtLogger. If that code gets deleted then the `mymap` code should be deleted as well.

log/logfmt_logger_test.go
ChrisHines: I would prefer to track the issue already created in the go-logfmt/logfmt repo linked above.

log/logfmt_logger.go
harlow: Nice find -- Wasn't aware of that either. Seems unfortunate it won't appear in GoDoc as it's the most important function in this package!

ChrisHines: NewLogfmtLogger has the required docs, which now that I'm looking at them needs updating.

ChrisHines:
I just pushed https://github.com/go-logfmt/logfmt/commit/c50dc9aa845e681529d8e49d3dc11ec0195aacac which I believe makes go-logfmt/logfmt good enough to merge this PR.

log/logfmt_logger.go
```diff
@@ -0,0 +1,29 @@
...
+// NewLogfmtLogger returns a logger that encodes keyvals to the Writer in
+// logfmt format.
+func NewLogfmtLogger(w io.Writer) Logger {
+	return &logfmtLogger{w}
+}
+
+func (l logfmtLogger) Log(keyvals ...interface{}) error {
+	b, err := logfmt.MarshalKeyvals(keyvals...)
```
peterbourgon: I'd like to see values encoded directly to the writer, rather than using this intermediate buffer. [From what I can see](https://github.com/go-logfmt/logfmt/blob/c50dc9aa845e681529d8e49d3dc11ec0195aacac/logfmt.go#L31) this should be possible?

log/logfmt_logger_test.go
```diff
@@ -32,21 +32,21 @@ func TestPrefixLogger(t *testing.T) {
 	if err := logger.Log("std_map", map[int]int{1: 2}, "my_map", mymap{0: 0}); err != nil {
 		t.Fatal(err)
 	}
-	if want, have := "std_map=map[1:2] my_map=special_behavior\n", buf.String(); want != have {
+	if want, have := "std_map=\"unsupported value type\" my_map=special_behavior\n", buf.String(); want != have {
```
peterbourgon:
```go
if want, have := "std_map=\"" + logfmt.ErrUnsupportedValueType.Error() + "\" ...", buf.String(); want != have {
```

peterbourgon:
Those points aside, LGTM.

log/logfmt_logger.go
ChrisHines: The intermediate buffer is there to avoid interleaving calls to the `Writer` from concurrent calls to `Log`. The goal is to have a 1:1 connection between calls to `Log` and calls to `w.Write`.

I will add a comment to document that constraint.

log/logfmt_logger_test.go
ChrisHines: Good idea.

peterbourgon:
Great! Thank you! Merge away.

PR #48, opened 2015-05-25
Add Log example output

harlow:
Show examples of log lines from os.Stderr output.


peterbourgon:
This PR seems oddly rebased. See [this comment in #41](https://github.com/go-kit/kit/pull/41#issuecomment-105650390)...

PR #50, opened 2015-05-27
Add Log example output

harlow:
Show examples of log lines from os.Stderr output.

### DIFF
--- a/log/README.md
+++ b/log/README.md
@@ -1,6 +1,7 @@
 # package log
 
-`package log` provides a minimal interface for structured logging in services.
+`package log` provides an interface for structured logging.
+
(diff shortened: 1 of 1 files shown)


peterbourgon:
Can you roll this and all other updates you plan to do in the scope of #52 into a single PR?

log/README.md
```diff
@@ -21,6 +22,12 @@ func main() {
 }
 ```
 
+The log lines are printed in tuples separated with by a `=` symbol.
+
+```
+question=what is the meaning of life? answer=42
```
BRMatt: Small note, I think that if a string contains spaces logfmt will [ensure it's quoted](http://godoc.org/github.com/kr/logfmt).

PR #57, opened 2015-06-10
package ratelimit, with token-bucket throttler

peterbourgon:
Hey @tsenart, does this make sense?


tsenart:
I have to admit I'm not so happy with how I built `tb`. Specifically, it could have been designed without extra go-routines and with much better APIs. It's on my TODO list to address this, but I haven't been able to find the time :/

ratelimit/token_bucket.go
```diff
@@ -0,0 +1,67 @@
...
+
+type tokenBucketThrottler struct {
+	freq time.Duration
+	key  string
+	rate int64
+	take int64
+}
+
+// TokenBucketOption sets an option on the token bucket throttler.
+type TokenBucketOption func(*tokenBucketThrottler)
```
tsenart: I wouldn't make a public type that takes privately typed arguments.

tsenart:
I don't know if the licenses are compatible but this package at least doesn't require extra background go-routines: https://github.com/juju/ratelimit

ratelimit/token_bucket.go
peterbourgon: This is the functional options idiom. The private type is an implementation detail, users never interact with it. Still no?

aybabtme:
Have this implementation that's not a token bucket persay but achieves the same result. 

https://github.com/aybabtme/iocontrol/blob/master/limiter.go

Using it is quite simple:

```go
func (t *throttledReader) Read(b []byte) (n int, err error) {
	canRead := t.limiter.CanDo()
	if len(b) <= canRead {
		// no throttling needed
		n, err = t.wrap.Read(b)
		t.limiter.Did(n)
		return n, err
	}

	if canRead > 0 {
		// read what can be read for this batch
		n, err = t.wrap.Read(b[:canRead])
	}

	t.limiter.Limit()

	// return bytes read and let caller try another read
	return n, err
}
```
https://github.com/aybabtme/iocontrol/blob/master/throttle.go#L41-L59

peterbourgon:
juju/ratelimit is LGPLv3 but they've carved out a static linking exception that should make it import-compatible with go-kit (MIT). I'll add an implementation based on that later tonight, maybe switch to it entirely if it's strictly superior.

peterbourgon:
Cool, we now have a TokenBucketLimiter, which just immediately errors out the request when the rate limit is reached, and a TokenBucketThrottler, which invokes a parameterized sleep function and then allows the request to proceed.

@davecheney and/or @rogpeppe — am I using this right?

@tsenart and/or @aybabtme — care to take a look?

peterbourgon:
I'm happy enough with this as-is to merge.

rogpeppe:
Sorry for late commenting - I only just noticed the PR.

PR #58, opened 2015-06-10
package circuitbreaker

peterbourgon:
Various implementations.

### DIFF
--- a/circuitbreaker/sony_gobreaker_test.go
+++ b/circuitbreaker/sony_gobreaker_test.go
@@ -1,0 +1,77 @@
+package circuitbreaker_test
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/sony/gobreaker"
+	"golang.org/x/net/context"
+
+	"github.com/go-kit/kit/circuitbreaker"
+	"github.com/go-kit/kit/endpoint"
+)
+
+func TestSonyCircuitBreaker(t *testing.T) {
+	var (
+		thru        int
+		last        gobreaker.State
+		myError     = errors.New("❤️")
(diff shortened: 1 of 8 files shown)


circuitbreaker/sony_gobreaker_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if want, have := gobreaker.StateOpen, last; want != have { // tripped
+		t.Errorf("want %v, have %v", want, have)
+	}
+
+	e(context.Background(), struct{}{})
+	if want, have := 6, thru; want != have { // didn't get thru
+		t.Errorf("want %d, have %d", want, have)
+	}
+
+	time.Sleep(2 * timeout)
```
aybabtme: using something like https://github.com/benbjohnson/clock to avoid sleeping in tests is great IMO. but I think it would require you to modify the circuit breaker vendored code.

tsenart: Totally agree that parameterizing the time function is a good idea but it's probably over-kill to add another dependency for that when you can simply do:
```go
type TimeFunc func() time.Time
```

However, `github.com/sony/gobreaker` doesn't allow this to happen currently. So it would need to be changed first.

tsenart:
Can we use labels in PRs describing their state? Is this a WIP?

circuitbreaker/sony_gobreaker_test.go
aybabtme: agreed

peterbourgon:
@tsenart — interested to give it a look?

peterbourgon:
Gonna try to generalize the unit tests before merging this...

PR #60, opened 2015-06-13
[WIP] package loadbalancer

peterbourgon:
Given a service with multiple instances,

- Publisher emits current set of endpoints
- Strategy wraps a publisher and yields endpoints


peterbourgon:
Closed in favor of #67

PR #61, opened 2015-06-14
add hystrix circuit breaker

josler:
adding on top of https://github.com/go-kit/kit/pull/58

### DIFF
--- a/circuitbreaker/hystrix.go
+++ b/circuitbreaker/hystrix.go
@@ -1,0 +1,37 @@
+package circuitbreaker
+
+import (
+	"github.com/afex/hystrix-go/hystrix"
+	"golang.org/x/net/context"
+
+	"github.com/go-kit/kit/endpoint"
+)
+
+// Hystrix returns an endpoint.Middleware that implements the circuit
+// breaker pattern using the afex/hystrix-go package.
+//
+// When using this circuit breaker, please configure your commands separately.
+//
+// See https://godoc.org/github.com/afex/hystrix-go/hystrix for more
+// information.
+func Hystrix(commandName string) endpoint.Middleware {
+	return func(next endpoint.Endpoint) endpoint.Endpoint {
+		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
+			output := make(chan interface{}, 1)
+			errors := hystrix.Go(commandName, func() error {
+				resp, err := next(ctx, request)
+				if err == nil {
+					output <- resp
+				}
+				return err
+			}, nil)
+
+			select {
+			case resp := <-output:
+				return resp, nil
+			case err := <-errors:
+				return nil, err
+			}
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


circuitbreaker/hystrix.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"errors"
+
+	"github.com/afex/hystrix-go/hystrix"
+	"github.com/go-kit/kit/endpoint"
+	"golang.org/x/net/context"
+)
+
+var ErrCircuitBreakerTimeout = errors.New("circuit breaker timeout")
```
josler: might want to rationalize this this with `ErrCircuitBreakerOpen` in `handy_breaker.go`

circuitbreaker/hystrix.go
```diff
@@ -0,0 +1,55 @@
+package circuitbreaker
+
+import (
+	"errors"
+
+	"github.com/afex/hystrix-go/hystrix"
+	"github.com/go-kit/kit/endpoint"
```
peterbourgon: goimports style: move this import statement to a third group, separated by a blank line, please :)

circuitbreaker/hystrix.go
```diff
@@ -0,0 +1,55 @@
...
+func wellKnownHystrixError(err error) error {
+	switch err {
+	case hystrix.ErrCircuitOpen:
+		return ErrCircuitBreakerOpen
+	case hystrix.ErrTimeout:
+		return ErrCircuitBreakerTimeout
+	default:
+		return err
+	}
+}
```
peterbourgon: Any reason we shouldn't return these errors directly?

circuitbreaker/hystrix_test.go
```diff
@@ -0,0 +1,116 @@
...
+		if _, err := e(context.Background(), struct{}{}); err != circuitbreaker.ErrCircuitBreakerOpen {
+			t.Errorf("with request #%d, want %v, have %v", primeWith+letThru+i, circuitbreaker.ErrCircuitBreakerOpen, err)
+		}
+	}
+
+	// Confirm the rest didn't get through.
+	if want, have := primeWith+letThru, thru; want != have {
+		t.Errorf("want %d, have %d", want, have)
+	}
+}
```
peterbourgon: We should factor a lot of this out to a general circuitbreaker test driver, but I'll do that afterwards.

circuitbreaker/hystrix.go
josler: Perhaps it'd be nice to return the same error type for all circuit breakers implemented? If that's not the desired behaviour we can just remove this :+1:

peterbourgon: There may indeed be value in doing that, but I think I prefer the idea of defaulting to transparency, and just returning the errors directly. If it turns out users want consistent errors independent of concrete implementation, we can add it in by request, later.

josler:
Updated - transparent errors, import cleanup

josler:
Hm. Travis seems to not like running the tests concurrently.

peterbourgon:
Thanks a bunch!

PR #65, opened 2015-06-19
ratelimit: simplify

rogpeppe:
Following on from my comments on #57, here's an implementation
of those ideas. This removes token bucket creation entirely
from the package, giving clients the freedom to create the
bucket in any way they choose.

For example, it's common to want to start with the token bucket
empty, but the current implementation does not provide this possibility
without adding extra options.

It also fixes the implementation - the current implementation will
allow a request even if there are less tokens available than requested.

### DIFF
--- a/ratelimit/export_test.go
+++ b/ratelimit/export_test.go
@@ -1,0 +1,3 @@
+package ratelimit
+
+var TimeSleep = &timeSleep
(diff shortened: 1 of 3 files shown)


peterbourgon:
Ah, I missed this PR before I started working on [my own simplification](). Sorry for that.

I'm into passing a juju/ratelimit.Bucket directly, but I'm not so into the intermediate Bucket interface, and I want to preserve two distinct limiter constructors (one to fail, one to sleep) because I think, from the user perspective, those are different-enough to warrant it.

If you don't mind, could you take a quick skim through #66 and let me know what you think?

peterbourgon:
In the interest of forward motion I'll do what I said above. @rogpeppe please do open another issue if there's something else that can be addressed.

PR #74, opened 2015-07-10
loadbalancer: do not expose sync.Mutex as public field

### LINKED ISSUE #72: loadbalancer.StaticPublisher should not embed sync.Mutex
External clients should not be able to manipulate the mutex,
which is an implementation detail.

rogpeppe:
Fixes #72.


tsenart:
LGTM

PR #77, opened 2015-07-11
basic generator for the boilerplate of an endpoint definition #70

dahernan:
This is a starting point to generate boilerplate code for a given endpoint (I don't expect this PR to be merged, could be use it as inspiration, or maybe try another approach to solve the problem).

Steps to generate boilerplate:
* Given the endpoint definition here `process/process.go` (similar to what is expose in issue #70)
* The file must be anotated with:
```
//go:generate go run ../gen-cmd/main.go packageName
//go:generate go fmt
```
* Generate the boilerplate doing:
```
$ cd process
$ go generate
```
* The generated file will be `process_definition.go`

Know issues
* Uses relatives paths to refer the template and the generator command
* I don't like the way I get the packageName as an argument in the generation, to use it as a filename


gen-cmd/main.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	ast.Walk(vs, f)
+
+	// TODO remove relative path
+	mergeTemplate(vs, "definition", "../gen-cmd/endpoint_definition.tmpl")
+}
+
+func buildAst(pck string) *ast.File {
+	definitions := pck + ".go"
```
dahernan: Not nice to get the file from the packageName requires the convention of the file `packageName.go`

gen-cmd/main.go
```diff
@@ -0,0 +1,208 @@
...
+
+		// function helpers
+		ReturnExp: ReturnExp,
+		ArgsExp:   ArgsExp,
+		CompoExp:  CompoExp,
+	}
+
+	ast.Walk(vs, f)
+
+	// TODO remove relative path
```
dahernan: I should get rid of relative paths, good only for first prototype

ernesto-jimenez:
@dahernan FYI, rather than running `go fmt` you can use stdlib's `go/format` package to format your generated code before writing it to a file.

dahernan:
My pleasure, thank you :smile:

Issue #78, opened 2015-07-21
Fixes to package zipkin


peterbourgon:
1. If the Scribe collector is down, we should drop the batch on the floor, rather than growing it indefinitely in memory.
2. We should provide support for sampling. Need to investigate how that is commonly done.
3. Remove package global logger in favor of logger parameter to relevant constructors (h/t @rogpeppe)

peterbourgon:
There is no spec/standard way to handle dynamic sampling in Zipkin. Therefore, #79 closes this issue.

PR #79, opened 2015-07-21
Zipkin enhancements

peterbourgon:
Addresses #78.


peterbourgon:
Still pending a dynamic updating mechanism for the sample rate, but that can be a later PR.

PR #85, opened 2015-07-29
Logging nil error pointer must not panic

husio:
When logging, every value is checked if it implements `error` interface. When
positive, `Error` method is called. This may cause panic when passed value is
`nil` pointer to structure that use it's internal state to produce `Error`
method optput.
Before calling `Error` on any error implementing structure, make sure it's not
nil pointer first.

Closing #86

### DIFF
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"reflect"
 )
 
 type jsonLogger struct {
@@ -37,7 +38,7 @@ func merge(dst map[string]interface{}, k, v interface{}) map[string]interface{} 
 	default:
 		key = fmt.Sprintf("%v", x)
 	}
-	if x, ok := v.(error); ok {
+	if x, ok := v.(error); ok && !reflect.ValueOf(v).IsNil() {
 		v = x.Error()
 	}
 	dst[key] = v
--- a/log/json_logger_test.go
+++ b/log/json_logger_test.go
@@ -20,6 +20,19 @@ func TestJSONLogger(t *testing.T) {
 	}
 }
 
+func TestJSONLoggerNilError(t *testing.T) {
+	var err error
+
+	buf := &bytes.Buffer{}
+	logger := log.NewJSONLogger(buf)
+	if err := logger.Log("err", err); err != nil {
+		t.Fatal(err)
+	}
+	if want, have := `{"err":null}`+"\n", buf.String(); want != have {
+		t.Errorf("want %#v, have %#v", want, have)
+	}
+}
+
 func BenchmarkJSONLoggerSimple(b *testing.B) {
 	benchmarkRunner(b, log.NewJSONLogger(ioutil.Discard), baseMessage)
 }


peterbourgon:
```
BenchmarkJSONLoggerSimple        master  300000   4000 ns/op  857 B/op   20 allocs/op
                                 husio   500000   3854 ns/op  809 B/op   19 allocs/op
                                 diff             -146        -48        -1
                                 pct              -4%         -6%        -5%

BenchmarkJSONLoggerContextual    master  300000   5375 ns/op  1089 B/op  26 allocs/op
                                 husio   300000   5309 ns/op  1122 B/op  27 allocs/op
                                 diff             -66         +33        +1
                                 pct              -1%         +3%        +4%

BenchmarkDiscard                 master  5000000  253 ns/op   64 B/op    3 allocs/op
                                 husio   5000000  232 ns/op   64 B/op    3 allocs/op
                                 diff             -21         0          0
                                 pct              -9%         0%         0%

BenchmarkOneWith                 master  5000000  362 ns/op   96 B/op    3 allocs/op
                                 husio   3000000  427 ns/op   128 B/op   4 allocs/op
                                 diff             +65         +32        +1
                                 pct              +15%        +25%       25%

BenchmarkTwoWith                 master  3000000  475 ns/op   160 B/op   3 allocs/op
                                 husio   3000000  509 ns/op   192 B/op   4 allocs/op
                                 diff             +34         +32        +1
                                 pct              7%          17%        25%

BenchmarkTenWith                 master  1000000  1290 ns/op  672 B/op   3 allocs/op
                                 husio   1000000  1233 ns/op  704 B/op   4 allocs/op
                                 diff             -57         +32        +1
                                 pct              -5%         5%         25%

BenchmarkLogfmtLoggerSimple      master  1000000  2185 ns/op  432 B/op   7 allocs/op
                                 husio   1000000  1884 ns/op  384 B/op   6 allocs/op
                                 diff             -301        -48        -1
                                 pct              -16%        -13%       -17%

BenchmarkLogfmtLoggerContextual  master  500000   3034 ns/op  560 B/op   9 allocs/op
                                 husio   500000   2873 ns/op  592 B/op   10 allocs/op
                                 diff             -161        +32        +1
                                 pct              -6%         5%         10%

BenchmarkValueBindingTimestamp   master  1000000  1165 ns/op  144 B/op   5 allocs/op
                                 husio   1000000  1159 ns/op  176 B/op   6 allocs/op
                                 diff             -6          +32        +1
                                 pct              -1%         18%        17%

BenchmarkValueBindingCaller      master  1000000  1403 ns/op  112 B/op   4 allocs/op
                                 husio   1000000  1588 ns/op  144 B/op   5 allocs/op
                                 diff             +185        +32        +1
                                 pct              12%         22%        20%
```

I'm concerned about the extra allocations per operation that the reflect call seems to cause...

husio:
That is correct, but I don't know any better way of checking if value is not `nil` if it's an interface.

I think safety is work the performance degradation. After all, this check is only for error logging.

ChrisHines:
I have found the best way to handle these situations is to recover the panic before using reflection to inspect for `nil`. This helps keep the fast path fast. Code I've written to handle this in the past:

- [logfmt.writeValue](https://github.com/go-logfmt/logfmt/blob/master/logfmt.go#L174), which calls various ["safe" helpers](https://github.com/go-logfmt/logfmt/blob/master/logfmt.go#L251)
- https://github.com/inconshreveable/log15/commit/03ef3a3e78757fb8596ab2971c0613711eec19aa

Both of the above were inspired by [fmt.catchPanic](https://github.com/golang/go/blob/go1.4.2/src/fmt/print.go#L624).

husio:
I believe this pull request is obsolete, because @ChrisHines is working on better fix.

Issue #83, opened 2015-07-28
Publisher might have a simpler interface


peterbourgon:
Since clients of Publisher always request-then-use an endpoint, we might be able to simplify the idea of a Publisher, removing the concept of a stream of endpoints, and instead providing something more like

```go
type Something interface {
    Get() Endpoint
}
```

peterbourgon:
Closed by #90.

Issue #73, opened 2015-07-10
loadbalancer.NewDNSSRVPublisher is in the wrong place


rogpeppe:
It doesn't seem right for the loadbalancer package to have domain-specific Publisher implementations inside it.

peterbourgon:
We should create packages loadbalancer/{dnssrv,consul,etcd,...}, and {dnssrv,consul,etcd,...}.NewPublisher constructors. I suppose StaticPublisher may continue to live in loadbalancer.

peterbourgon:
Closed by #90.

PR #90, opened 2015-08-01
Enhancements to loadbalancer

peterbourgon:
Addresses #73 and #83.


peterbourgon:
@rogpeppe Thanks for your suggestions at GopherCon. I've changed the package and API and it feels a lot smoother. Have a few moments for a quick code review?

loadbalancer/README.md
```diff
@@ -17,28 +17,47 @@ TODO
...
-	return func(w http.ResponseWriter, r *http.Request) {
-		// foo is usable as a load-balanced remote endpoint.
-	}
+func fooFactory(instance string) (endpoint.Endpoint, error) {
+	// Convert an instance (host:port) to an endpoint, via a defined transport binding.
 }
 ```
+
+It's also possible to wrap a load balancer with a retry strategy, so that it
+can used as an endpoint directly. This may make load balancers more convenient
```
ChrisHines: s/can used/can be used/

loadbalancer/dnssrv/publisher.go
```diff
@@ -0,0 +1,122 @@
...
+
+		case <-p.quit:
+			return
+		}
+	}
+}
+
+// Endpoints implements the Publisher interface.
+func (p *Publisher) Endpoints() ([]endpoint.Endpoint, error) {
+	return <-p.endpoints, nil
```
ChrisHines: This could block forever once Stop() has been called. What about this?

```go
select {
    case endpoints := <- p.endpoints:
        return endpoints, nil
    case <-quit:
        return nil, ErrPublisherStopped
}

File: loadbalancer/README.md [hunk 3]
```diff
@@ -17,28 +17,47 @@ TODO
...
-}
+	// Construct a load balancer for foosvc, which gets foosvc instances by
+	// polling a specific DNS SRV name.
+	p := dnssrv.NewPublisher("foosvc.internal.domain", 5*time.Second, fooFactory, logger)
+	lb := loadbalancer.NewRoundRobin(p)
+
+	// Get a new endpoint from the load balancer.
+	endpoint, err := lb.Endpoint()
+	if err != nil {
+		panic(err)
```
Mistobaan: shouldn't just loop until we have a valid endpoint after waiting few seconds ?

loadbalancer/dnssrv/publisher.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	sort.Sort(sort.StringSlice(hostports))
+	h := md5.New()
+	for _, hostport := range hostports {
+		fmt.Fprintf(h, hostport)
+	}
+	return addrs, fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
+func lift(addrs []*net.SRV, f loadbalancer.Factory, logger log.Logger) []endpoint.Endpoint {
```
Mistobaan: what is the rationale for lift ? I don't understand the meaning of it in this context :)

loadbalancer/README.md
peterbourgon: If your DNS name has no instances, I believe it's much better to crash your program on startup. It's much better to do that, than to successfully start up, but leave you in a state where you can't complete any requests.

loadbalancer/dnssrv/publisher.go
peterbourgon: To lift, or convert, addrs to usable endpoints.

loadbalancer/dnssrv/publisher_internal_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+func TestBadLookup(t *testing.T) {
+	oldLookup := lookupSRV
+	defer func() { lookupSRV = oldLookup }()
+	lookupSRV = mockLookupSRV([]*net.SRV{}, errors.New("kaboom"), nil)
+
+	var (
+		name    = "some-name"
+		ttl     = time.Second
+		factory = func(string) (endpoint.Endpoint, error) { return nil, errors.New("unreachable") }
```
Mistobaan: See: http://play.golang.org/p/MGk22izFHt

Just an idea to don't repeat too much code

loadbalancer/dnssrv/publisher.go
Mistobaan: never heard of it :) devops slang ? good to know!

peterbourgon: No, @tsenart taught it to me, see https://wiki.haskell.org/Lifting

ChrisHines: Regardless of the theoretical basis, the verb lift is rather abstract in this context. Consider that the following function is named `addr2instance`, but it also seems to be a lifting function. In the interest of being less abstract, perhaps rename `lift` to something less abstract. My attempt at a name is `makeEndpoints`.

peterbourgon: Noted and agreed.

loadbalancer/README.md
peterbourgon: I've been thinking about this and may revisit this decision...

Issue #84, opened 2015-07-29
Panic in log package


husio:
`kit/log` package is using `panic` in several places, that are executed conditionally. While the code that is causing it (odd number of key-value pair elements) is logically invalid, `panic` is not something expected by external package, especially logging.
Odd number of elements means that there is one key with no value. My proposal is to pair such key with empty value. While it's not perfect, it does not break the code and can be detected later while looking at logs or by code analyzers. 
Similar logic is provided for example by `fmt` package - [using `XXXf` with invalid amount of parameters](http://play.golang.org/p/wq2D1_IybJ)  does not panic, but rather produce result with warning. Wrong parameters are also detected by [go vet](http://godoc.org/golang.org/x/tools/cmd/vet#hdr-Printf_family). 

* https://github.com/go-kit/kit/blob/master/log/log.go#L47
* https://github.com/go-kit/kit/blob/master/log/log.go#L61
* https://github.com/go-kit/kit/blob/master/log/log.go#L82

peterbourgon:
@ChrisHines what do you think?

```diff
 if len(keyvals)%2 != 0 {
-    panic("bad keyvals")
+    keyvals = append(keyvals, "(MISSING)")
 }
```

husio:
I don't have any preferences here. I think `(MISSING)` is easier to spot when browsing logs than empty string or `nil`. 
:+1:

tsenart:
+1

ChrisHines:
I agree that we should remove these panics.

In [logfmt.MarshalKeyvals](https://github.com/go-logfmt/logfmt/blob/master/logfmt.go#L20) I appended `nil` in this scenario. That package logs a `nil` value the same as JSON (i.e. "null").

At the slightly higher level of abstraction here in the `log` package it seems reasonable to provide a more helpful indicator. I would also like to make it easy for applications to control how a missing value is ultimately handled, so I don't like a "magic" string.

I propose adding an exported error variable that is appended and can be checked by a downstream logger if a different behavior is desired.

```go
var ErrMissingValue interface{} = errors.New("(MISSING)")
```

Issue #86, opened 2015-07-29
Panic when logging nil pointer to error with JSON formatter


husio:
When logging, every value is checked if it implements error interface. When positive, Error method is called. This may cause panic when passed value is `nil` pointer to structure that use it's internal state to produce Error method optput.

peterbourgon:
Does this test case capture the behavior you describe?

```go
type Foo struct { 
    msg string 
}

func (f *Foo) Error() { 
    return fmt.Errorf("%s", f.msg) 
}

func main() {
    var f *Foo // nil
    logger.Log("foo", f)
}
```

peterbourgon:
Oh, I see the PR now. Disregard my question.

ChrisHines:
This JSON logging code has other insidious nil panic issues. For example this test will panic as well:

```go
type stringStringer string

func (s stringStringer) String() string {
	return string(s)
}

func TestJSONLoggerNilPointers(t *testing.T) {
	buf := &bytes.Buffer{}
	logger := log.NewJSONLogger(buf)
	if err := logger.Log((*stringStringer)(nil), "v"); err != nil {
		t.Fatal(err)
	}
	if want, have := `{"null":"v"}`+"\n", buf.String(); want != have {
		t.Errorf("want %#v, have %#v", want, have)
	}
}
```

I have dealt with the same issue maintaining log15. I found that recovering the panic is the best approach. See also: https://github.com/inconshreveable/log15/issues/46.

I will add some comments regarding implementation on the PR.

PR #92, opened 2015-08-07
No log panics.

ChrisHines:
Fixes issues #84, #86 and supersedes PR #85.

### DIFF
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -22,7 +22,7 @@ func (l *jsonLogger) Log(keyvals ...interface{}) error {
 	m := make(map[string]interface{}, n)
 	for i := 0; i < len(keyvals); i += 2 {
 		k := keyvals[i]
-		v := ErrMissingValue
+		var v interface{} = ErrMissingValue
 		if i+1 < len(keyvals) {
 			v = keyvals[i+1]
 		}
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -17,12 +17,15 @@ func NewJSONLogger(w io.Writer) Logger {
 }
 
 func (l *jsonLogger) Log(keyvals ...interface{}) error {
-	if len(keyvals)%2 == 1 {
-		panic("odd number of keyvals")
-	}
-	m := make(map[string]interface{}, len(keyvals)/2)
+	n := (len(keyvals) + 1) / 2 // +1 to handle case when len is odd
+	m := make(map[string]interface{}, n)
 	for i := 0; i < len(keyvals); i += 2 {
-		merge(m, keyvals[i], keyvals[i+1])
+		k := keyvals[i]
+		v := ErrMissingValue
+		if i+1 < len(keyvals) {
+			v = keyvals[i+1]
+		}
+		merge(m, k, v)
 	}
 	return json.NewEncoder(l.Writer).Encode(m)
 }
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"reflect"
 )
 
 type jsonLogger struct {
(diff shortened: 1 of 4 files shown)


ChrisHines:
@peterbourgon & @husio, please take a look.

log/json_logger.go
```diff
@@ -17,29 +18,59 @@ func NewJSONLogger(w io.Writer) Logger {
...
+	}()
+	s = str.String()
+	return
+}
+
+func safeError(err error) (s interface{}) {
+	defer func() {
+		if panicVal := recover(); panicVal != nil {
+			if v := reflect.ValueOf(err); v.Kind() == reflect.Ptr && v.IsNil() {
+				s = nil
```
peterbourgon: Is it OK to interpret a nil reference to an error type as a nil error? Is it preferable to "NULL", as above?

log/json_logger_test.go
```diff
@@ -10,16 +10,67 @@ import (
...
+
+func TestJSONLoggerMissingValue(t *testing.T) {
+	t.Parallel()
+	buf := &bytes.Buffer{}
+	logger := log.NewJSONLogger(buf)
+	if err := logger.Log("k"); err != nil {
+		t.Fatal(err)
+	}
+	if want, have := `{"k":"(MISSING)"}`+"\n", buf.String(); want != have {
+		t.Errorf("\nwant %#v\nhave %#v", want, have)
```
peterbourgon: Are the newlines necessary here?

log/log.go
```diff
@@ -15,6 +18,10 @@ type Logger interface {
 	Log(keyvals ...interface{}) error
 }
 
+// ErrMissingValue may be appended to keyval slices with an odd length to
+// populate the missing value.
+var ErrMissingValue interface{} = errors.New("(MISSING)")
```
peterbourgon: - Right now it reads as if users of the package should do the appending, when actually it's the package itself. Can we change the wording, e.g. "ErrMissingValue may be appended by package log to keyval slices with an odd length, to populate the missing value"? — Or, will users even need to introspect for this value? Could this var be unexported?
- Is the explicit interface{} necessary?

peterbourgon:
:+1:

log/json_logger.go
```diff
@@ -17,29 +18,59 @@ func NewJSONLogger(w io.Writer) Logger {
...
 	}
 	if x, ok := v.(error); ok {
-		v = x.Error()
+		v = safeError(x)
 	}
 	dst[key] = v
-	return dst
+}
+
+func safeString(str fmt.Stringer) (s string) {
```
Mistobaan: is this really necessary? Feels very heavy to do this for each key

log/log.go
```diff
@@ -15,6 +18,10 @@ type Logger interface {
 	Log(keyvals ...interface{}) error
 }
 
+// ErrMissingValue may be appended to keyval slices with an odd length to
```
Mistobaan: rephrase suggestion: ErrMissingValue is appended to keyvals slices with odd length to substitute the missing value

log/log.go
Mistobaan: No is not http://play.golang.org/p/-fGdzvv9Nu

log/json_logger.go
ChrisHines: Good question. We only need it because of the special handling of `fmt.Stringer`. We could eliminate that case and just let `fmt.Sprint` handle them. The down side is that `fmt.Sprint` will return `"<nil>"` instead of `"NULL"`, and the angle brackets will get escaped in the JSON output, which is pretty ugly (http://play.golang.org/p/ap0GT3KO0l).

log/json_logger.go
ChrisHines: I believe it is ok. Also note that above is dealing with keys, and this is dealing with values.

log/json_logger_test.go
ChrisHines: Not necessary, but rather helpful when the test fails IMO. It creates vertical alignment that human pattern recognition deals with very efficiently, especially when the message is long and noisy as with JSON.

Without:
```
json_logger_test.go:32: want "{\"k\":\"(MISSING)\"}\n", have "{\"k\":\"(MISING)\"}\n"
```

With:
```
json_logger_test.go:32: 
	want "{\"k\":\"(MISSING)\"}\n"
	have "{\"k\":\"(MISING)\"}\n"
```

peterbourgon: Sure. Then can you make the other t.Error invocations consistent?

log/log.go
ChrisHines: Very nice. I will update the docs.

log/log.go
ChrisHines: That was misguided. I will remove.

log/json_logger_test.go
ChrisHines: Yes.

log/json_logger.go
Mistobaan: what about handling that special case and just use `fmt.Sprint`? http://play.golang.org/p/NnMLtszxRE

ChrisHines: I'm not a fan of that approach. It is slower and relies on undocumented behavior. Note that `fmt.Sprint` eventually uses the same technique as `safeString` to produce `"<nil>"`. See: 

https://github.com/golang/go/blob/go1.4.2/src/fmt/print.go#L678
and
https://github.com/golang/go/blob/go1.4.2/src/fmt/print.go#L624

ChrisHines:
@peterbourgon I just realized that my last two commits were done with the wrong email address attached. I would like to rewrite them. Any objections? I know how to do it on my side, any suggestions for handling the github side? Should I force push?

peterbourgon:
Yes, it's always fine to force push to a branch.

PR #94, opened 2015-08-11
Examples and documentation

peterbourgon:
This will greatly expand the example coverage, and greatly improve documentation.


examples/README.md
```diff
@@ -0,0 +1,370 @@
...
+```go
+type loggingMiddleware struct{
+	logger log.Logger
+	StringService
+}
+
+
+func (mw loggingMiddleware) Uppercase(s string) (output string, err error) {
+	defer func(begin time.Time) {
+		logger.Log(
```
xla: Do you want to refer to `mv.logger` here? Same in the `mv.Count` method.

ChrisHines:
@peterbourgon I've been too busy to give all of this a full read, but keep it coming, this looks awesome.

peterbourgon:
The addsvc needs to be retooled, but there's enough here that we can merge, I think.

PR #95, opened 2015-08-12
Add Middleware composition functions

shore:
Hi,
We like to reuse our middleware functions for various endpoints and like to put them in named functions.  We have some composition functions that make this easy.  In particular, we can compose individual middleware functions into chains, or we can compose more middleware functions onto the end of a chain.  This also creates fewer anonymous functions than rewrapping all middleware for each endpoint.

http://play.golang.org/p/MQ5iTyr789

ReduceMW is analogous to the reduce / foldl functions.  The composer is written such that the first MW function in the composition is the first one called, but it can trivially be replaced / altered to reverse the order if someone prefered the first function in the composition to instead be closest to the endpoint in the recursion.

### DIFF
--- a/endpoint/endpoint.go
+++ b/endpoint/endpoint.go
@@ -13,6 +13,26 @@ type Endpoint func(ctx context.Context, request interface{}) (response interface
 // Middleware is a chainable behavior modifier for endpoints.
 type Middleware func(Endpoint) Endpoint
 
+type MWComposer func(Middleware, Middleware) Middleware
+
+func ComposeMW(m1, m2 Middleware) Middleware {
+	return func(e Endpoint) Endpoint {
+		return m2(m1(e))
+	}
+}
+
+func ReduceMW(c MWComposer, m Middleware, mw ...Middleware) Middleware {
+	l := len(mw)
+	if l > 1 {
+		// merge last two elements in list, recurse
+		return ReduceMW(c, m, append(mw[:l-2], c(mw[l-1], mw[l-2]))...)
+	} else if l == 1 {
+		return c(mw[0], m)
+	} else {
+		return m
+	}
+}
+
 // ErrBadCast indicates an unexpected concrete request or response struct was
 // received from an endpoint.
 var ErrBadCast = errors.New("bad cast")


peterbourgon:
Hi, thanks for the contribution! I'm in favor of adding a helper type to chain middlewares, to avoid having to do it the manual way. But I'm not immediately convinced by your API, and I definitely wouldn't bake in an abbreviation like "MW" :) Have you seen https://github.com/justinas/alice? What do you think of that model? 

```go
// Current
var e endpoint.Endpoint
e = myEndpoint
e = inner(param)(e)
e = outer(key, val)(e)

// Proposed e.g.
e := endpoint.Chain(outer(key, val), inner(param), myEndpoint)
```

anarcher:
+1 I like it like `Ailce`.  In my very simple code,It looks like `ChainMiddeware` :-) 

```
func Chain(middewares ...endpoint.Middleware) func(endpoint.Endpoint) endpoint.Endpoint {
    return func(e endpoint.Endpoint) endpoint.Endpoint {
        for i := len(middewares) - 1; i >= 0; i-- { // reverse
            e = middewares[i](e) 
        }
        return e
    }
}

mw1 := outer(key,val)
mw2 := inter(params)
e1 := endpoint.Chain(mw1,mw2)(e1)
e2 := endpoint.Chain(mw1,mw2)(e2)
```

shore:
I'm not particularly attached to function names or recursion vs. iteration.  I do think that Chain should not be passed an endpoint, rather it should simply return a Middleware (chain) that can subsequently be applied to multiple endpoints (as in @anarcher's sample function) -- there's no point in rebuilding the same chain for multiple endpoints.

Also, I think Chain is better written to require at least one Middlware.  There's no reason to call Chain() if no Middleware is being passed, and if none are passed, there's no good return value without also returning an error (as the wrapper around the endpoint does nothing useful in that case).

How about this:

```go
func Chain(outer Middleware, others ...Middleware) Middleware {
	return func(e Endpoint) Endpoint {
		for i := len(others) - 1; i >= 0; i-- { // reverse
			e = others[i](e)
		}
		return outer(e)
	}
}
```

anarcher:
@shore It's better than my simple func.  :+1: :-)

peterbourgon:
Closed via #96! Thanks for the suggestion!

PR #96, opened 2015-08-14
Adaptation of @shore's proposal from #95

peterbourgon:
@shore @anarcher What do you think about this? Is it what you have in mind?


peterbourgon:
The build failures appear to be from a change in the go-hystrix dependency. I'll investigate and fix that in master, and rebase.

anarcher:
Yes, I like this feature! :+1:

shore:
Looks great!

PR #100, opened 2015-08-17
Add Consul publisher

xla:
**open questions**

- [x] Is it allowed to return stale endpoints?
- [x] Are publishers expected to return `ip:port` or are `hostnames` equally valid?
- [x] Does it suffice to only be able to filter by one tag?

**todos**

- [x] only emit endpoints when different
- [ ] documentation
- [ ] example
- [x] tests

***
*The commits in this branch will be cleaned up in time and likely stashed into a single coherent change-set.*

### DIFF
--- a/loadbalancer/consul/client.go
+++ b/loadbalancer/consul/client.go
@@ -1,0 +1,30 @@
+package consul
+
+import consul "github.com/hashicorp/consul/api"
+
+// Client is a wrapper around the Consul API.
+type Client interface {
+	Service(service string, tag string, queryOpts *consul.QueryOptions) ([]*consul.ServiceEntry, *consul.QueryMeta, error)
+}
+
+type client struct {
+	consul *consul.Client
+}
+
+// NewClient returns an implementation of the Client interface expecting a fully
+// setup Consul Client.
+func NewClient(c *consul.Client) Client {
+	return &client{
+		consul: c,
+	}
+}
+
+// GetInstances returns the list of healthy entries for a given service filtered
+// by tag.
+func (c *client) Service(
+	service string,
+	tag string,
+	opts *consul.QueryOptions,
+) ([]*consul.ServiceEntry, *consul.QueryMeta, error) {
+	return c.consul.Health().Service(service, tag, true, opts)
+}
--- a/loadbalancer/consul/publisher.go
+++ b/loadbalancer/consul/publisher.go
@@ -1,0 +1,175 @@
+package consul
+
+import (
+	"fmt"
+
+	consul "github.com/hashicorp/consul/api"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
+)
+
+const defaultIndex = 0
+
+// Publisher yields endpoints for a service in Consul. Updates to the service
+// are watched and will update the Publisher endpoints.
+type Publisher struct {
+	cache      *loadbalancer.EndpointCache
+	client     Client
+	logger     log.Logger
+	service    string
+	tags       []string
+	endpointsc chan []endpoint.Endpoint
+	quitc      chan struct{}
+}
+
+// NewPublisher returns a Consul publisher which returns Endpoints for the
+// requested service. It only returns instances for which all of the passed
+// tags are present.
+func NewPublisher(
+	client Client,
+	factory loadbalancer.Factory,
+	logger log.Logger,
+	service string,
+	tags ...string,
+) (*Publisher, error) {
+	logger = log.NewContext(logger).With("component", "Consul Publisher")
+
+	p := &Publisher{
+		cache:   loadbalancer.NewEndpointCache(factory, logger),
+		client:  client,
+		logger:  logger,
+		service: service,
+		tags:    tags,
+		quitc:   make(chan struct{}),
+	}
+
+	is, index, err := p.getInstances(defaultIndex)
+	if err != nil {
+		return nil, err
+	}
+
+	p.cache.Replace(is)
+
+	go p.loop(index)
+
+	return p, nil
+}
+
+// Endpoints implements the Publisher interface.
+func (p *Publisher) Endpoints() ([]endpoint.Endpoint, error) {
+	return p.cache.Endpoints()
+}
+
+// Stop terminates the publisher.
+func (p *Publisher) Stop() {
+	close(p.quitc)
+}
+
+func (p *Publisher) loop(lastIndex uint64) {
+	var (
+		errc = make(chan error, 1)
+		resc = make(chan response, 1)
+	)
+
+	for {
+		go func() {
+			is, index, err := p.getInstances(lastIndex)
+			if err != nil {
+				errc <- err
+				return
+			}
+
+			resc <- response{
+				index:     index,
+				instances: is,
+			}
+		}()
+
+		select {
+		case err := <-errc:
+			p.logger.Log("service", p.service, "err", err)
+		case res := <-resc:
+			p.cache.Replace(res.instances)
+			lastIndex = res.index
+		case <-p.quitc:
+			return
+		}
+	}
+}
+
+func (p *Publisher) getInstances(lastIndex uint64) ([]string, uint64, error) {
+	tag := ""
+
+	if len(p.tags) > 0 {
+		tag = p.tags[0]
+	}
+
+	entries, meta, err := p.client.Service(
+		p.service,
+		tag,
+		&consul.QueryOptions{
+			WaitIndex: lastIndex,
+		},
+	)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// If more than one tag is passed we need to filter it in the publisher until
+	// Consul supports multiple tags[0].
+	//
+	// [0] https://github.com/hashicorp/consul/issues/294
+	if len(p.tags) > 1 {
+		entries = filterEntries(entries, p.tags[1:]...)
+	}
+
+	return makeInstances(entries), meta.LastIndex, nil
+}
+
+// response is used as container to transport instances as well as the updated
+// index.
+type response struct {
+	index     uint64
+	instances []string
+}
+
+func filterEntries(entries []*consul.ServiceEntry, tags ...string) []*consul.ServiceEntry {
+	var es []*consul.ServiceEntry
+
+ENTRIES:
+	for _, entry := range entries {
+		ts := make(map[string]struct{}, len(entry.Service.Tags))
+
+		for _, tag := range entry.Service.Tags {
+			ts[tag] = struct{}{}
+		}
+
+		for _, tag := range tags {
+			if _, ok := ts[tag]; !ok {
+				continue ENTRIES
+			}
+		}
+
+		es = append(es, entry)
+	}
+
+	return es
+}
+
+func makeInstances(entries []*consul.ServiceEntry) []string {
+	is := make([]string, len(entries))
+
+	for i, entry := range entries {
+		addr := entry.Node.Address
+
+		if entry.Service.Address != "" {
+			addr = entry.Service.Address
+		}
+
+		is[i] = fmt.Sprintf("%s:%d", addr, entry.Service.Port)
+	}
+
+	return is
+}
--- a/loadbalancer/consul/publisher_test.go
+++ b/loadbalancer/consul/publisher_test.go
@@ -1,0 +1,207 @@
+package consul
+
+import (
+	"io"
+	"testing"
+
+	consul "github.com/hashicorp/consul/api"
+	"golang.org/x/net/context"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/log"
+)
+
+var consulState = []*consul.ServiceEntry{
+	&consul.ServiceEntry{
+		Node: &consul.Node{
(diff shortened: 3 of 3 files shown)


peterbourgon:
> Is it allowed to return stale endpoints?

Is stale a term of art in Consul? In the normal case, if the publisher _knows_ an endpoint is broken, it shouldn't emit it via Endpoints. If a publisher _suspects_ an endpoint _might_ be broken, then it's a gray area, perhaps best resolved with a configuration setting during construction, or deferring to best practice of the SD system.

> Are publishers expected to return ip:port or are hostnames equally valid?

Is a hostname sufficient to create a connection to the remote service? If yes, then it's fine. But that means you're carrying the service along implicitly—I can't think of a good use case for that. Do you have one?

> Does it suffice to only be able to filter by one tag?

Per our discussion yesterday, I'm guessing not.

EcofitBrendan:
I've been integrating this code into our migration to using a full publisher/loadbalancer implementation with go-kit and have been very happy with it! One thing I have noticed about the implementation:

When reading a service from a Consul health API check there are actually two potential fields the address of a service can appear in, either **ServiceEntry.Node.Address** or **ServiceEntry.Service.Address**.

The reason these two can differ is as a result of the configuration of a particular Consul datacenter. The service address field is populated if a specific address has been specified for a service configuration on a given node. This is usually a result of loading a service into Consul that cannot have a Consul agent run on its host OS. This results in the Node address referring to the hosting agent, and the service address referring to the actual resolved service location.

In our use cases, we treat the service address as an override to the node address, as it is empty when not in use. This is helpful in development environments when you want to run a local Consul server pointing to external dependencies.

peterbourgon:
@EcofitBrendan Great to hear! Would you be willing to contribute your implementation as a PR when you're happy with it?

EcofitBrendan:
@peterbourgon Definitely, I should have something functional available over the next few days.

peterbourgon:
@EcofitBrendan I'm going to pick this up now, unless you have something ready to go?

xla:
This is ready for review, in the meantime I'll work on the README with examples and high-level documentation. As soon as approved I'm going to stash the changes.

@EcofitBrendan Thanks for pointing out this intricacy. I added a [test case](https://github.com/xla/kit/blob/consul-publisher/loadbalancer/consul/publisher_test.go#L128) to make sure the publisher accounts for it.

/cc @dlsniper @peterbourgon @tsenart

xla:
Currently not covered is the case for an update to the endpoint. This might be better suited for integration tests. Happy to cover it here as well, if deemed important for acceptance.

loadbalancer/consul/client.go
```diff
@@ -0,0 +1,30 @@
...
+
+// NewClient returns an implementation of the Client interface expecting a fully
+// setup Consul Client.
+func NewClient(c *consul.Client) Client {
+	return &client{
+		consul: c,
+	}
+}
+
+// GetInstances returns the list of entries for a given service filtered by
```
tsenart: This comment is outdated.

loadbalancer/consul/client.go
```diff
@@ -0,0 +1,30 @@
+package consul
```
tsenart: A `doc.go` would be nice!

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
+	consul "github.com/hashicorp/consul/api"
+)
+
+const defaultIndex = 0
+
+// Publisher yields endpoints for the given service filtered by tags.
```
tsenart: What given service? This is a struct, not a method.

tsenart: Maybe `a given service` would ring better,

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+	return p.cache.Endpoints(), nil
+}
+
+// Stop terminates the publisher.
+func (p *Publisher) Stop() {
+	close(p.quitc)
+}
+
+func (p *Publisher) loop(
+	lastIndex uint64,
```
tsenart: Why is this broken up? It doesn't seem like it would cross 80 columns.

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+) {
+	var (
+		errc = make(chan error)
+		resc = make(chan response)
+	)
+
+	for {
+		go func() {
```
tsenart: I don't see a need for the use of concurrency here. Why do we need the go-routine? The only thing you need so select on here is the `s.quitc` channel as far as I understand it.

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+		tag,
+		&consul.QueryOptions{
+			WaitIndex: lastIndex,
+		},
+	)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if len(p.tags) > 1 {
```
tsenart: Shouldn't this be `> 0`?

tsenart: If not, a comment explaining what's going on would be nice.

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+	index     uint64
+	instances []string
+}
+
+func filterEntries(entries []*consul.ServiceEntry, tags ...string) []*consul.ServiceEntry {
+	es := []*consul.ServiceEntry{}
+
+ENTRIES:
+	for _, entry := range entries {
+		fmt.Println(entry.Service.ID)
```
tsenart: Debugging traces?

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+
+// response is used as container to transport instances as well as the updated
+// index.
+type response struct {
+	index     uint64
+	instances []string
+}
+
+func filterEntries(entries []*consul.ServiceEntry, tags ...string) []*consul.ServiceEntry {
+	es := []*consul.ServiceEntry{}
```
tsenart: `var es []*consul.ServiceEntry` won't allocate anything if it isn't used.

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+
+ENTRIES:
+	for _, entry := range entries {
+		fmt.Println(entry.Service.ID)
+		tagMap := map[string]struct{}{}
```
tsenart: This is actually a set, and we know its exact size: `tagSet := make(map[string]struct{}, len(entry.Service.Tags))`

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		es = append(es, entry)
+	}
+
+	return es
+}
+
+func makeInstances(entries []*consul.ServiceEntry) []string {
+	is := []string{}
```
tsenart: `is := make([]string, len(entries))

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+
+	for _, entry := range entries {
+		addr := entry.Node.Address
+
+		if entry.Service.Address != "" {
+			addr = entry.Service.Address
+		}
+
+		is = append(is, fmt.Sprintf("%s:%d", addr, entry.Service.Port))
```
tsenart: https://golang.org/pkg/net/#JoinHostPort

loadbalancer/consul/publisher.go
xla: Consul implements subscriptions to state changes via a [Blocking Queries](https://consul.io/docs/agent/http.html). Depending how the consul client and more specifically its timeouts are configured the to `getInstances` could block for minutes. I see this behaviour conflicting with the need of a swift teardown when `Stop` is invoked.

loadbalancer/consul/publisher.go
xla: `entry.Service.Port` is an `int`.

loadbalancer/consul/publisher.go
tsenart: Makes sense, thanks for clarifying. We should probably add timeouts to those blocking queries though.

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,177 @@
...
+		select {
+		case err := <-errc:
+			p.logger.Log("service", p.service, "err", err)
+		case res := <-resc:
+			p.cache.Replace(res.instances)
+
+			lastIndex = res.index
+		case <-p.quitc:
+			return
+		default:
```
tsenart: Why the default case? Don't we want to block until we get a response?

loadbalancer/consul/publisher.go
xla: I always assume it's safe to push that responsibility to the client [configuration](https://godoc.org/github.com/hashicorp/consul/api#Config) and have the control left with the caller.

loadbalancer/consul/client.go
```diff
@@ -0,0 +1,30 @@
...
+
+import consul "github.com/hashicorp/consul/api"
+
+// Client is a wrapper around the Consul API.
+type Client interface {
+	Service(string, string, *consul.QueryOptions) ([]*consul.ServiceEntry, *consul.QueryMeta, error)
```
peterbourgon: Variable names for these method parameters, please?

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
+package consul
+
+import (
+	"fmt"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
+	consul "github.com/hashicorp/consul/api"
```
peterbourgon: Please move this import to its own group between stdlib and go-kit

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+			if err != nil {
+				errc <- err
+				return
+			}
+
+			resc <- response{
+				index:     index,
+				instances: is,
+			}
+		}()
```
peterbourgon: This goroutine will leak unless you make resc with a buffer of 1, or wrap it in a select, e.g.

```go
select {
    case <-quitc:
    case resc <- response{...}
}
```

loadbalancer/consul/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+
+		select {
+		case err := <-errc:
+			p.logger.Log("service", p.service, "err", err)
+		case res := <-resc:
+			p.cache.Replace(res.instances)
+
```
peterbourgon: rm this \n?

loadbalancer/consul/publisher.go
xla: Buffer of 1 seems tempting, which way of expressing it is preferred? Any precedence in the codebase?

peterbourgon: > Buffer of 1 seems tempting

SGTM.

> precendence?

[circuitbreaker/hystrix.go#L20](https://github.com/go-kit/kit/blob/b50c5697c253c3ea01687acbef12ac1b2d75a3af/circuitbreaker/hystrix.go#L20); [loadbalancer/retry.go#L25-26](https://github.com/go-kit/kit/blob/b50c5697c253c3ea01687acbef12ac1b2d75a3af/loadbalancer/retry.go#L25-L26).

loadbalancer/consul/publisher_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+import (
+	"io"
+	"testing"
+
+	"golang.org/x/net/context"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/log"
+	consul "github.com/hashicorp/consul/api"
```
peterbourgon: Please group with net/context import.

peterbourgon:
Haven't audited for correctness or tested with a real Consul, but code structure LGTM. Do you want me to test in a live environment, or are you confident to merge?

xla:
@peterbourgon Would appreciate a real world test, if it doesn't mean too much time consumption for you.

loadbalancer/consul/publisher.go
xla: Changed in f1ac217 for `errc` as well.

peterbourgon:
I tested this with a live Consul cluster and it works well. I am happy to merge it. Is there anything else you'd like to do before I click the green button?

xla:
@peterbourgon Thanks for taking the time. I squashed the commits and am fine with merging this one. Will work on examples and documentation independently. Unfortunately the tests are failing since rebasing against master, it looks to be a change in the `go-etcd` dependency.

peterbourgon:
Yep, I will wait until [etcd #246](https://github.com/coreos/go-etcd/issues/246) is fixed and we can get a greenlight on CI before merging.

xla:
@peterbourgon Full ack.

peterbourgon:
Sorry, can you rebase on master and fix the build error?

xla:
@peterbourgon Done.

peterbourgon:
Woop woop

PR #102, opened 2015-08-19
Add tracing/appdash

anarcher:
Hi, I'm not familiar with`appdash` and I'm not used to `appdash`. 
But This PR could be a starting point, or use for inspiration.
Thanks.

### DIFF
--- a/tracing/appdash/appdash.go
+++ b/tracing/appdash/appdash.go
@@ -1,0 +1,56 @@
+package appdash
+
+import (
+	"github.com/go-kit/kit/endpoint"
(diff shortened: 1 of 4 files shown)


peterbourgon:
Wow, very interesting! Thank you very much for the contribution. I'd like to stand up an Appdash server and try this out. But I'll only have time for that in about a week. Please feel free to make updates in the meantime.

anarcher:
Thanks! I will make some updates to fix the weird things and to add helper functions (such as `ToRequest` for clients and `FromHttp` extracts HTTP Header'). :-)

PR #104, opened 2015-08-20
Implement a colorful TerminalLogger

tgulacsi:
Includes a copy of gopkg.in/inconshreveable/log15.v2/term, which is copied from github.com/Sirupsen/logrus.

### DIFF
--- a/log/term/LICENSE
+++ b/log/term/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2014 Simon Eskildsen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/log/term/terminal_appengine.go
+++ b/log/term/terminal_appengine.go
@@ -1,0 +1,13 @@
+// Based on ssh/terminal:
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build appengine
+
+package term
+
+// IsTty always returns false on AppEngine.
+func IsTty(fd uintptr) bool {
+	return false
+}
--- a/log/term/terminal_darwin.go
+++ b/log/term/terminal_darwin.go
@@ -1,0 +1,12 @@
+// Based on ssh/terminal:
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package term
+
+import "syscall"
+
+const ioctlReadTermios = syscall.TIOCGETA
+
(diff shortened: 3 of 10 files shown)


ChrisHines:
@tgulacsi I've spent some time reviewing this code and there are several details I'd like to see improved. But before I get into the details, I would like to discuss the general approach. I have a feeling that `TerminalLogger` in its current form is doing too much and I'd like to explore other designs.

My biggest concerns are the awkwardness of the `NewTerminalLogger` constructor, and the way that the `Log` method is tightly coupled to the existence of specific keys—keys which log15 ensures are present but go-kit log does not.

Rather than just copy the behavior of the log15 code, I would like to take a step back and discuss what we can do that is more modular and fits better with the rest of the Go kit design philosphy.

Breaking the functionality into separate concerns I see three mostly separate features implemented by `TerminalLogger` that perhaps should be factored into orthogonal pieces.

1. An alternate log event format ([level] [time] msg k1=v1 k2=v2 ...)
1. Choosing between two `Logger`s based on a precicate (in this case, `IsTty`)
1. Injecting color codes based on a color mapping.

The first concern, when taken separately boils down to formatting some keyvals differently than others with a secondary concern of enforcing certain keys are present in every log event. Perhaps this should be broken into two concerns. More thought required here.

If we provide a useful `term.IsTty` function, the second concern is easily accomplished by an `if` statement when wiring up logging in an app's main package:
```go
var w io.Writer
var logger log.Logger
if term.IsTty(w) {
    logger = log.NewTerminalLogger(w)
} else {
    logger = log.NewLogfmtLogger(w)
}
```

The third concern poses a more interesting challenge. The first simplification I want to make is to color entire log events instead of just the level field. Then we would only need to put color codes at the beginning and ending of the line. We could leave the rest of the format produced by another `Logger` unmodified. Then this `ColorLogger` could wrap any log record format.

Allowing a configurable color mapping seems like a good way to let apps chose color based on what is important to them, so we could accept a `func(keyvals ...interface{}) (color int)` mapping function. We could provide a default mapping that works well with package `log/levels`, but I'm sure others will come up with creative ways to use this feature for coloring log lines based on all kinds of criteria.

Finally, we will need a way for `ColorLogger` to:

1. Receive a structured log event (i.e. `keyvals ...interface{}),
1. Use the mapping to get a color,
1. Call another `Logger` to format `keyvals` into a `[]byte`,
1. Inject the color code sequences,
1. And finally write the result to an `io.Writer`.

The bit about wrapping another `Logger`, but getting access to its formatted output before sending it to an `io.Writer` requires some thought. It feels vaguely similar to functionality of `net/http.Hijacker`, but I'm just thinking out loud on that point.

tgulacsi:
@ChrisHines good thoughts. I've feeling the same about term.IsTty: just provide the functionality, and document an example.

About coloring logs: I don't know yet how to inject the colorings in the correct positions when serializing the log. Maybe a ColoredKey, ColoredValue fmt.Stringer/fmt.Formatter enclosing the real key/value?

About log15: I do want to copy the LvlFilterLogger functionality, too :)
That'd need to inspect the structured logs, assume which key contains the level and what value is the below the threshold - so here we meet the need to assume about the structure of keyvals, too.
Maybe define a default mapping for all known keys, and use it in level, and in the FilterLogger and ColorLogger?

ChrisHines:
@tgulacsi I think the `ColoredKey`, and `ColoredValue` ideas are worth exploring. I had a similar thought after I posted my comments above.

I too want `LvlFilterLogger` eventually. I use the log15 equivalent in all my apps. My current thinking is that such a feature belongs in package `log/levels`.

tgulacsi:
I've implemented a ColorLogger in term, which wraps a log.Logger: https://github.com/tgulacsi/kit/tree/colorlog

But a self-concious, safe logger will escape those escape codes...

So this is only half success, as it needs a special logger (TerminalLogger) to be useful.

ChrisHines:
What if a `ColoredValue` returned non-colored text from its `String` method, and had a `ColoredString` method that `TerminalLogger` used via doing a type check against the following interface?

```go
type ColoredStringer interface {
    ColoredString() string
}
```

tgulacsi:
It's ok for me, and it will be ok with TerminalLogger, my only problem is
that we cannot use this any standerd Logger as ColorLogger's backend.

What about specializing the Writer:

type KeyValueWriter interface{
  WriteKey([]byte) (int, error)
  WriteValue([]byte) (int, error)
  Write([]byte) (int, error)
}

And specialize all loggers, to use WriteKey when writing a key, and use
WriteValue when a value, and Write for other times?

This seems to big hurdle to generalize the colorizer...


So I'd conclude that do ColorLogger as is, and TerminalLogger to use
ColoredString. That way any future logger could use this info, and use
ColorLogger.



---
https://onename.com/tgulacsi

2015-08-25 15:18 GMT+02:00 Chris Hines <notifications@github.com>:

ChrisHines:
I think `KeyValueWriter` is a non-starter. We don't want to go down the path of rewriting `encoding/json` for example.

If try to keep coloring and text formatting separate concerns then I don't think it will be practical to color individual keys or values. Most record encoders (logfmt, JSON, etc.) will typically not have hooks to inject (non-escaped) color codes mid-record. That is why I suggested we color entire records rather than individual keys or values. I'm not sure if that suggestion was overlooked or rejected.

PR #105, opened 2015-08-21
Modify all import paths to gopkg.in/kit.v0

tgulacsi:
If I import gopkg.in/kit.v0/metrics and
gopkg.in/kit.v0/metrics/prometheus, I get compile errors, as prometheus
uses metrics.Field from github.com/go-kit/kit, but I have metrics.Fields
from gopkg.in/kit.v0/metrics.

### DIFF
--- a/addsvc/add.go
+++ b/addsvc/add.go
@@ -3,9 +3,9 @@ package main
 import (
 	"golang.org/x/net/context"
 
-	"github.com/go-kit/kit/addsvc/reqrep"
-	"github.com/go-kit/kit/endpoint"
-	"github.com/go-kit/kit/log"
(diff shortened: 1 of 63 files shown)


ChrisHines:
I don't believe we have officially adopted gopkg.in for Go kit. Correct me if I am wrong @peterbourgon.

peterbourgon:
There are no plans to adopt gopkg.in at this point. As enumerated in the README, Go kit consumers are responsible for vendoring Go kit into their binaries, and managing that dependency using their vendoring tool of choice.

peterbourgon:
I'll make the vendoring expectation more explicit.

PR #106, opened 2015-08-24
Keep log.Context.Log stack depth consistent.

ChrisHines:
Change `log.Context` methods to take pointer receivers to prevent the compiler from introducing generated functions to dereference pointers when calling methods via a `log.Logger` interface.

See comments in `log/log.go` and the new `TestContextStackDepth` for more details.


ChrisHines:
@peterbourgon This fixes the issue highlighted in d95da40cb7b5c0b1815fc4961d1b8df38843f577.

peterbourgon:
Thanks for the detailed comment. :+1:

PR #108, opened 2015-08-25
Update examples README.md

twanies:
Updated function declarations that are not compatible with the transport/http.DecodeFunc and transport/http.EncodeFunc


peterbourgon:
Thank you.

PR #109, opened 2015-08-25
Implement term.ColorLogger

tgulacsi:
This needs a Hijack method for Logger - this is implemented on Logfmt.
This logger colorizes whole records, for the first matching key/value.

This way you can colorize the record if it contains a given key, or if it contains a non-nil error, for example.
Or colorize levels differently.

### DIFF
--- a/log/term/colorlogger.go
+++ b/log/term/colorlogger.go
@@ -31,14 +31,9 @@ func (c FgBgColor) IsZero() bool {
 	return c.Fg == NoColor && c.Bg == NoColor
 }
 
-type ColorOption struct {
-	Key   FgBgColor
-	Value func(interface{}) FgBgColor
-}
-
 var _ = log.Logger((*colorLogger)(nil))
 
-// NewColorLogger returns a log.Logger which prouces nice colored logs.
+// NewColorLogger returns a log.Logger which produces nice colored logs.
 // It colors whole records based on the FgBgColor returned by the color function.
 //
 // For example for such a function, see LevelColor.
@@ -61,16 +56,21 @@ var _ = log.Logger((*colorLogger)(nil))
 //	logger.Log("c", "c is uncolored value", "err", nil)
 //	logger.Log("c", "c is colored 'cause err colors it", "err", errors.New("coloring error"))
 func NewColorLogger(logger log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger {
+	// FIXME(tgulacsi): I don't understand why can't I use log.Hijacker here...
+	hj, ok := logger.(interface {
+		Hijack(func(io.Writer) io.Writer)
+	})
+	if !ok {
+		return logger
+	}
 	cl := &colorLogger{
 		logger: logger,
 	}
-	if hj, ok := logger.(log.Hijacker); ok {
-		cl.color = color // otherwise, no coloring is possible!
-		hj.Hijack(func(w io.Writer) io.Writer {
-			cl.w = w
-			return cl
-		})
-	}
+	cl.color = color // otherwise, no coloring is possible!
+	hj.Hijack(func(w io.Writer) io.Writer {
+		cl.w = w
+		return cl
+	})
 	return cl
 }
 
@@ -78,6 +78,7 @@ type colorLogger struct {
 	logger log.Logger
 	color  func(keyvals ...interface{}) FgBgColor
 	w      io.Writer
+	mu     sync.RWMutex // protects w
 
 	actColor   FgBgColor
 	actColorMu sync.RWMutex // protects actColor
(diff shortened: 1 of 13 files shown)


ChrisHines:
There is a lot to review here so I am going to start with a few big things I'd like to see changed, and will hold off on other comments until we see the result of the bigger changes.

First, I think we should change the `NewColorLogger` signature to:
```go
func NewColorLogger(logger log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger
```
There are two changes here.

1. The logger parameter is now type `log.Logger`. This change requires using a type assertion in the implementation to get access to the `Hijack` method, but it also keeps variables in client code as type `log.Logger`. Also you can restore the return types of `NewLogfmtLogger` to `log.Logger`.
1. We replace the `opts` variadic parameter with a function that is more flexible and will simplify the implementation of `ColorLogger`. Now `ColorLogger` is doing one thing, coloring the logs; the `color` function decides what the colors to paint; and the wrapped `Logger` controls the textual encoding. Three nicely orthogonal pieces working together.

In concert with the above change we can change `HijackLogger` with
```go
type Hijacker interface {
    Hijack(func(io.Writer) io.Writer)
}
```
I don't think there is any reason for it to embed the `Logger` interface since it is an optional interface. We will require `Logger` in the parameter type, and assert internally that it implements `Hijacker`.

Update the PR if you agree that these are good changes. I think there will be much less to review after that.

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+
+type FgBgColor struct {
+	Fg, Bg Color
+}
+
+func (c FgBgColor) IsZero() bool {
+	return c.Fg == NoColor && c.Bg == NoColor
+}
+
+type ColorOption struct {
```
ChrisHines: ColorOption is not used and should be removed.

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+	Key   FgBgColor
+	Value func(interface{}) FgBgColor
+}
+
+var _ = log.Logger((*colorLogger)(nil))
+
+// NewColorLogger returns a log.Logger which prouces nice colored logs.
```
ChrisHines: s/prouces/produces/

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+// It colors whole records based on the FgBgColor returned by the color function.
+//
+// For example for such a function, see LevelColor.
```
ChrisHines: `LevelColor` does not exist.

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+		logger: logger,
+	}
+	if hj, ok := logger.(log.Hijacker); ok {
+		cl.color = color // otherwise, no coloring is possible!
+		hj.Hijack(func(w io.Writer) io.Writer {
+			cl.w = w
+			return cl
+		})
+	}
+	return cl
```
ChrisHines: Why not return `logger` when we cannot hijack?

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+//					return term.FgBgColor{Fg: term.White, Bg: term.Red}
+//				}
+//			}
+//			return term.FgBgColor{}
+//		})
+//
+//	logger.Log("c", "c is uncolored value", "err", nil)
+//	logger.Log("c", "c is colored 'cause err colors it", "err", errors.New("coloring error"))
+func NewColorLogger(logger log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger {
+	cl := &colorLogger{
```
ChrisHines: Move this inside the following `if` statement.

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+	n += m
+	m, err = l.w.Write([]byte("\x1b[0m"))
+	return n + m, err
+}
+
+func (l *colorLogger) Hijack(f func(io.Writer) io.Writer) {
+	l.w = f(l.w)
+}
+
+type ColoredValue struct {
```
ChrisHines: `ColoredValue` is not used and can be removed.

log/term/colorlogger.go
```diff
@@ -0,0 +1,160 @@
...
+	if cv.Bg == 0 {
+		return fmt.Sprintf("\x1b[%dm%v\x1b[0m", 30+cv.Fg, cv.Value)
+	}
+	if cv.Fg == 0 {
+		return fmt.Sprintf("\x1b[%dm%v\x1b[0m", 40+cv.Bg, cv.Value)
+	}
+	return fmt.Sprintf("\x1b[%dm\x1b[%dm%v\x1b[0m", 30+cv.Fg, 40+cv.Bg, cv.Value)
+}
+
+func AsString(v interface{}) string {
```
ChrisHines: `AsString` is only used in test code, move it to colorlogger_test.go and unexport it.

log/term/colorlogger_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	buf.Reset()
+	if err := logger.Log("a", 1, "err", errors.New("error")); err != nil {
+		t.Fatal(err)
+	}
+	if want, have := "\u001b[32m\u001b[48ma=1 err=error\n\u001b[0m", buf.String(); want != have {
+		t.Errorf("want %#v, have %#v", want, have)
+	}
+
+	buf.Reset()
```
ChrisHines: This section doesn't test anything related to color logging, remove it.

ChrisHines:
Aside from my inline comments above, I'm a little concerned about the concurrency story. We should think about it carefully. I have two concerns.

1. The `Hijacker` interface implementations added to the JSON and Logfmt Loggers introduces potential race conditions to those loggers.
1. The `actColorMu` in `colorLogger` seems awkward and it doesn't prevent potentially strange coloring behavior if the wrapped `Logger` is used directly while concurrently being used by `colorLogger`.

log/term/colorlogger_test.go
```diff
@@ -0,0 +1,108 @@
...
+	if err := logger.Log("std_map", map[int]int{1: 2}, "my_map", mymap{0: 0}); err != nil {
+		t.Fatal(err)
+	}
+	if want, have := "std_map=\""+logfmt.ErrUnsupportedValueType.Error()+"\" my_map=special_behavior\n", buf.String(); want != have {
+		t.Errorf("want %#v, have %#v", want, have)
+	}
+}
+
+func newColorLogger(t testing.TB, w io.Writer) log.Logger {
```
ChrisHines: parameter `t` is not used and can be removed.

ChrisHines:
@tgulacsi Thanks for the updates.

The bad news is that the current code really shows the problems with the `Hijacker` interface. That approach introduces too many edge cases to easily reason about and is too invasive of loggers that we want to wrap with `ColorLogger`.

The good news is that I have come up with an alternate approach that solves all of those problems. I have already coded it, so I'm going to start a new PR building on this one.

I will leave this PR open until we have reviewed the new one and agree it supersedes this one.

tgulacsi:
Ok, I'm curiously waiting for that!

---
https://onename.com/tgulacsi

2015-08-31 6:22 GMT+02:00 Chris Hines <notifications@github.com>:

ChrisHines:
@tgulacsi See PR #116.

peterbourgon:
Can this be closed in deference to #116?

tgulacsi:
Yes.

PR #115, opened 2015-08-30
[WIP] tracing/appdash

peterbourgon:
This is an extension of #102.


peterbourgon:
I think this is now too old to be useful. If we will pick this up again, it will be from scratch. Leaving the branch, though.

PR #116, opened 2015-08-31
Implement term.ColorLogger (take 2)

ChrisHines:
This is an evolution of #109 to improve the concurrency story.

The key change is to have `ColorLogger` completely own the wrapped formatting logger and thereby eliminate the possibility that some other part of the application potentially calls its `Log` method outside of the `ColorLogger`'s control and cause havoc.

To achieve this goal we change the signature of `NewColorLogger` to the following.
```go
func NewColorLogger(w io.Writer, newLogger func(io.Writer) log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger
```
It now accepts an `io.Writer` and a constructor function. It uses `newLogger` to create `Logger`s that write to buffers controlled by the `ColorLogger`. The `ColorLogger` adds color codes to the buffered results and writes the combined results to `w`.

Createing a `ColorLogger` now looks like this
```go
logger := term.NewColorLogger(os.Stdout, log.LogfmtLogger, term.LevelColor)
```

Furthermore, `log.Hijacker` and all of its implementations are eliminated. The concurrency question is handled by a `sync.Pool` that manages a free list of buffer-Logger pairs. Using a buffer pool reduces the amortized allocation cost of `ColorLogger` to zero and eliminates the `sync.Mutex`s from the previous version.

```
BenchmarkLogfmtLoggerSimple-4             500000              2508 ns/op             384 B/op          6 allocs/op
BenchmarkColorLoggerSimple-4              500000              2672 ns/op             384 B/op          6 allocs/op

BenchmarkLogfmtLoggerContextual-4         500000              3656 ns/op             560 B/op          9 allocs/op
BenchmarkColorLoggerContextual-4          300000              4270 ns/op             560 B/op          9 allocs/op
```

### DIFF
--- a/log/term/colorlogger.go
+++ b/log/term/colorlogger.go
@@ -1,6 +1,7 @@
 package term
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"sync"
@@ -21,7 +22,20 @@ const (
 	Cyan
 	White
 	Default
+
+	maxColor
 )
+
+var resetColorBytes = []byte("\x1b[0m")
+var fgColorBytes [][]byte
+var bgColorBytes [][]byte
+
+func init() {
+	for color := NoColor; color < maxColor; color++ {
+		fgColorBytes = append(fgColorBytes, []byte(fmt.Sprintf("\x1b[%dm", 30+color)))
+		bgColorBytes = append(bgColorBytes, []byte(fmt.Sprintf("\x1b[%dm", 40+color)))
+	}
+}
 
 type FgBgColor struct {
 	Fg, Bg Color
@@ -55,82 +69,70 @@ var _ = log.Logger((*colorLogger)(nil))
 //
 //	logger.Log("c", "c is uncolored value", "err", nil)
 //	logger.Log("c", "c is colored 'cause err colors it", "err", errors.New("coloring error"))
-func NewColorLogger(logger log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger {
-	// FIXME(tgulacsi): I don't understand why can't I use log.Hijacker here...
-	hj, ok := logger.(interface {
-		Hijack(func(io.Writer) io.Writer)
-	})
-	if !ok {
-		return logger
+func NewColorLogger(w io.Writer, newLogger func(io.Writer) log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger {
+	if color == nil {
+		panic("color func nil")
 	}
-	cl := &colorLogger{
-		logger: logger,
+	return &colorLogger{
+		w:             w,
+		newLogger:     newLogger,
+		color:         color,
+		bufPool:       sync.Pool{New: func() interface{} { return &loggerBuf{} }},
+		noColorLogger: newLogger(w),
 	}
-	cl.color = color // otherwise, no coloring is possible!
-	hj.Hijack(func(w io.Writer) io.Writer {
-		cl.w = w
-		return cl
-	})
-	return cl
 }
 
 type colorLogger struct {
-	logger log.Logger
-	color  func(keyvals ...interface{}) FgBgColor
-	w      io.Writer
-	mu     sync.RWMutex // protects w
-
-	actColor   FgBgColor
-	actColorMu sync.RWMutex // protects actColor
+	w             io.Writer
+	newLogger     func(io.Writer) log.Logger
+	color         func(keyvals ...interface{}) FgBgColor
+	bufPool       sync.Pool
+	noColorLogger log.Logger
 }
 
 func (l *colorLogger) Log(keyvals ...interface{}) error {
-	l.actColorMu.Lock() // Unlock is in Write!
-	if l.color != nil {
-		l.actColor = l.color(keyvals...)
-	}
-	l.actColorMu.Unlock()
-	return l.logger.Log(keyvals...)
-}
-
-func (l *colorLogger) Write(p []byte) (int, error) {
-	l.actColorMu.RLock()
-	color := l.actColor
-	l.actColorMu.RUnlock()
+	color := l.color(keyvals...)
 	if color.IsZero() {
-		return l.w.Write(p)
+		return l.noColorLogger.Log(keyvals...)
 	}
-	var n int
+
+	lb := l.getLoggerBuf()
+	defer l.putLoggerBuf(lb)
 	if color.Fg != NoColor {
-		m, err := fmt.Fprintf(l.w, "\x1b[%dm", 30+color.Fg)
-		if err != nil {
-			return m, err
-		}
-		n += m
+		lb.buf.Write(fgColorBytes[color.Fg])
 	}
 	if color.Bg != NoColor {
-		m, err := fmt.Fprintf(l.w, "\x1b[%dm", 40+color.Bg)
-		if err != nil {
-			return n + m, err
-		}
-		n += m
+		lb.buf.Write(bgColorBytes[color.Bg])
 	}
-	m, err := l.w.Write(p)
+	err := lb.logger.Log(keyvals...)
 	if err != nil {
-		return n + m, err
+		return err
 	}
-	n += m
-	l.mu.RLock()
-	w := l.w
-	l.mu.RUnlock()
-	m, err = w.Write([]byte("\x1b[0m"))
-	return n + m, err
+	if color.Fg != NoColor || color.Bg != NoColor {
+		lb.buf.Write(resetColorBytes)
+	}
+	_, err = io.Copy(l.w, lb.buf)
+	return err
 }
 
-func (l *colorLogger) Hijack(f func(io.Writer) io.Writer) {
-	l.mu.Lock()
-	l.w = f(l.w)
-	l.mu.Unlock()
+type loggerBuf struct {
+	buf    *bytes.Buffer
+	logger log.Logger
+}
+
+func (l *colorLogger) getLoggerBuf() *loggerBuf {
+	lb := l.bufPool.Get().(*loggerBuf)
+	if lb.buf == nil {
+		lb.buf = &bytes.Buffer{}
+		lb.logger = l.newLogger(lb.buf)
+	} else {
+		lb.buf.Reset()
+	}
+	return lb
+}
+
+func (l *colorLogger) putLoggerBuf(cb *loggerBuf) {
+	l.bufPool.Put(cb)
 }
 
 func asString(v interface{}) string {
--- a/log/term/colorlogger.go
+++ b/log/term/colorlogger.go
@@ -9,11 +9,22 @@ import (
 	"github.com/go-kit/kit/log"
 )
 
-// Color is the abstract color, the zero value is the Default.
+// Color represents an ANSI color. The zero value is Default.
 type Color uint8
 
+// ANSI colors.
 const (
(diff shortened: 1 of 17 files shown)


tgulacsi:
Nice solution.

peterbourgon:
Commenting before I read the PR thoroughly, but this caught my eye...

> The concurrency question is handled by a sync.Pool that manages a free list of buffer-Logger pairs.

Could this technique be applied to the other loggers, too?

ChrisHines:
>>The concurrency question is handled by a sync.Pool that manages a free list of buffer-Logger pairs.

>Could this technique be applied to the other loggers, too?

In some cases, yes, but I don't see any opportunities in the other Go kit loggers at the moment.

log/term/colorlogger.go
```diff
@@ -0,0 +1,149 @@
...
+var resetColorBytes = []byte("\x1b[0m")
+var fgColorBytes [][]byte
+var bgColorBytes [][]byte
+
+func init() {
+	for color := NoColor; color < maxColor; color++ {
+		fgColorBytes = append(fgColorBytes, []byte(fmt.Sprintf("\x1b[%dm", 30+color)))
+		bgColorBytes = append(bgColorBytes, []byte(fmt.Sprintf("\x1b[%dm", 40+color)))
+	}
+}
```
peterbourgon: I strongly want Go kit to be totally side-effect free. Is there any way this initialization can be expanded to const literals?

log/term/colorlogger.go
```diff
@@ -0,0 +1,149 @@
...
+	return c.Fg == NoColor && c.Bg == NoColor
+}
+
+// NewColorLogger returns a log.Logger which writes colored logs to w. It
+// colors whole records based on the FgBgColor returned by the color function.
+// Log events are formatted by the Logger returned by newLogger.
+func NewColorLogger(w io.Writer, newLogger func(io.Writer) log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger {
+	if color == nil {
+		panic("color func nil")
+	}
```
peterbourgon: I guess we also want an IsTty check here somewhere?

peterbourgon:
> In some cases, yes, but I don't see any opportunities in the other Go kit loggers at the moment.

Ah, you're right, as the JSON logger writes directly to the writer, and the Logfmt logger defers to the logfmt package. But perhaps it would be a nice optimization [for package logfmt](https://github.com/go-logfmt/logfmt/blob/master/logfmt.go#L27)?

peterbourgon:
I think this package would benefit from a usage example, in e.g. example_test.go. But apart from that, :+1:! Looks good.

log/term/colorlogger.go
tgulacsi: //go:generate sh -c sh -c 'echo "var fgColorBytes=[...]byte{"$(seq -f "\"\\x1b[%gm\"" 30 39 | xargs echo | sed -e "s/ /,/g")"}"; echo "var bgColorBytes=[...]byte{"$(seq -f "\"\\x1b[%gm\"" 40 49 | xargs echo | sed -e "s/ /,/g")"}"'

?

ChrisHines: These optimizations are inspired by similar code in `fmt`. See for example [fmt/print.go](http://golang.org/src/fmt/print.go) and [fmt/format.go](http://golang.org/src/fmt/format.go). I am not aware of a way to do this with constants.

ChrisHines: @tgulacsi That's not going to work for me as a maintainer of this package because my development environment doesn't have `sh`, `echo`, `sed`, or `xargs`.

log/term/colorlogger.go
ChrisHines: In my review of https://github.com/go-kit/kit/pull/104 I suggested that it's up to the application to call `IsTty` and do the right thing.

log/term/colorlogger.go
tgulacsi: Then just

```
var
fgColorBytes=[...]byte{"\x1b[30m","\x1b[31m","\x1b[32m","\x1b[33m","\x1b[34m","\x1b[35m","\x1b[36m","\x1b[37m","\x1b[38m","\x1b[39m"}
var
bgColorBytes=[...]byte{"\x1b[40m","\x1b[41m","\x1b[42m","\x1b[43m","\x1b[44m","\x1b[45m","\x1b[46m","\x1b[47m","\x1b[48m","\x1b[49m"}
```

Only needed once.

---
https://onename.com/tgulacsi

2015-08-31 19:41 GMT+02:00 Chris Hines <notifications@github.com>:

ChrisHines:
>But perhaps it would be a nice optimization for package logfmt?

It doesn't help in that context because the `logfmt.MarshalKeyvals` returns the underlying `[]byte` to the caller and therefore cannot reuse the buffer. But `LogfmtLogger` could potentially use this strategy with a `logfmt.Encoder`. I'll benchmark that sometime to see if there is any benefit.

log/term/colorlogger.go
ChrisHines: Sure, that works, but I don't really so how manually listing them out is an improvement over an init function. This is an implementation detail that doesn't leak outside the package, I don't consider it a side-effect. No one is going to be adding
```go
import (
    _ "github.com/go-kit/kit/log/term"
)
```
to their code because of this init function.

tsenart: `var foo = genFoo()`?

peterbourgon: That's true. Besides a futile quest for purity, the concrete scenario I'm concerned with is someone who (accidentally) imports two versions of this package to their dependency graph, and the two versions declare the arrays slightly differently. But on reflection, I guess that wouldn't be any different than a package-global var, flatly declared. If you agree & I'm not missing anything, consider my request withdrawn.

ChrisHines: Two copies of the package causes problems when trying to interchange values between the public API of the two instances or when interacting with external resources. These arrays have neither problem. Another copy of go-kit (under a different import path) can get at them than no more than any other package.

log/term/colorlogger.go
peterbourgon: I'm not sure about that. By constructing a ColorLogger, the user is implicitly declaring that they intend to read its output on a TTY. I don't think a further if-guard on that intent is appropriate; it makes sense to me to bake that assumption into the constructor. And, taking it a step further, if that assumption is invalidated at runtime, it would be incidental, and as a user I'd expect the logger to do the right thing.

What do you think?

ChrisHines: I think explicit is better than implicit and that baking this decision into `ColorLogger` control away from the application developer.

For example, I plan to use `ColorLogger` with github.com/mattn/go-colorable, which provides the `NewColorableStdout` and `NewColorableStderr` functions. On Linux these function just return `os.Stdout` and `os.Stderr` respectively, but on Windows they return an `io.Writer` that wraps the appropriate stream and converts the color codes into function calls to change the text color in the Windows console. The Windows `io.Writer`s would not satisfy `IsTty`.

There may be other situations where `ColorLogger` is only *indirectly* writing to a TTY. So, I believe `ColorLogger` should do what it says, and let the application decide if it is needed.

Also, without the check for a TTY, `ColorLogger` is easier to test.

peterbourgon: Hmm... I'm afraid I'm not convinced. I find an explicit test for IsTty at the callsite to be too user-hostile. The semantics of ColorLogger are: wrap another logger, and colorize the output according to some declared rules, if possible. If the final Writer can't handle color codes, there's no reason to punish me for the mistake of not properly sanitizing my operating context, as that detection •can• happen automatically. Put another way: colorization is a pure value-add, and should fail gracefully; a broken ColorLogger should no-op, rather than produce mangled output.

If colorizing output on Windows isn't possible without wiring in mattn's lib, we should ship with that package, too, same as Simon's.

ChrisHines: I respect your opinion on this, but I also feel that the emergent philosophy of Go kit aligns with the Unix philosophy which pushes the other direction. I propose we keep `ColorLogger` as it is with the semantics of *write colored log events to an `io.Writer`*, and provide a higher level convenience that combines the `IsTty` check, `ColorLogger`, and a vendored go-colorable to provide the convenience you want. I think having all of these in the log/term package makes sense. The main package documentation can focus on the most convenient approach but the building blocks are still accessible to anyone with more specialized needs.

I imagine the package would export:
- `TtyLogger`  (the all-in-one convenience)
- `ColorLogger`
- `TtyColorWriter` (go-colorable logic as an `io.Writer` decorator)
- `IsTty`

peterbourgon: > `TtyLogger` ... `ColorLogger`

Okay, this is exactly what I'm looking for. :+1:!

ChrisHines:
@peterbourgon, @tgulacsi Please take another look at this. I have polished it up quite a bit. It could still use some more package docs, but I believe this should be feature complete now.

log/term/colorlogger.go
```diff
@@ -0,0 +1,171 @@
...
+	Yellow
+	Blue
+	Magenta
+	Cyan
+	White
+
+	numColors
+)
+
+var resetColorBytes = []byte("\x1b[39;49m")
```
peterbourgon: Is there some reference for these escape codes?

log/term/colorlogger.go
```diff
@@ -0,0 +1,171 @@
...
+var fgColorBytes [][]byte
+var bgColorBytes [][]byte
```
peterbourgon: How about one `var` block?

log/term/colorlogger.go
```diff
@@ -0,0 +1,171 @@
...
+	switch x := v.(type) {
+	case string:
+		return x
+	default:
+		return fmt.Sprint(x)
+	}
+}
+
+// LevelColor returns the color for the record based on the value of the
+// "level" key, if it exists.
```
peterbourgon: This doesn't compose with user-defined level keys or values. At a minimum, we should document that, by adding e.g. "...based on the _default_ value of the "level" key, if it exists." Maximally, we should allow these things to be modified—but I'm not sure if that would make the API terrible.

peterbourgon:
LGTM, those points aside.

log/term/colorlogger.go
tgulacsi: http://invisible-island.net/xterm/ctlseqs/ctlseqs.html

https://en.m.wikipedia.org/wiki/ANSI_escape_code
2015. szept. 6. de. 10:04 ezt írta ("Peter Bourgon" <
notifications@github.com>):

log/term/colorlogger.go
ChrisHines: I agree. I would like to punt on that for now. Since the API allows application provided color functions we don't really need to provide any right away. I will convert this code into another example and we can defer providing concrete color functions until we get more experience with these packages.

ChrisHines:
Personally I'm still a little unhappy with the parameters types for
```go
IsTerminal(fd uintptr) bool
NewColorWriter(w FdWriter) io.Writer
```
I am going to change both of these to accept an `io.Writer` instead? It makes `IsTerminal` less general purpose—it cannot be used to check if `os.Stdin` is a terminal—but I think it is more convenient. I don't think the loss of generality is a concern since log/term is only concerned with producing output.

This change means we do not need to export the FdWriter interface.

PR #118, opened 2015-09-04
Fixed README.md

nii236:
- `ErrEmpty` was not declared

- For some reason io.writer and io.reader was used instead of http request and response bodies. I had to check against the example source code to find out why the README code wasn't working.


peterbourgon:
Sorry for that, and thank you!

PR #119, opened 2015-09-04
Adding Prometheus handler

domudall:
Adding Prometheus handler to allow metrics to be visible on `/metric` in examples stringsvc2 and stringsvc3.

Just provides a bit more of a full view of an example service with the metrics available via an endpoint.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -422,6 +422,8 @@ func main() {
 		Endpoint: makeCountEndpoint(svc),
 		// ...
 	}
+
+	http.Handle("/metrics", stdprometheus.Handler())
 }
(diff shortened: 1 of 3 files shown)


peterbourgon:
Yes! Thanks. Only on a phone right now — do the code snippets in the README also need to be updated for consistency?

domudall:
Not sure, also on phone so will check when I get in

domudall:
@peterbourgon I've updated the README where it seemed relevant; it's only adding a single handler so only one line added =)

peterbourgon:
Yep, that'll do. Thanks!

PR #120, opened 2015-09-05
Add StringifyLogger to eliminate "unsupported value type"

tgulacsi:
Just a simple wrapper to tame LogfmtLogger.

### DIFF
--- a/log/stringify_logger.go
+++ b/log/stringify_logger.go
@@ -1,0 +1,36 @@
+package log
+
+import "fmt"
+
+// StringifyLogger stringifies every value to make it printable by logfmt.
+//
+// Example:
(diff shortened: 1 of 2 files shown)


ChrisHines:
This seems like a cop out. I'd like to either come up with a good answer to the questions in https://github.com/go-logfmt/logfmt/issues/1 or incorporate some opinion about them into Go kit's LogfmtLogger directly.

peterbourgon:
Yes, let's solve this in the affected logger(s).

ChrisHines:
@tgulacsi Please create an issue with a statement of the problem from a Go kit perspective where we can discuss potential solutions.

PR #126, opened 2015-09-14
Testing if LoadBalancer is nil before Retry

jesselucas:
In cases that a LoadBalancer isn't created correctly a nil value can be passed in to the Retry function. `panic: runtime error: invalid memory address or nil pointer dereference` This checks if the LoadBalancer value is nil and returns with an error passed to the errs channel.

### DIFF
--- a/loadbalancer/retry.go
+++ b/loadbalancer/retry.go
@@ -15,6 +15,10 @@ import (
 // balancer. Requests that return errors will be retried until they succeed,
 // up to max times, or until the timeout is elapsed, whichever comes first.
 func Retry(max int, timeout time.Duration, lb LoadBalancer) endpoint.Endpoint {
+	if lb == nil {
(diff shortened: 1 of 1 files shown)


peterbourgon:
A nil LoadBalancer is a programmer error. I'd prefer a panic, e.g.

```go
func Retry(max int, timeout time.Duration, lb LoadBalancer) endpoint.Endpoint {
    if lb == nil {
        panic("nil LoadBalancer")
    }
    // ...
}
```

jesselucas:
As you described above, I amended the previous commit to panic if the LoadBalancer is nil.

peterbourgon:
```
loadbalancer/retry.go:4: imported and not used: "errors"
```

jesselucas:
Fixed. Go imports really wanted it in there :)

PR #127, opened 2015-09-14
Updating loadbalancer package README.md

jesselucas:
Updating to show dnssrv.NewPublisher returns a *Publisher and error value to fix `multiple-value dnssrv.NewPublisher() in single-value context`


peterbourgon:
Thank you!

PR #128, opened 2015-09-18
loadbalancer: closable endpoints

### LINKED ISSUE #122: publisher/factory workflow should provide mechanism for cleanup of persistent resources
When using the publisher/factory/load balancer workflow, there is no mechanism currently provided for a publisher to trigger the cleanup of an endpoint which is no longer provided.

peterbourgon:
This PR introduces a change that allows endpoints to be closed. This is useful for factories which create endpoints that need some kind of lifecycle management. Closes #122.

### DIFF
--- a/examples/stringsvc3/proxying.go
+++ b/examples/stringsvc3/proxying.go
@@ -64,12 +64,12 @@ func (mw proxymw) Uppercase(s string) (string, error) {
 }
 
 func factory(ctx context.Context, qps int) loadbalancer.Factory {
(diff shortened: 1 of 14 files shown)


peterbourgon:
@EcofitBrendan would you mind taking a look if you have a free moment or two?

EcofitBrendan:
This looks good to me, the change to the factory pattern is relatively low impact and the implementation of endpointCache makes a lot of sense.

peterbourgon:
Thanks!

PR #135, opened 2015-10-01
Websockets support proposal [rfc]

gngeorgiev:


peterbourgon:
I still don't understand what you mean by "support". If we're just talking about treating websockets as a transport for RPC semantics, then I'm :+1: on this idea, if you can provide a prototype implementation. If we're talking about websockets as persistent bidirectional streams, then this is not (yet) in the scope of Go kit.

peterbourgon:
Closed per above comments. To continue with this proposal, please provide a prototype websocket transport implementation as a PR.

PR #136, opened 2015-10-04
Make more sence to go.

wangtuanjie:


peterbourgon:
Failures are related to a gRPC update, I think.

wangtuanjie:
I think we can use `vendor` to manage dependency.

peterbourgon:
> [Go kit] recommend our users use vendoring for all of their dependencies, including Go kit. To avoid compatibility and availability issues, Go kit doesn't vendor its own dependencies, and doesn't recommend use of third-party import proxies.

—https://github.com/go-kit/kit#dependency-management

wangtuanjie:
If Go kit doesn't vendor its own dependencies, it  must keep up upstream changes which is a `big work`.

peterbourgon:
That's correct.

peterbourgon:
Thank you!

PR #138, opened 2015-10-09
log: Fix StdlibAdapter parsing of Windows file paths when stdlib log has the Llongfile flag set.

juRiii:
the following code should generate the specified output 

# code
```
	stdlog.SetFlags(stdlog.Ldate | stdlog.Ltime | stdlog.Llongfile)
	logger := kitlog.NewJSONLogger(os.Stdout)
	stdlog.SetOutput(kitlog.NewStdlibAdapter(logger))
	stdlog.Println("log...")
```
# output
**expected:**
`{"file":"C:/Dev/go/src/github.com/juRiii/gokittest/gokit_test.go:14","msg":"log...","ts":"2015/10/09 18:29:50"}`
**received:**
`{"msg":"C:/Dev/go/src/github.com/juRiii/test/gokit_test.go:13: log...","ts":"2015/10/09 18:47:07"}`

looks like, this occurs only on windows with the flag `stdlog.Llongfile`. is there a reason, why `(?P<file>[^:]+:[0-9]+)?` was used instead of `(?P<file>.+:[0-9]+)?`?

### DIFF
--- a/log/stdlib.go
+++ b/log/stdlib.go
@@ -95,7 +95,7 @@ func (a StdlibAdapter) Write(p []byte) (int, error) {
 const (
 	logRegexpDate = `(?P<date>[0-9]{4}/[0-9]{2}/[0-9]{2})?[ ]?`
 	logRegexpTime = `(?P<time>[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?)?[ ]?`
-	logRegexpFile = `(?P<file>[^:]+:[0-9]+)?`
+	logRegexpFile = `(?P<file>.+:[0-9]+)?`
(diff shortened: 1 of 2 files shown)


ChrisHines:
Please add a test for this issue.

juRiii:
done, is it what you have expected?

ChrisHines:
Thanks for adding the tests. They look good.

It would be nice if you would rebase these changes onto the current go-kit/kit:master and update this PR before we merge? Normally it's not a big deal, but your first commit started from a rather old commit and then go-kit/kit:master got merged into that. Rebasing would clean that up nicely. Thanks.

juRiii:
done

ChrisHines:
Thanks.

PR #141, opened 2015-10-11
log: Fix parsing with special characters

juRiii:
- Related to #138 
- Was not working with some special characters, added some test cases


peterbourgon:
Ah, nice, thank you.

PR #143, opened 2015-10-25
Loadbalancer: satisfy Publisher interface with EndpointCache

cihangir:
Allow usages of EndpointCache as Publisher

### DIFF
--- a/loadbalancer/dnssrv/publisher.go
+++ b/loadbalancer/dnssrv/publisher.go
@@ -72,7 +72,7 @@ func (p *Publisher) loop() {
 
 // Endpoints implements the Publisher interface.
 func (p *Publisher) Endpoints() ([]endpoint.Endpoint, error) {
-	return p.cache.Endpoints(), nil
+	return p.cache.Endpoints()
 }
 
 var (
--- a/loadbalancer/endpoint_cache.go
+++ b/loadbalancer/endpoint_cache.go
@@ -78,13 +78,14 @@ func (t *EndpointCache) Replace(instances []string) {
 	t.m = m
 }
 
-// Endpoints returns the current set of endpoints in undefined order.
-func (t *EndpointCache) Endpoints() []endpoint.Endpoint {
+// Endpoints returns the current set of endpoints in undefined order. Satisfies
+// Publisher interface.
+func (t *EndpointCache) Endpoints() ([]endpoint.Endpoint, error) {
 	t.mtx.RLock()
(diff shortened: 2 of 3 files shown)


cihangir:
It seems there are some problems with etcd client package. Is there any action plan about this kind of cases?

peterbourgon:
What motivates this PR?

peterbourgon:
re: the test failures I've reproduced and filed [coreos/go-etcd #246](https://github.com/coreos/go-etcd/issues/246).

cihangir:
https://github.com/go-kit/kit/pull/143#issuecomment-150924101

When i check static, fixed and EndpointCache they have very little different code and can be unified. After this PR, i can refactor the code to utilize `fixed` to be used in each of them and/or can remove EndpointCache completely. 

What motivates me, apart from code cleanup, in case you are not OK with removing some of the LBs, i want to have the `Closer` utility of the EndpointCache with the flexibility of `Replace` as my Publisher.

What do you think?

peterbourgon:
I see. Yes, I think this is a fine idea. I need to take a little while and read through the proposal; perhaps there is more code that can be removed?

cihangir:
Refactoring looks more complex than i imagined :) i _hacked_ it a bit here https://github.com/cihangir/kit/commit/9a1b71fabfb283f88d94f47f77409f202a971f65
when you check the code you will see: 
1)  fixed.NewPublisher constructor func signature is completely changed
2) endpoint_cache is removed 
3) static.Publisher is kinda useless now 
4) all Publishers are using fixed internally

so, 
1) are you OK with backward incompatible constructor changes? ( or in general )
2) are you OK with removing static.Publisher?

peterbourgon:
@cihangir Thank you for the contribution. I haven't had time to give it the attention it deserves; I hope I'll find the time shortly. 

As preliminary feedback: I am totally fine with backwards-incompatible changes of any form, and I'm probably OK with removing something like static.Publisher. But I'm not so sure about all publishers using (being implemented in terms of) fixed.Publisher. That makes a strange import graph; I would rather see that functionality in package loadbalancer directly, so I feel like I'd prefer to see fixed.Publisher removed, and endpoint cache kept.

I'm also raising my eyebrow a bit at fixed.Publisher.Replace, it seems like more code/logic than it needs to be, though I have to give it a proper read to be sure.

peterbourgon:
For the record I haven't forgotten about this, still trying to find the time :)

PR #145, opened 2015-10-26
transport/http: add BadRequestError.Raw()

### LINKED ISSUE #144: BadRequestError does not reveal the embedded error to ErrorEncoder
`BadRequestError` has `error` as embedded type. Since `error` starts with small cap, it is an implicit unexported field of `BadRequestError`. That means only code within `transport/http` package can access the raw error value. This leave no room for `ErrorEncoder` implementation to unpack error messages of custom error types.

Suggest either to:
1. To add a `BadRequestError.Raw{}` method that returns the raw error inside.
2. Change `BadRequestError` from `type struct{error}` to `type struct{ Raw: error }`.

yookoala:
Alternative solution 1.
Fixes #144.

### DIFF
--- a/transport/http/error.go
+++ b/transport/http/error.go
@@ -1,0 +1,9 @@
+package http
+
+// BadRequestError is an error in decoding the request.
+type BadRequestError struct{ error }
+
+// Raw returns the raw embedded error
+func (err BadRequestError) Raw() error {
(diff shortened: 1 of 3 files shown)


peterbourgon:
I'll take #146 over this one.

PR #146, opened 2015-10-26
transport/http: expose BadRequestError raw error

### LINKED ISSUE #144: BadRequestError does not reveal the embedded error to ErrorEncoder

yookoala:
Alternative solution 2.

Expose BadRequestError embedded error value as BadRequestError.Err.

Fixes #144.

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -118,4 +118,11 @@ func defaultErrorEncoder(w http.ResponseWriter, err error) {
 }
 
 // BadRequestError is an error in decoding the request.
-type BadRequestError struct{ error }
+type BadRequestError struct {
+	Err error
+}
+
+// Error implements the error interface.
+func (err BadRequestError) Error() string {
+	return err.Err.Error()
+}
--- a/transport/http/server_test.go
+++ b/transport/http/server_test.go
@@ -118,3 +118,20 @@ func testServer(t *testing.T) (cancel, step func(), resp <-chan *http.Response) 
 	}()
 	return cancelfn, func() { stepch <- true }, response
 }
+
+type testBadRequestError struct {
+	code int
+}
+
+func (err testBadRequestError) Error() string {
+	return "Bad Request"
+}
+
+func TestBadRequestError(t *testing.T) {
+	inner := testBadRequestError{1234}
+	var outer error = httptransport.BadRequestError{inner}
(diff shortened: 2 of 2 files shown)


transport/http/server.go
```diff
@@ -118,4 +118,11 @@ func defaultErrorEncoder(w http.ResponseWriter, err error) {
 }
 
 // BadRequestError is an error in decoding the request.
-type BadRequestError struct{ error }
+type BadRequestError struct {
+	Raw error
```
peterbourgon: I think it's more idiomatic to call this Err.

transport/http/server_test.go
```diff
@@ -118,3 +118,20 @@ func testServer(t *testing.T) (cancel, step func(), resp <-chan *http.Response)
...
+
+func (err testBadRequestError) Error() string {
+	return "Bad Request"
+}
+
+func TestBadRequestError(t *testing.T) {
+	inner := testBadRequestError{1234}
+	var outer error = httptransport.BadRequestError{inner}
+	err := outer.(httptransport.BadRequestError)
+	if err.Raw != inner {
```
peterbourgon:
```go
if want, have := inner, err.Raw; want != have {
    // ...
}
```

transport/http/server.go
```diff
@@ -118,4 +118,11 @@ func defaultErrorEncoder(w http.ResponseWriter, err error) {
...
+}
+
+// Error returns string error message
```
peterbourgon: Please use proper grammar and punctuation.

```
Error implements the Error interface.
```

yookoala:
Amended the previous commit and adopt changes suggested by @peterbourgon.

peterbourgon:
:+1:! Thank you. I will wait until [etcd #246](https://github.com/coreos/go-etcd/issues/246) is fixed, and CI can be successfully run, before merging this.

PR #111, opened 2015-08-26
Add NewTerminalLogger for colored leveled logging on terminals.

tgulacsi:
Uses ColorLogger for coloring the lines based on the "level"'s value.

PR #150, opened 2015-11-18
Add log/term/NewTerminalLogger as log15.TerminalFormat

tgulacsi:
Configurable thorugh LogOpts, bat that's maybe overkill, I don't know.

Try to add the remaining from #111

### DIFF
--- a/log/term/termlogger.go
+++ b/log/term/termlogger.go
@@ -1,0 +1,216 @@
+package term
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"gopkg.in/logfmt.v0"
+)
+
+var _ = log.Logger((*terminalLogger)(nil))
+
+// NewTerminalLogger returns a log.Logger which formats its output as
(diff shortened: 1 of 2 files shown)


ChrisHines:
@tgulacsi Is there any way the level functionality here could build on or work with the log/levels package?

tgulacsi:
One option is to let levels/Levels.TermLogOpts to provide a populated term.LogOpts (ofc that has to move to some common pkg), but as TermLogOpts is almost the same as Levels (with the exception of ctx), maybe factor out the levelKey, ...Value part?
Or create a ColoredLevels copy of Levels?

peterbourgon:
Is this closeable?

peterbourgon:
Whoops, meant that as a question.

ChrisHines:
We can't merge it until it is made more orthogonal with the log/levels package.

@tgulacsi Do you plan to put more work into this?

tgulacsi:
As levels' internals are sealed (Options sets unexported params, no way to retrieve those values), and I don't want to import FgBgColor into levels (to provide a Color function in levels), I don't see a solution to integrate this into log/levels. 
But this is awkward now, to have to duplicate the level key and value info.

PR #155, opened 2015-11-24
httptransport: use ctxhttp to propagate context

### LINKED ISSUE #154: HTTP client transport should use ctxhttp?
Since https://github.com/golang/go/issues/11904, the context package has a nice wrapper around http.Client that is context aware. Would there be any interest in a PR that uses that package internally to wrap the http.Client so `cancel()`s will propagate to the request?

simonjefford:
Fixes #154

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 
 	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
 
 	"github.com/go-kit/kit/endpoint"
 )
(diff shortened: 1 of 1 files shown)


peterbourgon:
That easy, huh? :)

transport/http/client.go
```diff
@@ -5,9 +5,9 @@ import (
 	"net/http"
 	"net/url"
 
-	"golang.org/x/net/context"
-
 	"github.com/go-kit/kit/endpoint"
+	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
```
peterbourgon: Please keep these imports where they were before, above the go-kit block.

peterbourgon:
That minor nit aside, :+1:

simonjefford:
Apologies - force of habit - I tend to delete blank lines in the "3rd party" import block then let `goimports` sort them.

PR #156, opened 2015-11-25
Json logger now calls fmt.Stringer if the type does not implement json.Marshaler or encoding.TextMarshaler

hydroflame:


ChrisHines:
Thanks!

PR #157, opened 2015-11-28
Make levels API more composable.

### LINKED ISSUE #131: Level logging difficult to work with
logger := log.NewLogfmtLogger(os.Stderr)
logger = log.NewContext(logger).With("caller", log.DefaultCaller)
lvlLogger := levels.New(logger)
lvlLogger.Info("hi", "there")

In the output you will get: **caller=levels.go:72**

This is fixed by using **log.Caller(4)** instead of **log.DefaultCaller** but makes passing and extending the logger harder than it should be.

ChrisHines:
Fixes #131.

### DIFF
--- a/log/levels/levels.go
+++ b/log/levels/levels.go
@@ -62,29 +62,29 @@ func (l Levels) With(keyvals ...interface{}) Levels {
 	}
 }
 
-// Debug logs a debug event along with keyvals.
-func (l Levels) Debug(keyvals ...interface{}) error {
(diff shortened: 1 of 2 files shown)


ChrisHines:
This is a breaking change. I will post it to the mailing list and Slack channel and see if there are any objections before I merge.

tgulacsi:
I like this change, takes the leveled logging back to the group of the other loggers.

peterbourgon:
May I delete the branch?

ChrisHines:
Taken care of.

PR #160, opened 2015-12-05
initial take on zookeeper publisher

basvanbeek:
Here's my initial take on a zookeeper publisher (closely resembling the etcd implementation). I didn't have time to build a standard publisher test but didn't want to wait for you guys to be able to review and discuss the code

### DIFF
--- a/loadbalancer/zk/client.go
+++ b/loadbalancer/zk/client.go
@@ -1,0 +1,157 @@
+package zk
+
+import (
+	"errors"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/eapache/channels"
+	"github.com/samuel/go-zookeeper/zk"
+
+	"github.com/go-kit/kit/log"
+)
+
+// DefaultACL is the default ACL to use for creating znodes
+var DefaultACL = zk.WorldACL(zk.PermAll)
+
+const (
+	// ConnectTimeout is the default timeout to establish a connection to a
+	// ZooKeeper node.
+	ConnectTimeout = 2 * time.Second
+	// SessionTimeout is the default timeout to keep the current ZooKeeper
+	// session alive during a temporary disconnect.
+	SessionTimeout = 5 * time.Second
+)
+
+// Client is a wrapper around a lower level ZooKeeper client implementation.
+type Client interface {
+	// GetEntries should query the provided path in ZooKeeper, place a watch on
+	// it and retrieve data from its current child nodes.
+	GetEntries(path string) ([]string, channels.SimpleOutChannel, error)
+	// CreateParentNodes should try to create the path in case it does not exist
+	// yet on ZooKeeper.
+	CreateParentNodes(path string) error
+}
+
+type clientConfig struct {
+	acl             []zk.ACL
+	connectTimeout  time.Duration
+	sessionTimeout  time.Duration
+	rootNodePayload [][]byte
+}
+
+type configOption func(c *clientConfig) error
+
+type client struct {
+	*zk.Conn
+	clientConfig
+	Eventc <-chan zk.Event
+	quit   chan struct{}
+}
+
+// WithACL returns a configOption specifying a non-default ACL.
+func WithACL(acl []zk.ACL) configOption {
+	return func(c *clientConfig) error {
+		c.acl = acl
+		return nil
+	}
+}
+
+// WithConnectTimeout returns a configOption specifying a non-default connection
+// timeout when we try to establish a connection to a ZooKeeper server.
+func WithConnectTimeout(t time.Duration) configOption {
+	return func(c *clientConfig) error {
+		if t.Seconds() < 1 {
+			return errors.New("Invalid Connect Timeout. Minimum value is 1 second")
+		}
+		c.connectTimeout = t
+		return nil
+	}
+}
+
+// WithSessionTimeout returns a configOption specifying a non-default session
+// timeout.
+func WithSessionTimeout(t time.Duration) configOption {
+	return func(c *clientConfig) error {
+		if t.Seconds() < 1 {
+			return errors.New("Invalid Session Timeout. Minimum value is 1 second")
+		}
+		c.sessionTimeout = t
+		return nil
+	}
+}
+
+// WithPayload returns a configOption specifying non-default data values for
+// each znode created by CreateParentNodes.
+func WithPayload(payload [][]byte) configOption {
+	return func(c *clientConfig) error {
+		c.rootNodePayload = payload
+		return nil
+	}
+}
+
+// NewClient returns a ZooKeeper client with a connection to the server cluster.
+// It will return an error if the server cluster cannot be resolved.
+func NewClient(servers []string, logger log.Logger, options ...configOption) (Client, error) {
+	config := clientConfig{
+		acl:            DefaultACL,
+		connectTimeout: ConnectTimeout,
+		sessionTimeout: SessionTimeout,
+	}
+	for _, option := range options {
+		if err := option(&config); err != nil {
+			return nil, err
+		}
+	}
+	// dialer overrides the default ZooKeeper library Dialer so we can configure
+	// the connectTimeout. The current library has a hardcoded value of 1 second
+	// and there are reports of race conditions, due to slow DNS resolvers
+	// and other network latency issues.
+	dialer := func(network, address string, _ time.Duration) (net.Conn, error) {
+		return net.DialTimeout(network, address, config.connectTimeout)
+	}
+	conn, eventc, err := zk.Connect(servers, config.sessionTimeout, withLogger(logger), zk.WithDialer(dialer))
+	if err != nil {
+		return nil, err
+	}
+	return &client{conn, config, eventc, make(chan struct{})}, nil
+}
+
+// CreateParentNodes implements the ZooKeeper Client interface.
+func (c *client) CreateParentNodes(path string) error {
+	payload := []byte("")
+	pathString := ""
+	pathNodes := strings.Split(path, "/")
+	for i := 1; i < len(pathNodes); i++ {
+		if i <= len(c.rootNodePayload) {
+			payload = c.rootNodePayload[i-1]
+		} else {
+			payload = []byte("")
+		}
+		pathString += "/" + pathNodes[i]
+		_, err := c.Create(pathString, payload, 0, c.acl)
+		if err != nil && err != zk.ErrNodeExists && err != zk.ErrNoAuth {
+			return err
+		}
+	}
+	return nil
+}
+
+// GetEntries implements the ZooKeeper Client interface.
+func (c *client) GetEntries(path string) ([]string, channels.SimpleOutChannel, error) {
+	// retrieve list of child nodes for given path and add watch to path
+	znodes, _, updateReceived, err := c.ChildrenW(path)
+	if err != nil {
+		return nil, channels.Wrap(updateReceived), err
+	}
+
+	var resp []string
+	for _, znode := range znodes {
+		// retrieve payload for child znode and add to response array
+		if data, _, err := c.Get(path + "/" + znode); err == nil {
+			resp = append(resp, string(data))
+		}
+	}
+	return resp, channels.Wrap(updateReceived), nil
+}
--- a/loadbalancer/zk/client.go
+++ b/loadbalancer/zk/client.go
@@ -6,98 +6,141 @@ import (
 	"strings"
 	"time"
 
-	"github.com/eapache/channels"
 	"github.com/samuel/go-zookeeper/zk"
 
 	"github.com/go-kit/kit/log"
 )
 
-// DefaultACL is the default ACL to use for creating znodes
-var DefaultACL = zk.WorldACL(zk.PermAll)
+// DefaultACL is the default ACL to use for creating znodes.
+var (
+	DefaultACL            = zk.WorldACL(zk.PermAll)
+	ErrInvalidPath        = errors.New("path must start with / character")
+	ErrInvalidCredentials = errors.New("invalid credentials provided")
+	ErrClientClosed       = errors.New("client service closed")
+)
 
 const (
-	// ConnectTimeout is the default timeout to establish a connection to a
-	// ZooKeeper node.
-	ConnectTimeout = 2 * time.Second
-	// SessionTimeout is the default timeout to keep the current ZooKeeper
-	// session alive during a temporary disconnect.
-	SessionTimeout = 5 * time.Second
+	// DefaultConnectTimeout is the default timeout to establish a connection to
+	// a ZooKeeper node.
+	DefaultConnectTimeout = 2 * time.Second
+	// DefaultSessionTimeout is the default timeout to keep the current
+	// ZooKeeper session alive during a temporary disconnect.
+	DefaultSessionTimeout = 5 * time.Second
 )
+
+// Event is a data container for znode watch events returned by GetEntries. It
+// provides a consistent trigger for watch events to Publishers, while still
+// allowing the underlying ZooKeeper client library watch payload to be
+// inspected.
+type Event struct {
+	payload interface{}
+}
 
 // Client is a wrapper around a lower level ZooKeeper client implementation.
 type Client interface {
 	// GetEntries should query the provided path in ZooKeeper, place a watch on
 	// it and retrieve data from its current child nodes.
-	GetEntries(path string) ([]string, channels.SimpleOutChannel, error)
+	GetEntries(path string) ([]string, <-chan Event, error)
 	// CreateParentNodes should try to create the path in case it does not exist
 	// yet on ZooKeeper.
 	CreateParentNodes(path string) error
+	// Stop should properly shutdown the client implementation
+	Stop()
 }
 
 type clientConfig struct {
+	logger          log.Logger
 	acl             []zk.ACL
+	credentials     []byte
 	connectTimeout  time.Duration
 	sessionTimeout  time.Duration
 	rootNodePayload [][]byte
+	eventHandler    func(zk.Event)
 }
 
-type configOption func(c *clientConfig) error
+// Option functions enable friendly APIs.
+type Option func(*clientConfig) error
 
 type client struct {
 	*zk.Conn
 	clientConfig
-	Eventc <-chan zk.Event
+	active bool
 	quit   chan struct{}
 }
 
-// WithACL returns a configOption specifying a non-default ACL.
-func WithACL(acl []zk.ACL) configOption {
+// ACL returns an Option specifying a non-default ACL for creating parent nodes.
+func ACL(acl []zk.ACL) Option {
 	return func(c *clientConfig) error {
 		c.acl = acl
 		return nil
 	}
 }
 
-// WithConnectTimeout returns a configOption specifying a non-default connection
-// timeout when we try to establish a connection to a ZooKeeper server.
-func WithConnectTimeout(t time.Duration) configOption {
+// Credentials returns an Option specifying a user/password combination which
+// the client will use to authenticate itself with.
+func Credentials(user, pass string) Option {
+	return func(c *clientConfig) error {
+		if user == "" || pass == "" {
+			return ErrInvalidCredentials
+		}
+		c.credentials = []byte(user + ":" + pass)
+		return nil
+	}
+}
+
+// ConnectTimeout returns an Option specifying a non-default connection timeout
+// when we try to establish a connection to a ZooKeeper server.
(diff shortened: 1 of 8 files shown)


loadbalancer/zk/logwrapper.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/samuel/go-zookeeper/zk"
+)
+
+type wrapLogger struct {
+	log.Logger
+}
```
peterbourgon: Is this only used for the Printf convenience function? If so, please remove it, and use `logger.Log("msg", "...")` directly.

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Client is a wrapper arround the zookeeper client.
+type Client interface {
+	// GetEntries queries the path in zookeeper, places a watch on it
+	// and retrieves data from its current child nodes
+	GetEntries(path string) ([]string, <-chan zk.Event, error)
+	// createParentNodes tries to create the path
+	// in case it does not exist yet on Zookeeper
+	createParentNodes(path string) error
+}
```
peterbourgon: I've never seen an exported interface with both exported and nonexported methods. Do you have some precedent for that? Unless there's something I'm overlooking, I think it must be a mistake. If the interface exists because you intend other packages to implement Client, the nonexported method prevents that. And if you intend the nonexported method to be for package zk internal usage only, then there's no reason to export Client...

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+
+type client struct {
+	*zk.Conn
+	evtConn <-chan zk.Event
```
peterbourgon: evtChan? eventc would be the more-canonical naming.

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+	acl     []zk.ACL
+	quit    chan struct{}
+}
+
+// NewClient returns a *zk.Client with a connection to the named server cluster.
+// It will return an error if the server cluster cannot be resolved.
+func NewClient(servers []string, acl *[]zk.ACL, logger log.Logger) (Client, error) {
```
peterbourgon: Why is acl a pointer-to-slice?

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+
```
peterbourgon: Please remove all extraneous newlines at the beginning of function implementation.

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+	conn, evt, err := zk.Connect(servers, 5*time.Second, withLogger(logger))
```
peterbourgon: What is 5s? Should it be parameterized? If not, can you make it a package-global constant?

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if acl == nil {
+		// no ACL was given, use most permissive ACL possible
+		defaultAcl := zk.WorldACL(zk.PermAll)
+		acl = &defaultAcl
+	}
```
peterbourgon: I think there's no reason to allow this parameter to be optional. Clients should be able to pass `zk.WorldACL(zk.PermAll)` if they don't care about permissions.

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return &client{conn, evt, *acl, make(chan struct{})}, nil
+}
+
+// createParentNodes implements the zookeeper Client interface.
```
peterbourgon: It's a minor nit, but please `s/zookeeper/ZooKeeper/g` in comments :)

loadbalancer/zk/logwrapper.go
```diff
@@ -0,0 +1,21 @@
+package zk
+
+import (
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/samuel/go-zookeeper/zk"
+)
```
peterbourgon: In this and other files, please group imports as in other packages:

```go
import (
    // stdlib

    // third-party

    // go-kit/kit/*
)
```

loadbalancer/zk/publisher.go
```diff
@@ -0,0 +1,84 @@
...
+type Publisher struct {
+	client Client
+	path   string
+	cache  *loadbalancer.EndpointCache
+	logger log.Logger
+	quit   chan struct{}
+}
+
+// NewPublisher returns a zookeeper publisher. Zookeeper will start watching the given
+// path for changes and update the Publisher endpoints.
```
peterbourgon: Can't quite tell in the GitHub interface, but if it's not already true, please wrap all comments at 80col.

peterbourgon:
Cool! I haven't tried it out yet, but it looks like a solid start. Please address those comments and I think you're good to start writing some tests :) Can you link me to a ZooKeeper quick-start guide so I can try it in a live environment? I've had not-so-much success getting ZK going in the past :sweat:

loadbalancer/zk/logwrapper.go
basvanbeek: No it's not a convenience function. wrapLogger wraps a go-kit logger and adds the Printf method so it can be used as the preferred logger in the underlying ZooKeeper library.

loadbalancer/zk/client.go
basvanbeek: Refactoring mistake, The method should have been exported. Thinking about it, I'm not entirely sure if it should have an interface at all but will keep that in so people are allowed to provide an alternative ZooKeeper client implementation in case it suits them better. Given that some of the other loadbalancer implementations have exported Clients too I'll keep it in.

loadbalancer/zk/client.go
basvanbeek: Old habit in naming logic dating from my Delphi days at a certain company I worked for. I'll rename.

loadbalancer/zk/logwrapper.go
peterbourgon: Ah, OK! Can you put a comment to that effect on the struct?

basvanbeek:
I just updated the ZooKeeper Publisher taking into account the remarks. Next up is a publisher test with a mock Client.

peterbourgon:
:tada:

basvanbeek:
Ok I fixed the test environment for both Travis and CircleCI to include testing against a real ZooKeeper server. Tomorrow I will squash the commits I made to get it to work so it will be a clean(er) pull request

basvanbeek:
Squashed the commits... Travis fails sometimes on one of the circuit breaker tests so it probably needs some examination:

    --- FAIL: TestHystrix-2 (0.01s)
    	util_test.go:51: hystrix_test.go:34: want "hystrix: circuit open", have "tragedy+disaster"
    FAIL
    FAIL	github.com/go-kit/kit/circuitbreaker	0.036s

peterbourgon:
Ah, that's frustrating. I've seen it before. I'll look at it and let you know when you can merge from master..

peterbourgon:
Can you try rebasing on master?

basvanbeek:
Done.

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,157 @@
...
+	// SessionTimeout is the default timeout to keep the current ZooKeeper
+	// session alive during a temporary disconnect.
+	SessionTimeout = 5 * time.Second
+)
+
+// Client is a wrapper around a lower level ZooKeeper client implementation.
+type Client interface {
+	// GetEntries should query the provided path in ZooKeeper, place a watch on
+	// it and retrieve data from its current child nodes.
+	GetEntries(path string) ([]string, channels.SimpleOutChannel, error)
```
peterbourgon: I initially thought the channels dep was to satisfy some other package's interface, but upon investigation it looks like not. Is there a strong reason why you wrap the chans with this package?

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,157 @@
...
+	sessionTimeout  time.Duration
+	rootNodePayload [][]byte
+}
+
+type configOption func(c *clientConfig) error
+
+type client struct {
+	*zk.Conn
+	clientConfig
+	Eventc <-chan zk.Event
```
peterbourgon: Why capitalized?

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+type clientConfig struct {
+	acl             []zk.ACL
+	connectTimeout  time.Duration
+	sessionTimeout  time.Duration
+	rootNodePayload [][]byte
+}
+
+type configOption func(c *clientConfig) error
```
peterbourgon: In keeping with other packages, this can be just Option, i.e. `zk.Option`

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil
+	}
+}
+
+// WithConnectTimeout returns a configOption specifying a non-default connection
+// timeout when we try to establish a connection to a ZooKeeper server.
+func WithConnectTimeout(t time.Duration) configOption {
+	return func(c *clientConfig) error {
+		if t.Seconds() < 1 {
+			return errors.New("Invalid Connect Timeout. Minimum value is 1 second")
```
peterbourgon: "[Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)."

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,157 @@
...
+
+type client struct {
+	*zk.Conn
+	clientConfig
+	Eventc <-chan zk.Event
+	quit   chan struct{}
+}
+
+// WithACL returns a configOption specifying a non-default ACL.
+func WithACL(acl []zk.ACL) configOption {
```
peterbourgon: In keeping with other packages, option functions can just be named after the thing they set, i.e. this can just be ACL, WithConnectTimeout can just be ConnectTimeout, etc.

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,157 @@
...
+// It will return an error if the server cluster cannot be resolved.
+func NewClient(servers []string, logger log.Logger, options ...configOption) (Client, error) {
+	config := clientConfig{
+		acl:            DefaultACL,
+		connectTimeout: ConnectTimeout,
+		sessionTimeout: SessionTimeout,
+	}
+	for _, option := range options {
+		if err := option(&config); err != nil {
+			return nil, err
```
peterbourgon: Nice idiom, in that it returns the error in the natural place in the callstack.

loadbalancer/zk/publisher.go
```diff
@@ -0,0 +1,83 @@
...
+	p := &Publisher{
+		client: c,
+		path:   path,
+		cache:  loadbalancer.NewEndpointCache(f, logger),
+		logger: logger,
+		quit:   make(chan struct{}),
+	}
+
+	// try to create path nodes if they are not available
+	p.client.CreateParentNodes(p.path)
```
peterbourgon: This returns an error, is it safe to ignore?

loadbalancer/zk/client_test.go
```diff
@@ -0,0 +1,149 @@
...
+	if want, have := 1, len(endpoints); want != have {
+		t.Errorf("want %q, have %q", want, have)
+	}
+}
+
+func startCluster(t *testing.T) (*stdzk.TestCluster, []string) {
+	// Start ZooKeeper Test Cluster
+	ts, err := stdzk.StartTestCluster(1, nil, logWriter{t: t, p: "[ZKERR] "})
+	if err != nil {
+		t.Fatal(err)
```
peterbourgon: Hmm. This will fail if the server jar isn't on the filesystem, which we can't expect to be true in the general case of `go test ./...`. It looks like these tests are really just integration tests, verifying the Client is interacting properly with a real ZooKeeper. Is that true? If so, can we put them behind an `integration` build tag, and update .travis.yml accordingly?

Or, given that no other Go kit component that interacts with a third-party software has integration tests at the moment, I'm also fine with leaving them out.

loadbalancer/zk/client.go
basvanbeek: To allow for a generic Client interface so people can swap out the existing client implementation in favor of another and as long as it will satisfy the Client interface no code change is needed for the ZooKeeper publisher. The channels package has decent wrap and unwrap functions and other channel helper functions so I used that instead of rolling my own. If we don't want the ability to swap implementations the entire Client interface could be left out but we end up with a tightly coupled Publisher.

loadbalancer/zk/client.go
basvanbeek: I need to change that to not be exported. I will add a configuration option to allow users to provide a session event handler as they might want to know about ZooKeeper cluster updates and provide a default handler just logging the event.

loadbalancer/zk/client.go
basvanbeek: Reason I did not export is so we can have an alternative client implementation in the same package if people would favor one ZooKeeper client library over another. Not sure if it is desirable though.

loadbalancer/zk/client_test.go
basvanbeek: The client tests are mixed at the moment. I will separate the integration tests from the unit tests. This will also allow us to revert the travis config file and remove the circleci configuration file.

loadbalancer/zk/client.go
peterbourgon: I think that is premature optimization; I would much rather see a plain `chan` type here. And I don't think that obviates the Client interface; you use it to good effect when mocking out ZooKeeper in the other test file, right?

loadbalancer/zk/client_test.go
peterbourgon: :+1:

loadbalancer/zk/client.go
basvanbeek: That is possible but we'd lose the ability to unwrap the underlying library's node event channel returned by GetEntries if used outside of the Publisher.

basvanbeek:
Updates taking into account previous comments. Fixed issue with how I was handling node ACLs and Client Authentication.

loadbalancer/zk/client.go
peterbourgon: I don't understand. The underlying library yields `[<-chan zk.Event](https://godoc.org/github.com/samuel/go-zookeeper/zk#Connect)` directly. I'm asking you to just pass it through. Are you intending to abstract over go-zookeeper/zk.Event and potentially another ZooKeeper library's Event type? If so, please don't! :)

loadbalancer/zk/client.go
```diff
@@ -0,0 +1,250 @@
...
+// inspected.
+type Event struct {
+	payload interface{}
+}
+
+// Client is a wrapper around a lower level ZooKeeper client implementation.
+type Client interface {
+	// GetEntries should query the provided path in ZooKeeper, place a watch on
+	// it and retrieve data from its current child nodes.
+	GetEntries(path string) ([]string, <-chan Event, error)
```
peterbourgon: OK, I see you've updated everything here. Thanks.

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+var (
+	host []string
+)
+
+func TestMain(m *testing.M) {
+	fmt.Println("ZooKeeper Integration Test Initializing. Starting ZooKeeper Server...")
+	ts, err := stdzk.StartTestCluster(1, nil, nil)
+	host = []string{fmt.Sprintf("localhost:%d", ts.Servers[0].Port)}
```
peterbourgon: Need to check err before using ts.Servers

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		fmt.Printf("Unable to start ZooKeeper Server: %s", err)
```
peterbourgon: - Need \n with fmt.Printf
- Use %v for errors

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	stdzk "github.com/samuel/go-zookeeper/zk"
+)
+
+var (
+	host []string
+)
+
+func TestMain(m *testing.M) {
+	fmt.Println("ZooKeeper Integration Test Initializing. Starting ZooKeeper Server...")
+	ts, err := stdzk.StartTestCluster(1, nil, nil)
```
peterbourgon: defer ts.Stop somewhere?

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+		t.Fatal(err)
+	}
+	defer p.Stop()
+
+	endpoints, err := p.Endpoints()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want, have := 0, len(endpoints); want != have {
+		t.Errorf("want %q, have %q", want, have)
```
peterbourgon: %d?

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func TestCreateParentNodesOnServer(t *testing.T) {
+	payload := [][]byte{[]byte("Payload"), []byte("Test")}
+	c1, err := NewClient(host, logger, Payload(payload))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if c1 == nil {
+		t.Error("expected pointer to client, got nil")
```
peterbourgon: t.Fatal?

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+		t.Errorf("want %q, have %q", want, have)
+	}
+
+	c2, err := NewClient(host, logger)
+	if err != nil {
+		t.Fatalf("Connect returned error: %+v", err)
+	}
+	defer c2.Stop()
+	c2impl, _ := c2.(*client)
+	data, _, err := c2impl.Get(path)
```
peterbourgon:
```go
data, _, err := c2.(*client).Get(path)
```

loadbalancer/zk/publisher_test.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (c *fakeClient) SendErrorOnWatch() {
+	c.result = false
+	c.triggerWatch()
+}
+
+func (c *fakeClient) Stop() {}
+
+func NewFactory(fakeError string) loadbalancer.Factory {
```
peterbourgon: Please don't export that.

loadbalancer/zk/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	c, err := NewClient(host, logger)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, eventc, err := c.GetEntries(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cimpl, _ := c.(*client)
+	_, err = cimpl.Create(
```
peterbourgon: Similarly to above,

```go
_, err := c.(*client).Create(...)
```

peterbourgon: Also, do you need to check this error? If not, you can just not assign it.

loadbalancer/zk/client.go
basvanbeek: Yes that was my exact idea so we have a loose coupling and ability to have an interface that could be implemented using other underlying libraries. If we don't want that we can skip having an interface all together as there is only one client implementation compatible with the publisher.

peterbourgon:
Thanks for enduring the gauntlet of my pedantry :) Happy to merge once those are settled.

loadbalancer/zk/client.go
peterbourgon: As we discussed on Slack, I think the interface is still valuable for testing. See the [consul.Client interface](https://github.com/go-kit/kit/blob/d81f9d953781636e3fd67c9fc4eab1210f3437b3/loadbalancer/consul/client.go#L6-L8) which is [accepted by the Publisher](https://github.com/go-kit/kit/blob/d81f9d953781636e3fd67c9fc4eab1210f3437b3/loadbalancer/consul/publisher.go#L32) and [faked for the tests](https://github.com/go-kit/kit/blob/d81f9d953781636e3fd67c9fc4eab1210f3437b3/loadbalancer/consul/publisher_test.go#L164).

peterbourgon:
Thank you!

PR #168, opened 2015-12-18
Fix grammar in README.md

carlmjohnson:


peterbourgon:
It should be "let's use a middleware, also known as a decorator."

carlmjohnson:
I agree.

carlmjohnson:
Updated.

peterbourgon:
Thanks!

PR #172, opened 2016-01-15
Add SetBufferedStream() ClientOption

### LINKED ISSUE #170: Support response stream from transport/http.Client
The `endpoint.Endpoint` produced by `transport/http.Client.Endpoint()` has a defer that closes the HTTP response body.
https://github.com/go-kit/kit/blob/master/transport/http/client.go#L79
Normally this is fine, especially when dealing with succinct messages encoded as JSON.

However, I have a use case where some (but not all) endpoints on a service are returning binary data (PDFs, to be specific).
I don't want to buffer the whole byte stream if I don't need to (eg. If I'm just going to pipe it to another output stream), which means I need to call response.Body.Close() myself after the endpoint has returned.

I've got a solution that involves adding a `stream bool` flag to the `Client` struct, changing the `Endpoint` method to wrap the defer call in a `if !c.stream { }` block, and then creating an Option function which sets `c.stream = true`.
That keeps everything backwards compatible while enabling streaming when required.

```go
// transport/http/client.go

type Client struct {
  // ...
  stream bool
}

// StreamClient prevents resp.Body.Close() from being called by the endpoint
func StreamClient() ClientOption {
	return func(c *Client) { c.stream = true }
}

func (c Client) Endpoint() endpoint.Endpoint {
  // ...
  if !c.stream {
    defer func() { _ = resp.Body.Close() }()
  }
 // ...
```

If that sounds reasonable, I'll write up a test and submit a PR.

rolaveric:
Add ClientOption "SetBufferedStream()" transport/http, which prevents Endpoint
from closing the http.Response.Body.
Fixes #170

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -13,23 +13,25 @@ import (
 
 // Client wraps a URL and provides a method that implements endpoint.Endpoint.
 type Client struct {
-	client *http.Client
-	method string
-	tgt    *url.URL
-	enc    EncodeRequestFunc
-	dec    DecodeResponseFunc
-	before []RequestFunc
+	client         *http.Client
+	method         string
+	tgt            *url.URL
+	enc            EncodeRequestFunc
+	dec            DecodeResponseFunc
+	before         []RequestFunc
+	bufferedStream bool
 }
 
 // NewClient returns a
 func NewClient(method string, tgt *url.URL, enc EncodeRequestFunc, dec DecodeResponseFunc, options ...ClientOption) *Client {
 	c := &Client{
-		client: http.DefaultClient,
-		method: method,
-		tgt:    tgt,
-		enc:    enc,
-		dec:    dec,
-		before: []RequestFunc{},
+		client:         http.DefaultClient,
+		method:         method,
+		tgt:            tgt,
+		enc:            enc,
+		dec:            dec,
+		before:         []RequestFunc{},
+		bufferedStream: false,
 	}
 	for _, option := range options {
 		option(c)
@@ -50,6 +52,12 @@ func SetClient(client *http.Client) ClientOption {
 // request before it's invoked.
 func SetClientBefore(before ...RequestFunc) ClientOption {
 	return func(c *Client) { c.before = before }
+}
+
+// SetBufferedStream sets whether the Response.Body is left open, allowing it
+// to be read from later. Useful for transporting a file as a buffered stream.
+func SetBufferedStream(buffered bool) ClientOption {
+	return func(c *Client) { c.bufferedStream = buffered }
 }
 
 // Endpoint returns a usable endpoint that will invoke the RPC specified by
@@ -76,7 +84,9 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		if err != nil {
 			return nil, fmt.Errorf("Do: %v", err)
 		}
-		defer func() { _ = resp.Body.Close() }()
+		if !c.bufferedStream {
+			defer resp.Body.Close()
+		}
 
 		response, err := c.dec(resp)
 		if err != nil {
--- a/transport/http/client_test.go
+++ b/transport/http/client_test.go
@@ -1,6 +1,7 @@
 package http_test
 
 import (
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -12,10 +13,20 @@ import (
 	httptransport "github.com/go-kit/kit/transport/http"
 )
 
+type TestResponse struct {
+	Body   io.ReadCloser
+	String string
+}
+
 func TestHTTPClient(t *testing.T) {
 	var (
-		encode    = func(*http.Request, interface{}) error { return nil }
-		decode    = func(*http.Response) (interface{}, error) { return struct{}{}, nil }
+		testbody = "testbody"
+		encode   = func(*http.Request, interface{}) error { return nil }
+		decode   = func(r *http.Response) (interface{}, error) {
+			buffer := make([]byte, len(testbody))
+			r.Body.Read(buffer)
+			return TestResponse{r.Body, string(buffer)}, nil
+		}
 		headers   = make(chan string, 1)
 		headerKey = "X-Foo"
 		headerVal = "abcde"
@@ -24,6 +35,7 @@ func TestHTTPClient(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		headers <- r.Header.Get(headerKey)
 		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(testbody))
 	}))
 
 	client := httptransport.NewClient(
@@ -34,7 +46,7 @@ func TestHTTPClient(t *testing.T) {
 		httptransport.SetClientBefore(httptransport.SetRequestHeader(headerKey, headerVal)),
 	)
 
-	_, err := client.Endpoint()(context.Background(), struct{}{})
+	res, err := client.Endpoint()(context.Background(), struct{}{})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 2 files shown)


transport/http/client.go
```diff
@@ -52,6 +54,12 @@ func SetClientBefore(before ...RequestFunc) ClientOption {
 	return func(c *Client) { c.before = before }
 }
 
+// BufferedStreamClient stops the Endpoint from automatically closing the response.Body
```
peterbourgon: Proper punctuation, please :)

transport/http/client.go
```diff
@@ -52,6 +54,12 @@ func SetClientBefore(before ...RequestFunc) ClientOption {
...
+// Useful for passing around large byte streams, like files.
+func BufferedStreamClient() ClientOption {
```
peterbourgon: In keeping with the other functions,

```go
func SetBufferedStream(buffered bool) ClientOption {
    return func(c *Client) { c.bufferedStream = buffered }
}
```

transport/http/client.go
```diff
@@ -76,7 +84,9 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		if err != nil {
 			return nil, fmt.Errorf("Do: %v", err)
 		}
-		defer func() { _ = resp.Body.Close() }()
+		if !c.bufferedStream {
+			defer func() { _ = resp.Body.Close() }()
```
peterbourgon: I think you can just `defer resp.Body.Close()`

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
 	}
+	// Check that Request Header was successfully received
 	if want := headerVal; want != have {
 		t.Errorf("want %q, have %q", want, have)
 	}
+
+	// Check that the response was successfully decoded
+	response, ok := res.(TestResponse)
+	if !ok {
+		t.Errorf("response should be TestResponse")
```
peterbourgon: t.Fatal here, and de-indent the else block.

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+	} else {
+		have = response.String
+		if want := "testbody"; want != have {
```
peterbourgon: Please make `"testbody"` a function-scoped variable.

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+			t.Errorf("want %q, have %q", want, have)
+		}
+
+		// Check that response body was closed
+		b := make([]byte, 1)
+		_, err = response.Body.Read(b)
+		if want := "http: read on closed response body"; err == nil || err.Error() != want {
```
peterbourgon: Where did this text come from? Is it unexported? :\

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+			t.Errorf("want %q, have %q", want, err)
+		}
```
peterbourgon:
```go
_, err = response.Body.Read(b)
if err == nil {
    t.Fatal("wanted error, got none")
}
if want, have := "http: read on closed response body", err.Error(); want != have {
}
```

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+	}
+}
+
+func TestHTTPClient_BufferedStream(t *testing.T) {
```
peterbourgon: `TestHTTPClientBufferedStream`

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+	res, err := client.Endpoint()(context.Background(), struct{}{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check that the response was successfully decoded
+	response, ok := res.(TestResponse)
+	if !ok {
+		t.Errorf("response should be TestResponse")
+	} else {
```
peterbourgon: Likewise above, t.Fatal and de-indent the else block.

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+		// Check that response body was NOT closed
+		buffer := make([]byte, 8)
+		_, err = response.Body.Read(buffer)
+		if want := io.EOF; err != want {
+			t.Errorf("want %q, have %q", want, err)
+		} else {
```
peterbourgon: Same as above, please.

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+	// Check that the response was successfully decoded
+	response, ok := res.(TestResponse)
+	if !ok {
+		t.Errorf("response should be TestResponse")
+	} else {
+		// Check that response body was NOT closed
+		buffer := make([]byte, 8)
+		_, err = response.Body.Read(buffer)
+		if want := io.EOF; err != want {
+			t.Errorf("want %q, have %q", want, err)
```
peterbourgon:
```go
b := make([]byte, 8)
_, err = response.Body.Read(b)
if want, have := io.EOF, err; want != have {
    t.Errorf("want %v, have %v", want, have)
}
```

transport/http/client_test.go
```diff
@@ -45,9 +56,73 @@ func TestHTTPClient(t *testing.T) {
...
+		} else {
+			have := string(buffer)
+			if want := "testbody"; want != have {
+				t.Errorf("want %q, have %q", want, have)
+			}
+		}
+	}
```
peterbourgon:
```go
testbody := "testbody"

// ...

b := make([]byte, len(testbody))
_, err = response.Body.Read(b)
if want, have := io.EOF, err; want != have {
    t.Fatalf("want %v, have %v", want, have)
}
if want, have := testbody, string(b); want != have {
}
```

transport/http/client_test.go
rolaveric: Yeah, that's a weird one. Going by the doco, I'd expect it to be [`http.ErrBodyReadAfterClose`](https://golang.org/pkg/net/http/#pkg-variables), but it's an unexported error.
[net/http/transport.go#L1290](https://golang.org/src/net/http/transport.go#L1290)
Still, you're right, that will make the test fragile so it would be better to test that it's not nil or io.EOF.

transport/http/client_test.go
```diff
@@ -12,10 +13,20 @@ import (
...
+}
+
 func TestHTTPClient(t *testing.T) {
 	var (
-		encode    = func(*http.Request, interface{}) error { return nil }
-		decode    = func(*http.Response) (interface{}, error) { return struct{}{}, nil }
+		testbody = "testbody"
+		encode = func(*http.Request, interface{}) error { return nil }
+		decode = func(r *http.Response) (interface{}, error) {
+			buffer := make([]byte, 8)
```
peterbourgon: Should this 8 be len(testbody)?

transport/http/client_test.go
```diff
@@ -45,9 +57,73 @@ func TestHTTPClient(t *testing.T) {
...
+	if doNotWant, have := io.EOF, err; doNotWant == have {
+		t.Errorf("do not want %q, have %q", doNotWant, have)
+	}
+}
+
+func TestHTTPClientBufferedStream(t *testing.T) {
+	var (
+		testbody = "testbody"
+		encode = func(*http.Request, interface{}) error { return nil }
+		decode = func(r *http.Response) (interface{}, error) {
```
peterbourgon: Did you run gofmt on this? Strange indentation outcome.

peterbourgon:
Except for those 2 little things, LGTM! Thank you very much for the contribution.

rolaveric:
Should be good now. Forgot to set up my gofmt watcher on this machine :P

peterbourgon:
:tada:

PR #173, opened 2016-01-29
Updated readme with more example usage.

taion809:
Added some more examples including timestamp, caller, and example output.


ChrisHines:
Thanks!

PR #177, opened 2016-02-04
Add support for metrics.Histogram.Distribution

peterbourgon:
Histograms gain a method to read the distribution (slice of buckets) that has been observed so far. In this PR, the only implementation that supports Distribution is expvar (via codahale/hdrhistogram). Prometheus support is possible and planned.

- Each metrics type gains a Name() method
- metrics.Histogram gains Distribution()
- metrics package gains PrintDistribution()
- Minor updates to README


adrianco:
Yes, this looks good to me, many thanks for getting it done so quickly.

PR #180, opened 2016-02-06
An example REST-y service, with gorilla/mux

peterbourgon:
Let me know what you think!


ae6rt:
Peter, this is beautiful.  This suggested pattern of weaving a well-known http router into go-kit will bring an entirely new group of developers to the go-kit.  (github star predictions for six months from now?) Love it.

ae6rt:
Once I start working with this, one of my first tasks is to see how OAuth2-secured endpoints works with this pattern.  And to explore whether business endpoints can remain unaware that a middleware has allowed them to run because the middleware knows consumption of the specific endpoint is allowed by the provided OAuth2 token scope(s).

peterbourgon:
Yup. My guess — some business middleware will look for the presence of auth keys in the context.Context, and error out if it doesn't find them. Then it'd be the responsibility of a transport middleware (e.g. http.BeforeFunc, or something more custom) to extract information from the transport and populate the context.

ae6rt:
btw, I could not tell:  were you kidding when you said you were going to work on code generation? :)

peterbourgon:
Yes, it's definitely the next thing I have to dive into. No ETA :)

PR #182, opened 2016-02-08
gofmt -s -w

peterbourgon:
And other things to improve the Go Report Card.


peterbourgon:
@ChrisHines [From go vet](https://goreportcard.com/report/github.com/go-kit/kit#go_vet):

```
github.com/go-kit/kit/log/log_test.go
 Line 74: unrecognized printf verb 'n'
 Line 85: unrecognized printf verb 'n'
```

What is meant there, do you think?

ae6rt:
There is no such verb as "%n"?  https://golang.org/pkg/fmt/

ChrisHines:
This is why go vet is sometimes wrong. Type [`gopkg.in/stack.v1.Call`](https://godoc.org/gopkg.in/stack.v1#Call.Format) implements the [`fmt.Formatter`](https://golang.org/pkg/fmt/#Formatter) interface. It mostly responds to standard formatting verbs that go vet is happy with, but it introduces the %n verb for *function name* because none of the traditional verbs made sense and it already had uses for %s and %v.

I have considered filing an issue against go vet along the lines of "don't flag formatting verbs if the corresponding argument implements `fmt.Formatter`", but I have never done so because there are so few implementations of `fmt.Formatter` in the wild that it hardly seems worth the effort.

peterbourgon:
Understood. I guess it is an unfortunate impedance mismatch. Thank you for the patient explanation.

ae6rt:
I see it now.  Sorry for the noise.

PR #183, opened 2016-02-09
Small doc cleanups

freeformz:
* Link to gopher slack signup has changed.
* It's `godep` not `Godep`.


peterbourgon:
Thanks. I think it's time to stop recommending godep, but I'll do that in a separate commit.

freeformz:
@peterbourgon I suspect I know why, but am wondering what your reasons are to stop recommending godep?

peterbourgon:
I could type at length about what I perceive to be the UX failures of Godep, but I think the tl;dr is we have GO15VENDOREXPERIMENT now, and that's what everyone should be using.

PR #189, opened 2016-02-13
add LTSV logger

kazegusuri:
LTSV is a structured logging format like JSON. Some of loggers support LTSV format. This is a LTSV logger for go-kit logger.
See also [ltsv.org](http://ltsv.org)

### DIFF
--- a/log/ltsv_logger.go
+++ b/log/ltsv_logger.go
@@ -1,0 +1,80 @@
+package log
+
+import (
+	"fmt"
+	"github.com/najeira/ltsv"
+	"io"
+	"reflect"
+)
+
+type ltsvLogger struct {
+	w io.Writer
+}
+
+// NewLTSVLogger returns a Logger that encodes keyvals to the Writer as
+// Labeled Tab-separated Values format.
+func NewLTSVLogger(w io.Writer) Logger {
+	return &ltsvLogger{
+		w: w,
+	}
+}
+
+func (l *ltsvLogger) Log(keyvals ...interface{}) error {
+	n := (len(keyvals) + 1) / 2 // +1 to handle case when len is odd
+	m := make(map[string]string, n)
+	for i := 0; i < len(keyvals); i += 2 {
+		k := keyvals[i]
+		var v interface{} = ErrMissingValue
+		if i+1 < len(keyvals) {
+			v = keyvals[i+1]
+		}
+		mergeStringMap(m, k, v)
+	}
+
+	w := ltsv.NewWriter(l.w)
+	if err := w.Write(m); err != nil {
+		return err
+	}
+	w.Flush()
+	return nil
+}
+
+func mergeStringMap(dst map[string]string, k, v interface{}) {
+	var key string
+	switch x := k.(type) {
+	case string:
+		key = x
+	case fmt.Stringer:
+		key = safeString(x)
+	default:
+		key = fmt.Sprint(x)
+	}
+
+	var val string
+	switch x := v.(type) {
+	case string:
+		val = x
+	case error:
+		val = safeErrorString(x)
+	case fmt.Stringer:
+		val = safeString(x)
+	default:
+		val = fmt.Sprint(x)
+	}
+
+	dst[key] = val
+}
+
+func safeErrorString(err error) (s string) {
+	defer func() {
+		if panicVal := recover(); panicVal != nil {
+			if v := reflect.ValueOf(err); v.Kind() == reflect.Ptr && v.IsNil() {
+				s = ""
+			} else {
+				panic(panicVal)
+			}
+		}
+	}()
+	s = err.Error()
+	return
+}
--- a/log/ltsv_logger_test.go
+++ b/log/ltsv_logger_test.go
@@ -1,0 +1,78 @@
+package log_test
+
+import (
+	"bytes"
+	"errors"
+	"io/ioutil"
+	"sort"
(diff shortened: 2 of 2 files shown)


ChrisHines:
This is pretty cool. I was not aware of LTSV. It seems superficially similar to the logfmt format but with different delimiters. Based on the website you linked it looks like it has decent support in other tools and language, so I think this will be a nice addition.

This PR does introduce [github.com/najeira/ltsv](https://github.com/najeira/ltsv) as a new dependency to package log. How confident can we be in the stability of this new dependency? Do we think it conforms to the [Go kit API Stability Policy](https://github.com/go-kit/kit/blob/master/rfc/rfc007-api-stability.md)?

kazegusuri:
> This PR does introduce github.com/najeira/ltsv as a new dependency to package log. How confident can we be in the stability of this new dependency? Do we think it conforms to the Go kit API Stability Policy?

Good point. It's easy to implement ltsv reader/writer. Most middleware include built-in implementation. There is not widely used famous ltsv package as a result. github.com/najeira/ltsv is one of the packages, but has more functionalities than others.
Actually the required functionality of this logger is very small. I can implement ltsv logger without new dependency with small code base.  That's one option.

ChrisHines:
My experience writing LogfmtLogger, which is backed by [gopkg.in/logfmt.v0](https://gopkg.in/logfmt.v0), the serialization code is strait forward. The challenge lies in handling all the different keys and values that could be passed in an `interface{}`. That package would serve as an excellent template for implementing a robust LSTV encoder that would work well with Go kit log.

ChrisHines:
@kazegusuri Do you plan to do any more work on this? Go-kit has dropped its API stability policy, so we have some more flexibility than before with regards to dependencies.

But I have a new concern. The LTSV FAQ says this:

>#### Doesn’t it have to have “escaping” in its spec?
>The spec of LTSV are just only not using “:” in the keys and separating by tab. There are some reasons why escaping is not in the spec.
>
>* Strictly considering escaping makes parsing difficult.
>* Aren’t there cases when tabs appear in the strings like User-Agent? No, there aren’t, because of vulnerability management.

Perhaps this approach is OK when formatting httpd access logs (which is where LTSV seems to have originated), but it is not acceptable for general purpose structured logging with arbitrary values.

peterbourgon:
@kazegusuri Ping, any update? :)

kazegusuri:
Sorry for late reply. I haven't touched from the last comment.

> Perhaps this approach is OK when formatting httpd access logs (which is where LTSV seems to have originated), but it is not acceptable for general purpose structured logging with arbitrary values.

Right.  Escaping is not in the spec, so it depends on implementation. Almost all libraries for ltsv don't handle escaping.  Then this logger is not truly general purpose, but it still useful for access logs and application logs not including user input values.

I haven't caught up with a new policy, is there something need to do to this PR?

ChrisHines:
My opinion is that since LTSV cannot handle arbitrary values then we should not include it in the main logging package. It could easily be made into a separate package that implemented the `log.Logger` interface. I'm not sure if that package should be in the Go kit repo or not. If it was in a separate repo maybe we could link to it from the README in a section listing additional `log.Logger` implementations.

What do you think @peterbourgon?

peterbourgon:
I think there is room in kit/log for logger implementations that are not general-purpose. In particular (unrelated to this PR) I would be very happy to see a restricted logger for event-source (in the e.g. lambda architecture) data; what that looks like in practice is still a bit fuzzy to me. I haven't heard of LTSV before now but based on the number of other implementations it seems a reasonable addition.

What I would want to see is some way of more formally specifying behavior re: escaping and unsupported values. Just to serve as a starting point, how about two constructors:

- NewLTSVLogger, that will attempt to coerce all reasonably detectable nonconforming keys or values to a conforming form, at the cost of some CPU overhead
- NewLTSVLoggerStrict, that will error or panic if fed nonconforming values

?

ChrisHines:
>Just to serve as a starting point, how about two constructors:
>
>* NewLTSVLogger, that will attempt to coerce all reasonably detectable nonconforming keys or values to a conforming form, at the cost of some CPU overhead
>* NewLTSVLoggerStrict, that will error or panic if fed nonconforming values

I think that is a reasonable approach.

Perhaps this is a good time to consider the impact of kit/log someday supporting a largish number of output formats. Each of those formats may bring along with it additional dependencies. At what point does that become a burden to consumers of kit/log who may end up having to vendor all those dependencies just to use one of the formats? I think we should consider moving log output format implementations into sub-directories of kit/log to make them more à la carte.

peterbourgon:
Yes, that would bring it in line with e.g. kit/metrics/{statsd,dogstatsd,prometheus,...}. +1.

peterbourgon:
Hmm. It would have the effect of changing the relatively-nice

```go
import (
    "github.com/go-kit/kit/log"
)

var logger log.Logger
logger = log.NewJSONLogger(os.Stdout)
logger = log.NewContext(logger).With("foo", 123)
```

to the slightly more awkward

```go
import (
    "github.com/go-kit/kit/log"
    jsonlog "github.com/go-kit/kit/log/json"
)

var logger log.Logger
logger = jsonlog.New(os.Stdout)
logger = log.NewContext(logger).With("foo", 123)
```

peterbourgon:
I'm pretty sure this has stalled. For the record I would still be happy to accept a PR which added this new backend.

PR #192, opened 2016-02-14
Added race detector to test runner

taion809:
Hello,
This pull request adds the go race detector to travis-ci.

From the output it looks like there are some races detected (#193)


peterbourgon:
There is no need to run the test twice; you can remove the call without the -race flag.

peterbourgon:
Whoops! Sorry, would have merged this one if I hadn't overlooked it. Thank you for the issue! :)

PR #195, opened 2016-02-15
Fixes for ZooKeeper Publisher Tests (race conditions and close panic)

basvanbeek:
This should take care of the close panic and race conditions in the mocked Client.

### DIFF
--- a/loadbalancer/zk/publisher_test.go
+++ b/loadbalancer/zk/publisher_test.go
@@ -2,7 +2,9 @@ package zk
 
 import (
 	"errors"
+	"fmt"
 	"io"
+	"sync"
 	"testing"
 	"time"
 
@@ -43,15 +45,14 @@ func TestBadFactory(t *testing.T) {
 	}
 	defer p.Stop()
 
-	endpoints, err := p.Endpoints()
-	if err != nil {
-		t.Fatal(err)
-	}
 	// instance1 came online
-	client.AddService(path+"/instance1", "zookeeper_node_data")
+	client.AddService(path+"/instance1", "kaboom")
 
-	if want, have := 0, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
+	// instance2 came online
+	client.AddService(path+"/instance2", "zookeeper_node_data")
+
+	if err = asyncTest(100*time.Millisecond, 1, p); err != nil {
+		t.Error(err)
 	}
 }
 
@@ -68,6 +69,7 @@ func TestServiceUpdate(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
 	if want, have := 0, len(endpoints); want != have {
 		t.Errorf("want %d, have %d", want, have)
 	}
@@ -75,49 +77,39 @@ func TestServiceUpdate(t *testing.T) {
 	// instance1 came online
 	client.AddService(path+"/instance1", "zookeeper_node_data")
 
-	// test if we received the instance
-	endpoints, err = p.Endpoints()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if want, have := 1, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
-	}
-
 	// instance2 came online
 	client.AddService(path+"/instance2", "zookeeper_node_data2")
 
-	// test if we received the instance
-	endpoints, err = p.Endpoints()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if want, have := 2, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
+	// we should have 2 instances
+	if err = asyncTest(100*time.Millisecond, 2, p); err != nil {
+		t.Error(err)
 	}
 
 	// watch triggers an error...
 	client.SendErrorOnWatch()
 
-	// test if we ignored the empty instance response due to the error
-	endpoints, err = p.Endpoints()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if want, have := 2, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
+	// test if error was consumed
+	if err = client.ErrorIsConsumed(100 * time.Millisecond); err != nil {
+		t.Error(err)
 	}
 
-	// instances go offline
+	// instance3 came online
+	client.AddService(path+"/instance3", "zookeeper_node_data3")
+
+	// we should have 3 instances
+	if err = asyncTest(100*time.Millisecond, 3, p); err != nil {
+		t.Error(err)
+	}
+
+	// instance1 goes offline
 	client.RemoveService(path + "/instance1")
+
+	// instance2 goes offline
 	client.RemoveService(path + "/instance2")
 
-	endpoints, err = p.Endpoints()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if want, have := 0, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
+	// we should have 1 instance
+	if err = asyncTest(100*time.Millisecond, 1, p); err != nil {
+		t.Error(err)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


loadbalancer/zk/publisher_test.go
```diff
@@ -43,13 +44,16 @@ func TestBadFactory(t *testing.T) {
 	}
 	defer p.Stop()
 
+	// instance1 came online
+	client.AddService(path+"/instance1", "zookeeper_node_data")
+	time.Sleep(1 * time.Millisecond)
```
peterbourgon: We see lots of problems on Travis when we assume 1ms is enough for the scheduler to get something to happen. Are you sure this is safe long-term?

loadbalancer/zk/publisher_test.go
```diff
@@ -148,6 +159,7 @@ type fakeClient struct {
 
 func newFakeClient() *fakeClient {
 	return &fakeClient{
+		sync.Mutex{},
```
peterbourgon: Not necessary; the zero value is usable.

loadbalancer/zk/publisher_test.go
```diff
@@ -141,6 +151,7 @@ func TestBadPublisherCreate(t *testing.T) {
 }
 
 type fakeClient struct {
+	sync.Mutex
```
peterbourgon: Do you intend the mutex to be lock/unlockable from outside of member methods? If not, please name it with an unexported name, i.e.

```
type fakeClient struct {
   mtx sync.Mutex
   // ...
}
```

loadbalancer/zk/publisher_test.go
basvanbeek: No I'm not, so we have two options:

1: longer timeout in main test goroutine to yield long enough to have publisher goroutine handle the update. this will slow down the unit test.
2: loop until expected value has been propagated or higher set timeout occurs (error). 

An unfortunate side effect from having to test asynchronous updates.

peterbourgon: I've used 2 to some success in the past. Another option, a signal chan:

```go
c := make(chan result, 1)
go func() {
    // do work
    c <- res
}()
select {
case res := <-c:
    if want, have := foo, res.value; want != have {
        t.Errorf("want %v, have %v", want, have)
    }
case <-time.After(100*time.Millisecond):
    t.Fatalf("timeout waiting for foo")
}
```

You can sometimes wrap that into a helper,

```
func within(d time.Duration, f func() error) error {
    c := make(chan error, 1)
    go func() { e <- f() }()
    select {
    case err := <-c:
        return err
    case <-time.After(d):
        return errors.New("timeout")
    }
}
```

basvanbeek:
Simplified the fakeClient, made mutex internal and removed the hardcoded sleep functions for eventual consistency testing with timeout.

loadbalancer/zk/publisher_test.go
ChrisHines: Even 2ms will likely be too short on Windows once we are on Go 1.6. From the 1.6 release notes:

>On Windows, Go programs in Go 1.5 and earlier forced the global Windows timer resolution to 1ms at startup by calling timeBeginPeriod(1). Go no longer needs this for good scheduler performance, and changing the global timer resolution caused problems on some systems, so the call has been removed.

The default timer resolution on Windows is 15.6ms. Here is an informative post about the details: https://randomascii.wordpress.com/2013/04/02/sleep-variation-investigated/.

The Go issue discussion is lengthy and informative as well: https://github.com/golang/go/issues/8687

This upcoming change also means that doing something like:

```go
start := time.Now()
// do work
elapsed := time.Since(start)
```

Will measure durations in 1/64s quanta on Windows. Which means that if *work* only takes a few milliseconds, `elapsed` will usually be zero.

Avoiding sleep like functions for tests is ideal, but when that is not practical I recommend sleeping at least 40ms so the test is more likely to work as intended on Windows.

loadbalancer/zk/publisher_test.go
```diff
@@ -174,38 +176,74 @@ func (c *fakeClient) GetEntries(path string) ([]string, <-chan zk.Event, error)
...
+	c.ch <- zk.Event{}
+}
+
+func (c *fakeClient) ErrorIsConsumed(t time.Duration) error {
+	timeout := time.After(t)
+	for {
+		select {
+		case <-timeout:
+			errMsg := fmt.Sprintf("expected error not consumed after timeout %s", t.String())
+			return errors.New(errMsg)
```
peterbourgon: fmt.Errorf?

loadbalancer/zk/publisher_test.go
```diff
@@ -174,38 +176,74 @@ func (c *fakeClient) GetEntries(path string) ([]string, <-chan zk.Event, error)
...
+func asyncTest(timeout time.Duration, want int, p *Publisher) (err error) {
+	var endpoints []endpoint.Endpoint
+	// want can never be -1
+	have := -1
+	t := time.After(timeout)
+	for {
+		select {
+		case <-t:
+			errMsg := fmt.Sprintf("want %d, have %d after timeout %s", want, have, timeout.String())
+			err = errors.New(errMsg)
```
peterbourgon: fmt.Errorf?

loadbalancer/zk/publisher_test.go
```diff
@@ -174,38 +176,74 @@ func (c *fakeClient) GetEntries(path string) ([]string, <-chan zk.Event, error)
...
+			return
+		default:
+			endpoints, err = p.Endpoints()
+			have = len(endpoints)
+			if err != nil || want == have {
+				return
+			}
+			// yield this goroutine
+			runtime.Gosched()
```
peterbourgon: Whoa, burning the CPU here. How about time.Sleep(time.Millisecond)?

loadbalancer/zk/publisher_test.go
```diff
@@ -174,38 +176,74 @@ func (c *fakeClient) GetEntries(path string) ([]string, <-chan zk.Event, error)
...
+		}
+	}
```
peterbourgon:
```go
func asyncTest(p *Publisher, want int, within time.Duration) {
    var endpoints []endpoint.Endpoint
    timeout := time.After(within)
    for {
        select {
        case <-timeout:
            return fmt.Errorf("timeout waiting for %d", want)
        default:
            endpoints, err = p.Endpoints()
            if have := len(endpoints); want == have {
                return nil
            }
            time.Sleep(time.Millisecond)
        }
    }
}
```

loadbalancer/zk/publisher_test.go
```diff
@@ -43,15 +46,15 @@ func TestBadFactory(t *testing.T) {
...
 
-	if want, have := 0, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
+	// instance2 came online
+	client.AddService(path+"/instance2", "zookeeper_node_data")
+
+	err = asyncTest(100*time.Millisecond, 1, p)
+	if err != nil {
+		t.Errorf("%v", err)
 	}
```
peterbourgon:
```go
if err = asyncTest(p, 1, 100*time.Millisecond); err != nil {
    t.Error(err)
}
```

loadbalancer/zk/publisher_test.go
```diff
@@ -75,49 +79,44 @@ func TestServiceUpdate(t *testing.T) {
...
-	endpoints, err = p.Endpoints()
+	// test if error was consumed
+	err = client.ErrorIsConsumed(100 * time.Millisecond)
 	if err != nil {
-		t.Fatal(err)
+		t.Errorf("%v", err)
 	}
-	if want, have := 2, len(endpoints); want != have {
-		t.Errorf("want %d, have %d", want, have)
+	fmt.Printf("Error has been consumed\n")
```
peterbourgon: Whoops, some debug logging left in here?

loadbalancer/zk/publisher_test.go
```diff
@@ -75,49 +79,44 @@ func TestServiceUpdate(t *testing.T) {
...
 	// watch triggers an error...
 	client.SendErrorOnWatch()
 
-	// test if we ignored the empty instance response due to the error
-	endpoints, err = p.Endpoints()
+	// test if error was consumed
+	err = client.ErrorIsConsumed(100 * time.Millisecond)
 	if err != nil {
-		t.Fatal(err)
+		t.Errorf("%v", err)
```
peterbourgon: As above, can just `t.Error(err)` (eliding these comments from now on)

basvanbeek:
making pb-linter happy ;)

peterbourgon:
:sunglasses: Thank you very much!

PR #200, opened 2016-02-29
Add DogStatsD metrics backend for go-kit

skidder:
DataDog is a metrics aggregation service that uses an extension of the StatsD protocol, dubbed DogStatsD. One of the extensions adds support for tags:
http://docs.datadoghq.com/guides/dogstatsd/#tags

This pull-request adds a metrics backend for DogStatsD that bundles tags (represented as a slice of `metrics.Field` instances) with measurements, unlike the StatsD metrics backend which does not support tags.

### DIFF
--- a/metrics/dogstatsd/dogstatsd.go
+++ b/metrics/dogstatsd/dogstatsd.go
@@ -237,12 +237,12 @@ func applyTags(value string, tags []metrics.Field) string {
 		for _, t := range tags {
 			switch tagsString {
 			case "":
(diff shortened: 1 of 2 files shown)


peterbourgon:
Nice! Thank you. I'll review shortly.

peterbourgon:
@skidder Is there a way to test this without signing up for Datadog? i.e. is there a DogStatsD **server**?

peterbourgon:
Eh, the protocol is simple enough I can read it with `nc -ulv` :)

peterbourgon:
Thanks very much for the contribution!

Issue #202, opened 2016-03-02
API gateway example


peterbourgon:
This is an extension of #91.

My thinking has evolved a bit on this after watching Chris Richardson speak about his work (http://microservices.io/). In particular my understanding of what an API gateway is has shifted, and I think there is definitely a role for Go kit there. I find it pretty straightforward how to create such a thing:

- Use transport/foo.Clients to establish endpoint.Endpoints for all services in your infra (backends)
- Use regular net/http or whatever to bind /api/foosvc/{param} handlers to those endpoints (frontend)

Let's add an example/apigateway that gateways to an example/profilesvc, example/addsvc, and example/stringsvc. Assume:

- frontend as plain net/http + gorilla/mux
- Consul for service discovery
- gRPC for communication with addsvc
- transport/http for communication with profilesvc and stringsvc

basvanbeek:
Step 2 of this could be adding Zipkin tracing to showcase tracing across these connected services and across different transports... that would be a little more challenging but I might be able to contribute after the above piece has been taken care of

marshauf:
I am new to go-kit and I am working on this to get to know the library.

peterbourgon:
@basvanbeek Sounds good to me. A `docker-compose`-style infra-in-a-box, with those Go kit services + Consul + Zipkin + Prometheus would make a compelling demo!

@marshauf Great! If you'd like, feel free to make a work-in-progress style PR, and ask questions as you go. It's the easiest way to learn :)

marshauf:
@peterbourgon I am discovering the services on-demand through http requests. Should the services be discovered on startup or demand? I am also not sure how to handle http/json on the frontend and a grpc service like addsvc, since the data requires a conversion. The simple solution would be to import the generated protobuf package and use jsonpb to convert the json data to protobuf messages. Which I don't like, because it would require recompiling the apigateway when a protobuf message changes. Any thoughts?

peterbourgon:
@marshauf re: service discovery, definitely defined on startup. I'm still pretty firmly convinced an API gateway that does dynamic service discovery and route wiring is an antipattern.

[...]

Issue #167, opened 2015-12-10
Packages registry/{consul,etcd}


bhnedo:
It would be useful to provide the service registry capability for Consul, etcd and eventually for the Eureka discovery server (if the support for this publisher is on the roadmap).

basvanbeek:
How do you envision that? Currently the publishers are architected so you can have any kind of data for your services in the discovery server. You as go-kit user provide your own factory function to convert that data in the discovery server to a go-kit endpoint. This allows completely free form service registry. By spawning a publishers' discovery client you have access to the exported functions of the underlying client library to place data into the discovery server. What that data needs to be, is completely up to you.

peterbourgon:
The publishers allow Go kit services to receive information from the source of authority, to make outgoing connections. I guess the idea here is to allow Go kit services to _publish_ information to the source of authority, to announce their presence. I need to do a bit of research to see if this is generalizable, and if it would actually be useful to bring into Go kit, or if each platform (Consul, etcd, ...) is sufficiently different that there's no point.

bhnedo:
@basvanbeek  When I say service registry, I mean the ability of the framework to expose the services, so any client capable to talk with discovery server can locate and consume them. I don't have strong experience with Go kit, but as far as I can see there is no such mechanism. I had to use Consul API to register the instance of the service, so it can be queried from the Java Spring application through Spring Cloud Consul and consumed using the well known `RestTemplate` abstraction.

	consulc, _ := consul.NewClient(consul.DefaultConfig())
	consulc.Catalog().Register(&consul.CatalogRegistration{
		Node:    "archrabbit",
		Address: "127.0.0.1",
		Service: &consul.AgentService{
			ID:      "1",
			Service: "todo-microservice",
			Port:    8000,
		},

	}, nil)
That's basically what @peterbourgon states about.

basvanbeek:
I understand what you mean by service registry, I just don't see the benefit of abstracting away or wrapping this in a generic way as the various stores we want to support can be quite different in their scope. Consul can also be used as K/V store, would we need to add the ability to control these to our interface? With for instance zookeeper the kind of data placed in its znodes can be anything as it's just byte strings. And organization of these items is done by a filesystem like tree. What about future systems and their functionality that can be used for service discovery?

[...]

PR #203, opened 2016-03-03
[WIP] apigateway example

marshauf:
This is a work-in-progress pull request as discussed in #202 .

Running consul with the provided service definition, stringsvc3 and apigateway, you can run 
```
for s in foo baar baz ; do curl -d"{\"s\":\"$s\"}" localhost:8000/api/stringsvc/uppercase ; done
```
instead of directly calling the stringsvc3 service method.

The apigateway currently only handles http json encoded calls. To support different transports I could check the consul service definition tags for values like "HTTP/JSON", "gRPC" to determine the transport and codec.

### DIFF
--- a/examples/apigateway/main.go
+++ b/examples/apigateway/main.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"bytes"
-	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
@@ -15,17 +13,19 @@ import (
 	"strings"
 	"syscall"
 
-	"golang.org/x/net/context"
-
-	"github.com/gorilla/mux"
-
 	"github.com/go-kit/kit/endpoint"
+	addsvc "github.com/go-kit/kit/examples/addsvc/client/grpc"
+	"github.com/go-kit/kit/examples/addsvc/server"
 	"github.com/go-kit/kit/loadbalancer"
 	"github.com/go-kit/kit/loadbalancer/consul"
 	log "github.com/go-kit/kit/log"
+	//grpctransport "github.com/go-kit/kit/transport/grpc"
 	httptransport "github.com/go-kit/kit/transport/http"
-
+	//proto "github.com/golang/protobuf/proto"
+	"github.com/gorilla/mux"
 	"github.com/hashicorp/consul/api"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
 )
 
 var (
@@ -69,10 +69,33 @@ func main() {
 	}
 	discoveryClient = consul.NewClient(consulClient)
 
+	// discover service stringsvc
+	uppercase, err := consul.NewPublisher(discoveryClient, routeFactory(ctx, "uppercase"), logger, "stringsvc")
+	if err != nil {
+		logger.Log("fatal", err)
+	}
+	count, err := consul.NewPublisher(discoveryClient, routeFactory(ctx, "count"), logger, "stringsvc")
+	if err != nil {
+		logger.Log("fatal", err)
+	}
+
+	// discover service addsvc
+	addsvcSum, err := consul.NewPublisher(discoveryClient, factoryAddsvc(ctx, logger, makeSumEndpoint), logger, "addsvc")
+	if err != nil {
+		logger.Log("fatal", err)
+	}
+	addsvcConcat, err := consul.NewPublisher(discoveryClient, factoryAddsvc(ctx, logger, makeConcatEndpoint), logger, "addsvc")
+	if err != nil {
+		logger.Log("fatal", err)
+	}
+
 	// apigateway
 	go func() {
 		r := mux.NewRouter()
-		r.HandleFunc("/api/{service}/{method}", apiGateway)
+		r.HandleFunc("/api/addsvc/sum", makeSumHandler(ctx, loadbalancer.NewRoundRobin(addsvcSum)))
+		r.HandleFunc("/api/addsvc/concat", makeConcatHandler(ctx, loadbalancer.NewRoundRobin(addsvcConcat)))
+		r.HandleFunc("/api/stringsvc/uppercase", factoryPassHandler(loadbalancer.NewRoundRobin(uppercase), logger))
+		r.HandleFunc("/api/stringsvc/count", factoryPassHandler(loadbalancer.NewRoundRobin(count), logger))
 		errc <- http.ListenAndServe(*httpAddr, r)
 	}()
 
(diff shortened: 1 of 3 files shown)


marshauf:
@peterbourgon Could you take a look at the last commit and tell me if that is what you have in mind, please.

peterbourgon:
Keeping my comments limited to the architecture for the time being...

- Static declaration of proxied services is what I had in mind -- great

- What is gained by the separate foosvc_def.json files, versus just hard-coding proxied services into the gateway?

- Having to duplicate per-endpoint code in make{Concat,Sum}{Endpoint,Handler} feels like a red flag. This code shouldn't need to exist in the gateway. Could it be handled by per-service client packages (e.g. addsvc/client)? What are your thoughts?

marshauf:
The json files are consul definition files. The example services don't [self-register](http://microservices.io/patterns/self-registration.html) with consul. I pass these files to consul on startup to let it know about the services which is I don't like but was the fastest way. I could add self-registration to the other example services or add a script which starts the service and does [3rd party registration](http://microservices.io/patterns/3rd-party-registration.html).

Yes, I need to clean up the code. The make{Concat,Sum}Handler can be merged together. I lean towards having the endpoint and handler factory functions in the client/service packages, to keep the code each service adds, to the apigateway, to a minimum. It would make swapping service in and out of the apigateway simpler and reducing the imports to just one package per services instead for example addsvc, being two imports at the moment.

peterbourgon:
> The json files are consul definition files. 

Ahh, I see. For the purposes of this example, let's leave those out, and just assume that the operator is already running the services and has them registered in Consul. (We should add self-registration in #167.)

> Yes, I need to clean up the code.

It's not just about cleaning up the existing code, it's about domain of responsibility. If the API needs to unmarshal and remarshal requests, then it seems like that code should be owned by the respective services and called, rather than copied and used in the API gateway. This implies the services need either have client packages, or export the MakeFoo{Handler, Endpoint} methods — not sure which right now. Right? What do you think?

marshauf:
> For the purposes of this example, let's leave those out, and just assume that the operator is already running the services and has them registered in Consul.
I agree, I removed the files.

> What do you think?
Endpoint methods should be in a separate package, like netflix is handling it.
Reducing the amount of handlers to one per API. The handler takes the request body and passes it to the endpoint. The endpoint creates a request based on the data and calls the service endpoint. The result of the call is passed back to the handler and written to the client(ResponseWriter).

peterbourgon:
OK. Where do we stand here? Is the PR in a state that you consider final? Or are you looking for more direction from me?

marshauf:
I consider it final. It shows a basic apigateway, which could be used as a starting point. I like to keep the service endpoint factory code (eg. makeSumEndpoint) in the apigateway package because it is just relevant to the apigateway example. Should I squash the commits into one?

examples/apigateway/main.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/go-kit/kit/loadbalancer/consul"
+	log "github.com/go-kit/kit/log"
+	//grpctransport "github.com/go-kit/kit/transport/grpc"
+	httptransport "github.com/go-kit/kit/transport/http"
+	//proto "github.com/golang/protobuf/proto"
+	"github.com/gorilla/mux"
+	"github.com/hashicorp/consul/api"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
```
peterbourgon: Please separate these imports into 3 groups:

```go
import (
    // stdlib

    // non go-kit

    // github.com/go-kit/kit/*
)
```

and remove the commented-out statements.

examples/apigateway/main.go
```diff
@@ -0,0 +1,203 @@
...
+
+func main() {
+	fs := flag.NewFlagSet("", flag.ExitOnError)
```
peterbourgon: No reason to use a flagset, as we're not importing any package with a transitive dependency on glog. You can just use flag.String, flag.Parse, etc.

examples/apigateway/main.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// log
+	logger := log.NewLogfmtLogger(os.Stderr)
+	logger = log.NewContext(logger).With("ts", log.DefaultTimestampUTC).With("caller", log.DefaultCaller)
+	stdlog.SetFlags(0)                             // flags are handled by Go kit's logger
+	stdlog.SetOutput(log.NewStdlibAdapter(logger)) // redirect anything using stdlib log to us
+
+	// errors
+	errc := make(chan error)
```
peterbourgon: Please put this down with the other goroutine pushing to errc. That would help reveal a bug: this channel needs a buffer size of 2 :)

examples/apigateway/main.go
```diff
@@ -0,0 +1,203 @@
...
+			"/api/addsvc/sum":    factoryAddsvc(ctx, logger, makeSumEndpoint),
+		},
+	})
+	serviceDefs = append(serviceDefs, &ServiceDef{
+		Name: "stringsvc",
+		Endpoints: map[string]loadbalancer.Factory{
+			"/api/stringsvc/uppercase": routeFactory(ctx, "uppercase"),
+			"/api/stringsvc/count":     routeFactory(ctx, "count"),
+		},
+	})
```
peterbourgon: There's a lot of indirection in this block of code. An intermediate type `ServiceDef`, a factory function that takes a factory function and is only used immediately below. I'm quite confused as to what is actually happening here...

examples/apigateway/main.go
```diff
@@ -0,0 +1,203 @@
...
+	r := mux.NewRouter()
+	for _, def := range serviceDefs {
+		for path, e := range def.Endpoints {
+			pub, err := consul.NewPublisher(discoveryClient, e, logger, def.Name)
+			if err != nil {
+				logger.Log("fatal", err)
+			}
+			r.HandleFunc(path, makeHandler(ctx, loadbalancer.NewRoundRobin(pub), logger))
+		}
+	}
```
peterbourgon: I think there's a way to do lines 69–96 much more declaratively, without the `ServiceDef` type, and without so many factory functions...

peterbourgon: OK, I have a patch incoming...

examples/apigateway/main.go
```diff
@@ -0,0 +1,203 @@
...
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+	return fmt.Errorf("%s", <-c)
+}
+
+func makeHandler(ctx context.Context, lb loadbalancer.LoadBalancer, logger log.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		e, err := lb.Endpoint()
+		if err != nil {
+			logger.Log("error", err)
+			return
```
peterbourgon: I think you have to write something to the ResponseWriter with `http.Error` or something..

peterbourgon:
Here's a patch: https://gist.github.com/peterbourgon/4385034065703985e2ef

You can apply it with `git apply /path/to/apigateway.patch`

Changes:

- Reorder imports
- Remove flagset
- Add loadbalancer.Retry and parameters
- Restructure service definitions
  - Make ServiceDef an anonymous inline type
  - Rename the factory functions for symmetry
  - Add a lot of comments
  - Pass method as parameter to httpFactory
- Passing "fatal" to logger.Log doesn't terminate; fix some instances of that
- Consistent logger.Log "err" key with error vals
- Changes to makeHandler
  - Use Retry-wrapped loadbalancer instead of invoking loadbalancer.Endpoint directly
  - Write http.Error to client in case of error

Does it make sense? Did I overlook anything? Do you like it? :)

marshauf:
The patch looks great. 

>  Please put this down with the other goroutine pushing to errc. That would help reveal a bug: this channel needs a buffer size of 2 :)
I don't see a problem with it being an unbuffered channel. You also didn't change it in the patch.
As soon as an error value is received through errc the main function returns and exits the program.
Which should only happen on an interrupt signal or the http server returning an error. Am I missing something?

peterbourgon:
@marshauf You're right. In some circumstances it's important for all goroutines to be able to push into the channel even if there's no receivers, to prevent goroutine leaks. But in this case we just teardown immediately, so no problem.

peterbourgon:
Nice! Thanks for the contribution!

PR #204, opened 2016-03-07
Add shipping example

marcusolsson:
This is an example based on the [goddd](https://github.com/marcusolsson/goddd) application to demonstrate a bit more ambitious example or Go kit.


marcusolsson:
Realized that there's actually a couple of things I want to fix before merging this. I'll let you know when they have been sorted out.

PR #207, opened 2016-03-09
[WIP] Provide rationale on log README

ifraixedes:
I wanted to make this proposal for rational section in log README.

Please review and provide feedback

### DIFF
--- a/log/README.md
+++ b/log/README.md
@@ -6,7 +6,11 @@ It can be used for both typical application log events, and log-structured data 
 
 ## Rationale
 
-TODO
+Logging is an important piece in almost any application, but it's a fundamental piece in distributed applications.
+
+Logs allow to track the system activity allowing to know the operations, errors, unlikely results, etc., which have happened at a specific time, with the possibility to be analyzed in the future for identifying system errors, security issues and so on.
+
+Structure logging allows to register discrete fields in addition to information messages, facilitating their process and storage, aiming to ease their post analysis.
 
 ## Usage
 
@@ -65,7 +69,7 @@ import (
 func main() {
 	logger := kitlog.NewJSONLogger(os.Stdout)
 	stdlog.SetOutput(kitlog.NewStdlibAdapter(logger))
-	
+
 	stdlog.Print("I sure like pie")
 }
 ```


log/README.md
```diff
@@ -6,7 +6,9 @@ It can be used for both typical application log events, and log-structured data
 
 ## Rationale
 
-TODO
+Logging is an important piece in al most any application, but it's a fundamental pice in distributed applications. Logs allow track the system activity allowing to know operations, errors, unlikely results, etc. happened at specific time, with the possibility to be analyzed in the future for identifying system errors, security issues and so on.
```
nelz9999: "pice" -> "piece"?

ChrisHines:
Thanks for providing a start on the rationale. The proposed text has many formatting, spelling, and grammatical errors. Also, the last sentence runs on a bit and should be broken up.

ifraixedes:
Thanks for the feedback and my apologies for the caused annoyance.

I was focused to provide a content which describes it than on the grammar side, moreover I wasn't in the best place to write it; anyway, I may not come up with a more perfect English as it isn't my native language and sometimes I'm struggling with it.

@ChrisHines I've tried to fix the spelling and grammatical errors, however, I'm not able to identify the formatting and the sentence broken up.

I pushed another version, hopefully it is better, although not perfect.

log/README.md
```diff
@@ -6,7 +6,11 @@ It can be used for both typical application log events, and log-structured data
 
 ## Rationale
 
-TODO
+Logging is an important piece in almost any application, but it's a fundamental piece in distributed applications.
+
+Logs allow to track the system activity allowing to know the operations, errors, unlikely results, etc., which have happened at a specific time, with the possibility to be analyzed in the future for identifying system errors, security issues and so on.
+
+Structure logging allows to register discrete fields in addition to information messages, facilitating their process and storage, aiming to ease their post analysis.
```
peterbourgon: The first two sections are a bit too generic. We don't need to justify the practice of logging generally, as I hope everyone understands what it is and why it's good! :) Rather we should justify specifically what logging can and should be used for (see [this article](http://peter.bourgon.org/blog/2016/02/07/logging-v-instrumentation.html) and [this article](https://blog.raintank.io/logs-and-metrics-and-graphs-oh-my/) for my thinking here) and why consumers should choose kit/log over other log packages. Your third touches on that: structure_d_ logging is a bit part of it, but I think not the only reason.

ifraixedes: Great!
I understand better what is needed; I wrote it based in another rationale (Tracing), but I didn't get the concept as you have specified.

Thanks

ifraixedes:
I don't think that I could provide something as expected in the following days, so feel free to close this if it's appropriated.

Thanks for your feedback!

PR #208, opened 2016-03-09
Fix dead link in tracing readme

jupp0r:
This change replaces a dead link for the initial appdash blog post with the correct url.


peterbourgon:
That link is dead too :( but we got another PR with a different link to the GitHub repo :) So I think we'll stick with that. But thank you very much for the contribution!

PR #211, opened 2016-03-11
kit/log failing tests on tip?

peterbourgon:
Apparently package log is panicking when tested on Travis against tip? This PR is to test that hypothesis...


peterbourgon:
https://travis-ci.org/go-kit/kit/jobs/115414128 ctrl-F "build failed"

@ChrisHines any guess?

ChrisHines:
There are two things going on here.

1. The go linker is panicing on tip. Not much we can do about that.
1. The tests for package `github.com/go-stack/stack` fail on TravisCI for everything except tip because TravisCI runs Go from a path that doesn't match the path it was built on.

In my opinion TravisCI is using a non-standard Go install for all but tip.

The TravisCI setup makes GOROOT differ from the paths compiled into the standard library and it breaks one of the tests in the github.com/go-stack/stack package. I discovered this quite a while ago when I tried to use TravisCI on that repo, which is why the test bothers to print the current GOROOT when it fails. See:

--- FAIL: TestTrimRuntime-2 (0.00s)
	stack_test.go:219: got len(trace) == 3, want 1, goroot: "/home/travis/.gimme/versions/go1.4.1.linux.amd64", trace: [/home/travis/gopath/src/github.com/go-stack/stack/stack_test.go:217 /usr/local/go/src/testing/testing.go:447 /usr/local/go/src/runtime/asm_amd64.s:2232]

Note how GOROOT=/home/travis/.gimme/versions/go1.4.1.linux.amd64, but the path to the standard library is /usr/local/go/src.

As a side note, why are the tests for github.com/go-stack/stack being run here anyway?

ChrisHines:
Wait. I was confused by the TravisCI UI. Let me dig into this more.  Deleteing my previous comment because it is irrelevant.

ChrisHines:
I would expect to see some compiler errors after a "build failed" from `go test`. Without that it is hard to say what is going on. Considering the panic from the linker earlier in the log I would say that Go tip is currently broken.

peterbourgon:
> I would say that Go tip is currently broken.

Yeah, I will just remove tip from the Travis target list for now. Thanks for taking a look!

dgryski:
Probably worth isolating the failure and submitting an issue for it if possible.

ChrisHines:
I'll give it a shot.

peterbourgon:
https://github.com/golang/go/issues/14783

edit: @ChrisHines if you file something better just let me know and I'll close my issue.

ChrisHines:
When I run `go test -v github.com/go-kit/kit/log` on tip I get:

```
C:\go\src>\go\bin\go test -v github.com/go-kit/kit/log
# testmain
C:\Users\Chris\AppData\Local\Temp\go-build245128694\github.com\go-kit\kit\log\_test\_testmain.go:96: too many values in struct initializer
FAIL    github.com/go-kit/kit/log [build failed]
```

This seems to have been introduced by https://github.com/golang/go/commit/9323de3da79d240b8223f74bd8ddd0d55dab8070. I will file an issue on the Go project as soon as I get a chance.

peterbourgon:
I'll merge this PR to get our builds green.

ChrisHines:
Testing this problem *correctly* is not easy. I was getting the previous error because my test environment was using the go tool from tip but building against the testing package from go 1.6. After fixing that I was able to reproduce the same error @peterbourgon reported already.

```
C:\GoTip\go\src>\GoTip\go\bin\go test -v github.com/go-kit/kit/log
# testmain
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0xf0 pc=0x4c453e]

goroutine 1 [running]:
panic(0x5b9340, 0xc082002100)
        C:/GoTip/go/src/runtime/panic.go:500 +0x197
cmd/link/internal/ld.decodetype_funcincount(0x0, 0x28)
        C:/GoTip/go/src/cmd/link/internal/ld/decodesym.go:182 +0x2e
cmd/link/internal/ld.decode_methodsig(0xc0830d50b0, 0x40070, 0x30, 0x1, 0x1, 0x3, 0x0)
        C:/GoTip/go/src/cmd/link/internal/ld/decodesym.go:281 +0x128
cmd/link/internal/ld.decodetype_methods(0xc0830d50b0, 0x0, 0x0, 0x0)
        C:/GoTip/go/src/cmd/link/internal/ld/decodesym.go:355 +0x170
cmd/link/internal/ld.(*deadcodepass).flood(0xc082029cd0)
        C:/GoTip/go/src/cmd/link/internal/ld/deadcode.go:296 +0xd57
cmd/link/internal/ld.deadcode(0xc082574000)
        C:/GoTip/go/src/cmd/link/internal/ld/deadcode.go:56 +0xcc
cmd/link/internal/ld.Ldmain()
        C:/GoTip/go/src/cmd/link/internal/ld/pobj.go:190 +0x1350
cmd/link/internal/amd64.Main()
        C:/GoTip/go/src/cmd/link/internal/amd64/obj.go:44 +0x20
main.main()
        C:/GoTip/go/src/cmd/link/main.go:27 +0x283
FAIL    github.com/go-kit/kit/log [build failed]
```

PR #212, opened 2016-03-11
TIP test... zookeeper test problems?

basvanbeek:
Somehow I think the test before kit/log is the issue...


basvanbeek:
Nope... it was not the test before....

PR #218, opened 2016-03-15
Fixed grpc.Client to be go-routine safe. 

basvanbeek:
This is an API breaking change as NewClient now wants a grpc Reply struct value instead of pointer

### DIFF
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -2,6 +2,7 @@ package grpc
 
 import (
 	"fmt"
+	"reflect"
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


transport/grpc/client.go
```diff
@@ -33,12 +34,16 @@ func NewClient(
...
-		before:    []RequestFunc{},
+		client: cc,
+		method: fmt.Sprintf("/pb.%s/%s", serviceName, method),
+		enc:    enc,
+		dec:    dec,
+		grpcReply: reflect.TypeOf(
+			reflect.Indirect(
+				reflect.ValueOf(grpcReply),
+			).Interface(),
+		),
```
jmank88: Can you explain why `reflect.TypeOf(grpcReply)` would be insufficient here?

basvanbeek: As I wanted to allow both specificReply{} and &specificReply{}

peterbourgon: Can you add that as a comment to this line, so we have context in the future?

peterbourgon:
Apart from that request for a comment, LGTM!

peterbourgon:
Thanks very much for the contribution!

Issue #217, opened 2016-03-14
examples/addsvc: use transport/grpc


peterbourgon:
Right now, the example just does a dumb binding to gRPC directly. Change it to use transport/grpc.

jupp0r:
I think this can be closed.

peterbourgon:
Righto.

PR #219, opened 2016-03-17
Use grpc transport in addsvc example client

jupp0r:
This change replaces manual usage of grpc in the addsvc example with the client from transport/grpc #217

### DIFF
--- a/examples/addsvc/client/grpc/client.go
+++ b/examples/addsvc/client/grpc/client.go
@@ -4,20 +4,28 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 
+	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/pb"
 	"github.com/go-kit/kit/examples/addsvc/server"
 	"github.com/go-kit/kit/log"
+	grpctransport "github.com/go-kit/kit/transport/grpc"
 )
 
 // New returns an AddService that's backed by the provided ClientConn.
 func New(ctx context.Context, cc *grpc.ClientConn, logger log.Logger) server.AddService {
-	return client{ctx, pb.NewAddClient(cc), logger}
+	return client{
+		Context: ctx,
+		Logger:  logger,
+		sum:     grpctransport.NewClient(cc, "Add", "sum", encodeSumRequest, decodeSumResponse, pb.SumReply{}).Endpoint(),
+		concat:  grpctransport.NewClient(cc, "Add", "concat", encodeConcatRequest, decodeConcatResponse, pb.ConcatReply{}).Endpoint(),
+	}
 }
 
 type client struct {
 	context.Context
-	pb.AddClient
 	log.Logger
+	sum    endpoint.Endpoint
+	concat endpoint.Endpoint
 }
 
 // TODO(pb): If your service interface methods don't return an error, we have
@@ -36,27 +44,31 @@ type client struct {
 //   important.
 
 func (c client) Sum(a, b int) int {
-	request := &pb.SumRequest{
-		A: int64(a),
-		B: int64(b),
+	request := &server.SumRequest{
+		A: a,
+		B: b,
 	}
-	reply, err := c.AddClient.Sum(c.Context, request)
+	reply, err := c.sum(c.Context, request)
 	if err != nil {
 		c.Logger.Log("err", err) // Without an error return parameter, we can't do anything else...
 		return 0
 	}
-	return int(reply.V)
+
+	r := reply.(server.SumResponse)
+	return r.V
 }
 
 func (c client) Concat(a, b string) string {
-	request := &pb.ConcatRequest{
+	request := &server.ConcatRequest{
 		A: a,
 		B: b,
 	}
-	reply, err := c.AddClient.Concat(c.Context, request)
+	reply, err := c.concat(c.Context, request)
 	if err != nil {
-		c.Logger.Log("err", err)
+		c.Logger.Log("err", err) // Without an error return parameter, we can't do anything else...
 		return ""
 	}
-	return reply.V
+
+	r := reply.(server.ConcatResponse)
+	return r.V
 }
--- a/examples/addsvc/client/grpc/encode_decode.go
+++ b/examples/addsvc/client/grpc/encode_decode.go
@@ -1,0 +1,45 @@
+package grpc
+
+import (
+	"github.com/go-kit/kit/examples/addsvc/pb"
+	"github.com/go-kit/kit/examples/addsvc/server"
+	"golang.org/x/net/context"
+)
+
+func encodeSumRequest(ctx context.Context, req interface{}) (interface{}, error) {
+	sumRequest := req.(server.SumRequest)
+
+	pbRequest := &pb.SumRequest{
+		A: int64(sumRequest.A),
+		B: int64(sumRequest.B),
+	}
+	return pbRequest, nil
+}
+
+func encodeConcatRequest(ctx context.Context, req interface{}) (interface{}, error) {
+	concatRequest := req.(server.ConcatRequest)
+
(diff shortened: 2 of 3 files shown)


examples/addsvc/client/grpc/client.go
```diff
@@ -4,20 +4,32 @@ import (
...
+	return client{ctx, sumClient.Endpoint(), concatClient.Endpoint(), logger, replies}
 }
 
 type client struct {
 	context.Context
-	pb.AddClient
+	sumEndpoint    endpoint.Endpoint
+	concatEndpoint endpoint.Endpoint
 	log.Logger
+	replies
```
peterbourgon: Why is this included? It's never referenced.

examples/addsvc/client/grpc/client.go
```diff
@@ -4,20 +4,32 @@ import (
...
 	"github.com/go-kit/kit/log"
+	grpctransport "github.com/go-kit/kit/transport/grpc"
 )
 
 // New returns an AddService that's backed by the provided ClientConn.
 func New(ctx context.Context, cc *grpc.ClientConn, logger log.Logger) server.AddService {
-	return client{ctx, pb.NewAddClient(cc), logger}
+	var replies replies
+	sumClient := grpctransport.NewClient(cc, "Add", "sum", encodeSumRequest, decodeSumResponse, replies.sumReply)
+	concatClient := grpctransport.NewClient(cc, "Add", "concat", encodeConcatRequest, decodeConcatResponse, replies.concatReply)
```
peterbourgon: Unless I'm missing something, you can pass concrete types directly to the NewClients, and remove the `replies` struct...

examples/addsvc/client/grpc/client.go
```diff
@@ -36,27 +48,41 @@ type client struct {
...
-	reply, err := c.AddClient.Sum(c.Context, request)
+	reply, err := c.sumEndpoint(c.Context, request)
 	if err != nil {
 		c.Logger.Log("err", err) // Without an error return parameter, we can't do anything else...
 		return 0
 	}
-	return int(reply.V)
+
+	r, ok := reply.(server.SumResponse)
+	if !ok {
```
peterbourgon: I think if this is not true, it's a programmer error, and it's better to just panic. So this (and all other type assertions) can elide the `, ok` check, IMO.

examples/addsvc/client/grpc/encode_decode.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/go-kit/kit/examples/addsvc/server"
+	"golang.org/x/net/context"
+)
+
+func encodeSumRequest(ctx context.Context, req interface{}) (interface{}, error) {
+	sumRequest, ok := req.(server.SumRequest)
+
+	if !ok {
+		return nil, errors.New("sum request broken")
+	}
```
peterbourgon: Likewise above, here.

peterbourgon:
Thank you! Apart from those few things, LGTM.

examples/addsvc/client/grpc/encode_decode.go
```diff
@@ -0,0 +1,45 @@
...
+	concatRequest := req.(server.ConcatRequest)
+
+	pbRequest := &pb.ConcatRequest{
+		A: concatRequest.A,
+		B: concatRequest.B,
+	}
+	return pbRequest, nil
+}
+
+func decodeSumResponse(ctx context.Context, req interface{}) (interface{}, error) {
```
jmank88: Should `req` be renamed to `resp` or `reply`?
Same for decodeConcatResponse.

jupp0r: Definitely, copy + paste error

examples/addsvc/client/grpc/client.go
```diff
@@ -4,19 +4,25 @@ import (
...
 	"github.com/go-kit/kit/log"
+	grpctransport "github.com/go-kit/kit/transport/grpc"
 )
 
 // New returns an AddService that's backed by the provided ClientConn.
 func New(ctx context.Context, cc *grpc.ClientConn, logger log.Logger) server.AddService {
-	return client{ctx, pb.NewAddClient(cc), logger}
+	var sumReply, concatReply string
+	sumClient := grpctransport.NewClient(cc, "Add", "sum", encodeSumRequest, decodeSumResponse, sumReply)
+	concatClient := grpctransport.NewClient(cc, "Add", "concat", encodeConcatRequest, decodeConcatResponse, concatReply)
```
basvanbeek: sumReply and concatReply should not be strings. You should use `pb.SumReply{}` and `pb.SumRequest{}` from the generated protobuf package.

examples/addsvc/client/grpc/client.go
```diff
@@ -4,19 +4,25 @@ import (
...
+
+	return client{ctx, sumClient.Endpoint(), concatClient.Endpoint(), logger}
 }
 
 type client struct {
 	context.Context
-	pb.AddClient
+	sumEndpoint    endpoint.Endpoint
+	concatEndpoint endpoint.Endpoint
```
basvanbeek: i would personally use `sum` and `concat` just like the other transports show.

examples/addsvc/client/grpc/client.go
```diff
@@ -4,19 +4,25 @@ import (
...
 )
 
 // New returns an AddService that's backed by the provided ClientConn.
 func New(ctx context.Context, cc *grpc.ClientConn, logger log.Logger) server.AddService {
-	return client{ctx, pb.NewAddClient(cc), logger}
+	var sumReply, concatReply string
+	sumClient := grpctransport.NewClient(cc, "Add", "sum", encodeSumRequest, decodeSumResponse, sumReply)
+	concatClient := grpctransport.NewClient(cc, "Add", "concat", encodeConcatRequest, decodeConcatResponse, concatReply)
+
+	return client{ctx, sumClient.Endpoint(), concatClient.Endpoint(), logger}
```
basvanbeek: I would follow the same coding style like httpjson/client.go and do something like this:
```
return client{
    Context: ctx,
    Logger:  logger,
    sum: grpctransport.NewClient(
        ...,
    ).Endpoint(),
    concat: grpctransport.NewClient(
        ...,
    ).Endpoint(),
}

File: examples/addsvc/client/grpc/client.go (hunk 6 above)
Comment (jupp0r) [inline examples/addsvc/client/grpc/client.go]: I took the usage of strings for the ```Invoke()``` from grpcs unit tests. Digging through the source, it's actually undefined behavior[1]:

> If the struct underlying pb does not match the data in buf, the results can be unpredictable.

[1] https://godoc.org/github.com/golang/protobuf/proto#Unmarshal

**Comment by jupp0r** [comment]
Build is broken because etcd upstream removed support for go 1.4: https://github.com/coreos/etcd/commit/25e47db4163505461ee51242d31342dcfe063405#diff-9c306d74ea02bd4f8619aaa482f8f341

**Comment by peterbourgon** [comment]
Too bad. Can you remove 1.4 from the travis.yml?

**Comment by jupp0r** [comment]
I opened #220 for removing 1.4 builds.

**Comment by jupp0r** [comment]
I addressed all comments so far and squashed the commit.

File: examples/addsvc/client/grpc/encode_decode.go [hunk 9]
```diff
@@ -0,0 +1,45 @@
+package grpc
+
+import (
+	"github.com/go-kit/kit/examples/addsvc/pb"
+	"github.com/go-kit/kit/examples/addsvc/server"
+	"golang.org/x/net/context"
+)
```
peterbourgon:
```go
import (
 	"golang.org/x/net/context"
 
	"github.com/go-kit/kit/examples/addsvc/pb"
 	"github.com/go-kit/kit/examples/addsvc/server"
 )
```

peterbourgon:
Yay! Thank you for the contribution!

PR #225, opened 2016-03-25
Export zipkin.FromContext for annotating spans inside service implementations

basvanbeek:
By passing context to a service implementation's methods we can pull the existing Zipkin span from context and Annotate items inside our business logic if we export zipkin.FromContext. This will typically be used to time the duration of calls to external resources like databases, caches, etc.

### DIFF
--- a/tracing/zipkin/zipkin_test.go
+++ b/tracing/zipkin/zipkin_test.go
@@ -61,6 +61,44 @@ func TestToContext(t *testing.T) {
 	}
 }
 
+func TestFromContext(t *testing.T) {
+	const (
+		hostport           = "5.5.5.5:5555"
+		serviceName        = "foo-service"
+		methodName         = "foo-method"
+		traceID      int64 = 14
+		spanID       int64 = 36
+		parentSpanID int64 = 58
+	)
+
+	ctx := context.WithValue(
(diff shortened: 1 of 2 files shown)


peterbourgon:
A+! Now that it's an exported method, would you mind adding a quick test?

basvanbeek:
No problem. Will have one in an hour or so.

basvanbeek:
added test and rebased

tracing/zipkin/zipkin_test.go
```diff
@@ -61,6 +61,44 @@ func TestToContext(t *testing.T) {
...
+	if span == nil {
+		t.Fatal("expected a Zipkin span object")
+	}
+	for want, haveFunc := range map[int64]func() int64{
+		traceID:      span.TraceID,
+		spanID:       span.SpanID,
+		parentSpanID: span.ParentSpanID,
+	} {
+		if have := haveFunc(); want != have {
+			name := runtime.FuncForPC(reflect.ValueOf(haveFunc).Pointer()).Name()
```
peterbourgon: Ha!

peterbourgon:
Brilliant. Thanks for the contribution!

Issue #197, opened 2016-02-22
Regarding use of gopkg.in


alimoeeny:
Hi there,
1- keep up the great work.
2- I was wondering if we can remove the use of `gopkg.in`. In `/log` (and maybe other places), a few packages are imported through `gopkg.in` like `gopkg.in/logfmt.v0` and `gopkg.in/stack.v1` . 
Although there is nothing wrong with this, it is yet another point of possible failure and possibility for attacks. 
I am basically suggesting we import them directly with their github urls like using `https://github.com/go-logfmt/logfmt` instead of `gopkg.in/logfmt.v0` 

If you are happy with this approach I'd be delighted to send a PR.

peterbourgon:
I am moderately to strongly in favor of this, but I know that @ChrisHines believes in the benefits of gopkg.in. Chris, would you be very much against using GitHub import paths directly?

ChrisHines:
I do see value in gopkg.in at least until the Go community has a better story for API versioning. I also see it as an implicit declaration of compatibility with the [Go kit API Stability Policy](https://github.com/go-kit/kit/blob/master/rfc/rfc007-api-stability.md).

As for those two specific packages:

gopkg.in/stack.v1 is very stable and importing from the master branch via github.com/go-stack/stack will work fine. Having said that, I am tracking [the discussion in Go issue 11432](https://github.com/golang/go/issues/11432) about a possible new API for call stack information in package runtime. That will more than likely just change the implementation rather than the API.

gopkg.in/logfmt.v0 is incomplete, meaning it hasn't met all of its goals yet. The portion used by Go kit is *likely* to be stable. I have just started working on this package again and hope to get it to a v1.0.0 soon (for some definition of soon). Importing it via github.com/go-logfmt/logfmt will also work fine and in fact looking at the repo now I see that the test code currently uses that path in the black-box tests which not the best practice when using gopkg.in, mea culpa.

ChrisHines:
@peterbourgon asked:

>Chris, would you be very much against using GitHub import paths directly?

I think it depends on how seriously we take the Go kit API stability policy. I think that in practice we won't have any problems using the github.com import paths, but it would be playing a bit loose with the rules we've established.

I'm happy to discuss and debate the topic to reach a consensus.

alimoeeny:
I honestly don't know what best fits the communities' needs and expectations, I of course personally prefer github imports. 
But I thought this sentence in the README  is relevant:
[...]

PR #226, opened 2016-03-25
Archive the RFCs

peterbourgon:
They bootstrapped the project but I think they've outlived their usefulness at this point.


peterbourgon:
In particular I think the API stability policy is in conflict with the current state of the Go package and vendoring ecosystem. I prefer to abstain from the discussion and push responsibility to `package main`. See #197.

Issue #231, opened 2016-04-01
Make the Errors returned by Client.Endpoint in transport/http.


Ayiga:
It would be incredibly beneficial, if an error returned from the Client.Endpoint call would return an error that would give you access to the original Error.

Something that would take the following code:

```go
if err = c.enc(req, request); err != nil {
	return nil, fmt.Errorf("Encode: %v", err)
}
```

and would return an Error with the original error reference intact, something like this:

```go
if err = c.enc(req, request); err != nil {
	return nil, Err{"Encode", err}
}
```

supposing that the structure for `Err` looking like:

```go
// Err represents an Error occurred in the Client transport level.
type Err struct {
	// Domain represents the domain of the error encountered.
	// Simply, this refers to the phase in which the error was
	// generated
	Domain string

	// Err references the underlying error that caused this error
	// overall.
	Err error
}

// Error implements the error interface
func (e Err) Error() string {
	return fmt.Sprintf("%s: %v", e.Domain, e.Err)
}
```

This would allow for the original Error to be referable, and would add a layer of indication to inform the overall scope of the error, and possibly why it was incurred.  This Error would also maintain the original output style that is already being used, if that is a requirement as well.

The primary use-case for doing this, is to make Exported errors easier to check without using string comparison.

For example, I could take the returned error, and see if the underlying issue was due to a specific cause that I might handle, as a special case.

```go
response, err := client.Endpoint()(ctx, request)
if err != nil {
	if e, ok := err.(httptransport.Err) {
		if e.Err == context.Canceled {
			// special handling
		}
	}
}
```

peterbourgon:
That merge closed the issue. Thanks!

PR #232, opened 2016-04-01
replace fmt.Errorf with Exported Err type

Ayiga:
This PR references issue https://github.com/go-kit/kit/issues/231

Create a new Err type that is able to represent the domain of the
error, as well as maintain a reference to the underlying error for
user reference.

Add test to ensure that the error formats in a similar way to the
previous errors being returned.

Add test to ensure that the new Err type is being returned from
Client.Endpoint when a failure occurs instead of an anonymous
errors.New error.

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"fmt"
 	"net/http"
 	"net/url"
 
@@ -75,11 +74,11 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		req, err := http.NewRequest(c.method, c.tgt.String(), nil)
 		if err != nil {
-			return nil, fmt.Errorf("NewRequest: %v", err)
+			return nil, Err{"NewRequest", err}
 		}
 
 		if err = c.enc(req, request); err != nil {
-			return nil, fmt.Errorf("Encode: %v", err)
+			return nil, Err{"Encode", err}
 		}
 
 		for _, f := range c.before {
@@ -88,7 +87,7 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		resp, err := ctxhttp.Do(ctx, c.client, req)
 		if err != nil {
-			return nil, fmt.Errorf("Do: %v", err)
+			return nil, Err{"Do", err}
 		}
 		if !c.bufferedStream {
 			defer resp.Body.Close()
@@ -96,7 +95,7 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		response, err := c.dec(resp)
 		if err != nil {
-			return nil, fmt.Errorf("Decode: %v", err)
+			return nil, Err{"Decode", err}
 		}
 
 		return response, nil
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -74,11 +74,11 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		req, err := http.NewRequest(c.method, c.tgt.String(), nil)
 		if err != nil {
-			return nil, Err{"NewRequest", err}
+			return nil, TransportError{DomainNewRequest, err}
 		}
 
 		if err = c.enc(req, request); err != nil {
-			return nil, Err{"Encode", err}
+			return nil, TransportError{DomainEncode, err}
 		}
 
 		for _, f := range c.before {
@@ -87,7 +87,7 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		resp, err := ctxhttp.Do(ctx, c.client, req)
 		if err != nil {
-			return nil, Err{"Do", err}
+			return nil, TransportError{DomainDo, err}
 		}
 		if !c.bufferedStream {
 			defer resp.Body.Close()
@@ -95,7 +95,7 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		response, err := c.dec(resp)
 		if err != nil {
-			return nil, Err{"Decode", err}
+			return nil, TransportError{DomainDecode, err}
 		}
 
 		return response, nil
--- a/transport/http/err.go
+++ b/transport/http/err.go
@@ -4,8 +4,28 @@ import (
 	"fmt"
 )
 
-// Err represents an Error occurred in the Client transport level.
-type Err struct {
+// These are some pre-generated constants that can be used to check against
+// for the DomainErrors.
+const (
+	// DomainNewRequest represents an error at the Request Generation
+	// Scope.
+	DomainNewRequest = "NewRequest"
+
+	// DomainEncode represent an error that has occurred at the Encode
+	// level of the request.
+	DomainEncode = "Encode"
+
+	// DomainDo represents an error that has occurred at the Do, or
+	// execution phase of the request.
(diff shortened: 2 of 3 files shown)


transport/http/err.go
```diff
@@ -0,0 +1,22 @@
+package http
+
+import (
+	"fmt"
+)
+
+// Err represents an Error occurred in the Client transport level.
+type Err struct {
```
peterbourgon: This should be of the form `SomethingError`, e.g. `TransportError`? See https://talks.golang.org/2014/names.slide#15

transport/http/err_test.go
```diff
@@ -0,0 +1,64 @@
...
+		"GET",
+		u,
+		enc,
+		dec,
+	)
+
+	_, err := c.Endpoint()(context.Background(), nil)
+	if err == nil {
+		t.Log("err == nil")
+		t.Fail()
```
peterbourgon: t.Log/t.Fail is better spelled t.Fatal

transport/http/err_test.go
```diff
@@ -0,0 +1,64 @@
...
+	e, ok := err.(transporthttp.Err)
+	if !ok {
+		t.Log("err is not of type github.com/go-kit/kit/transport/http.Err")
+		t.Fail()
+	}
+
+	if e.Err != sampleErr {
+		t.Logf("e.Err != sampleErr, %s vs %s", e.Err, sampleErr)
+		t.Fail()
+	}
```
peterbourgon:
```go
if want, have := sampleErr, e.Err; want != have {
    t.Fatalf("want %v, have %v", want, have)
}
```

transport/http/err_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"golang.org/x/net/context"
+
+	transporthttp "github.com/go-kit/kit/transport/http"
+
+	"testing"
```
peterbourgon: Please move it up with the other stdlib imports

transport/http/err_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func TestClientEndpointEncodeError(t *testing.T) {
+	var (
+		sampleErr = errors.New("Oh no, an error")
+		enc       = func(r *http.Request, request interface{}) error {
+			return sampleErr
+		}
+		dec = func(r *http.Response) (response interface{}, err error) {
+			return nil, nil
+		}
```
peterbourgon: Can these methods be defined on one line?

transport/http/client.go
```diff
@@ -75,11 +74,11 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		req, err := http.NewRequest(c.method, c.tgt.String(), nil)
 		if err != nil {
-			return nil, fmt.Errorf("NewRequest: %v", err)
+			return nil, Err{"NewRequest", err}
```
peterbourgon: Can the domains be exported constants in the package scope, so consumers may test for them?

transport/http/err_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"golang.org/x/net/context"
+
+	transporthttp "github.com/go-kit/kit/transport/http"
```
peterbourgon: Typically we name this `httptransport`

transport/http/err_test.go
Ayiga: I disagree, only slightly.  If it were t.Log/t.FailNow, then yes, it would be better as t.Fatal.  However, t.Fatal stops execution, whereas t.Fail allows it to continue.

Ayiga: I disagree, only slightly.  If it were t.Log/t.FailNow, then yes, it would be better as t.Fatal.  However, t.Fatal stops execution, whereas t.Fail allows it to continue.

peterbourgon: Ah, sorry, then you're looking for t.Error :)

Ayiga: Excellent

transport/http/err_test.go
Ayiga: Which line are you referring to in this one?

transport/http/client.go
Ayiga: Do you perfer a verbose constant, like:

```go
const (
	DomainRequestGeneration = "NewRequest"
)
```

Or would you prefer something smaller like:

```go
const (
	DomNewRequest = "NewRequest"
)
```

transport/http/err_test.go
peterbourgon: testing.

transport/http/client.go
peterbourgon: `DomainNewRequest` makes sense to me.

transport/http/err_test.go
Ayiga: I just looked at client_test.go, I'm assuming this is the testing package

Ayiga:
I apologize in general, for this, but I don't have any experience with Pull Requests, am I able to edit them, or do I have to submit a new one?

Furthermore, what's the best process of doing this, since I have this commit here on my master, should I create a new branch, or re-fork, then submit my new changes?

peterbourgon:
Ah no worries! Just push additional commits to your fork, they'll show up here automatically.

transport/http/err_test.go
peterbourgon: Sorry for the ambiguity. I just mean re-shape the import block like this:

```go
import (
	"errors"
	"fmt"
	"net/http"
	"net/url"
	"testing"

	"golang.org/x/net/context"

	transporthttp "github.com/go-kit/kit/transport/http" // actually httptransport per other comment
)
```

transport/http/err_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Error("err == nil")
+	}
+
+	e, ok := err.(httptransport.TransportError)
+	if !ok {
+		t.Error("err is not of type github.com/go-kit/kit/transport/http.Err")
+	}
+
+	if want, have := sampleErr, e.Err; want != have {
+		t.Error("want %v, have %v", want, have)
```
peterbourgon: Errorf here

peterbourgon:
With that minor nit fixed, LGTM! Thanks very much!

transport/http/err_test.go
Ayiga: Good catch, I'm also making the cast above Fatal, as the cast is assumed to of succeed if it moved past that condition.

transport/http/err_test.go
```diff
@@ -0,0 +1,56 @@
...
+	c := httptransport.NewClient(
+		"GET",
+		u,
+		enc,
+		dec,
+	)
+
+	_, err := c.Endpoint()(context.Background(), nil)
+	if err == nil {
+		t.Error("err == nil")
```
peterbourgon: I guess this should be t.Fatal, otherwise the next line will panic, right?

peterbourgon: Eh, I'll make this fix post-merge.

peterbourgon:
Thanks for the contribution! :)

PR #236, opened 2016-04-16
Zipkin Updates

basvanbeek:
Updated Zipkin Readme
Added CollectorToContext and NewChildSpan functions
addsvc update to use Kafka Collector instead of the deprecated Scribe collector

With NewChildSpan you can annotate calls to resources like databases and caches
as if they are regular services. This is done by calling NewChildSpan at call
time, annotating items and calling the CollectFunc returned from NewChildSpan
when done.

NewChildSpan will pull the parent Span from Context, create a child Span,
annotate ClientSend and return the child Span and a CollectFunc. When
CollectFunc is called the Span is annotated with ClientReceive and sent to the
Collector retrieved from Context.


basvanbeek:
Also added Sampled header support (both HTTP and gRPC) for informing downstream services to sample their spans for the trace

tracing/zipkin/zipkin.go
```diff
@@ -259,6 +312,12 @@ func FromContext(ctx context.Context) (*Span, bool) {
 	return span, true
 }
 
+// CollectorToContext stores a Zipkin collector to be used when annotating
+// resources using their own child spans with NewChildSpan.
+func CollectorToContext(ctx context.Context, c Collector) context.Context {
+	return context.WithValue(ctx, CollectorContextKey, c)
+}
```
peterbourgon: Hmm, this doesn't appear to be used, and I don't totally understand the comment. Can you explain more?

tracing/zipkin/zipkin_test.go
```diff
@@ -141,6 +152,9 @@ func TestToGRPCContext(t *testing.T) {
 			t.Errorf("%s: want %d, have %d", name, want, have)
 		}
 	}
+	if want, have := true, span.IsSampled(); want != have {
+		t.Errorf("IsSampled: want %d, have %d", want, have)
```
peterbourgon: `%v`

tracing/zipkin/zipkin_test.go
```diff
@@ -20,18 +20,20 @@ import (
...
-		traceID      int64 = 12
-		spanID       int64 = 34
-		parentSpanID int64 = 56
+		hostport            = "5.5.5.5:5555"
+		serviceName         = "foo-service"
+		methodName          = "foo-method"
+		traceID      int64  = 12
+		spanID       int64  = 34
+		parentSpanID int64  = 56
+		sampled      string = "1"
```
peterbourgon: extraneous `string`

tracing/zipkin/collector.go
```diff
@@ -6,18 +6,33 @@ import "strings"
...
+type NopCollector struct {
+	shouldSample Sampler
+}
 
 // Collect implements Collector.
 func (NopCollector) Collect(*Span) error { return nil }
 
+// ShouldSample implements Collector.
+func (n NopCollector) ShouldSample(span *Span) bool {
+	return span.debug || span.sampled || n.ShouldSample(span)
```
peterbourgon: The third clause looks like it will trigger an infinite loop?

tracing/zipkin/collector.go
```diff
@@ -6,18 +6,33 @@ import "strings"
...
 type Collector interface {
 	Collect(*Span) error
+	ShouldSample(*Span) bool
 	Close() error
 }
 
 // NopCollector implements Collector but performs no work.
-type NopCollector struct{}
+type NopCollector struct {
+	shouldSample Sampler
```
peterbourgon: Is it necessary? Can `NopCollector` simply always return `false`?

tracing/zipkin/zipkin.go
basvanbeek: The Zipkin Collector needs to be available inside service methods when creating a child span with NewChildSpan(). Best transport is context. By wiring your Zipkin collector in main to the root context using this function, you can now use NewChildSpan() inside service methods having a copy of context.

Example of NewChildSpan:

```go
// Foo example method of our Demo Service
func (demo *Demo) Foo(ctx context.Context, a, b string) (string, error) {
	// First we'll need to retrieve the span.
	span, collect := zipkin.NewChildSpan(
		ctx,
		"query", 
		zipkin.ServerAddr("my.mysql.host:3306", "MySQL"),
	)
	// Now we can annotate this resource
	span.Annotate("query start")
	span.AnnotateBinary("query", "select * from something where someotherthing = 'somevalue' order by bla limit 100")
	// run the query...

	span.Annotate("query end")
	span.Annotate("query start")
	/// do more and annotate...

	// we are done with this span... time to collect
	collect()
}
```

tracing/zipkin/collector.go
basvanbeek: yes... I first thought I needed it for testing but found out writing the tests I didn't. so it can be removed...

tracing/zipkin/zipkin.go
peterbourgon: So, until now, tracing was exclusively managed by endpoint middlewares. The collector was given as a dependency to the middleware during construction, and the span(s) were passed through the context because they needed to cross various boundaries and that's the only way to do it.

But if you need to do tracing stuff within your service methods, then by definition tracing is now part of your business logic, and the collector should be an explicit dependency of your service and fed as part of the constructor. Basically I am objecting to passing anything more than dumb data through the context object. It's a little bit of a gray area with the collector, but I want to hold a really strict line here, so we don't start seeing loggers or database connections passed thru the context too.

Does this make sense? Can NewChildSpan etc. be refactored?

yurishkuro: FWIW, that's not an idiomatic use of `net.Context`. From the [docs](https://godoc.org/golang.org/x/net/context):

> Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.

In [Go OpenTracing implementation](https://github.com/opentracing/opentracing-go/blob/master/span.go#L90) the Span exposes a reference to the Tracer

```go
	// Provides access to the Tracer that created this Span.
	Tracer() Tracer
```

peterbourgon: ....or, yeah, what @yurishkuro said ;)

basvanbeek:
If one needs to use NewChildSpan() within a service implementation, one will need to pass a Collector to the Service as its part of business logic...

tracing/README.md
```diff
@@ -18,93 +18,106 @@ it.
...
-vagrant init bas303/zipkin
-vagrant up --provider virtualbox
+If running on Linux `HOSTNAME` can be set to `localhost`. If running on Mac OS X
+or Windows you probably need to set the hostname environment variable to the
+hostname of the VM running the docker containers.
+
+```sh
+git clone https://github.com/basvanbeek/docker-zipkin.git
+cd docker-zipkin
+HOSTNAME=localhost docker-compose -f docker-compose.yml -f docker-compose-kafka.yml up
```
peterbourgon: Can you update this line please?

tracing/zipkin/span.go
```diff
@@ -197,14 +219,80 @@ func (s *Span) AnnotateString(key, value string) {
...
+func Debug(debug bool) SpanOption {
+	return func(s *Span) {
+		s.debug = debug
+	}
+}
+
+// CollectFunc will collect the span created with NewChildSpan.
+type CollectFunc func()
+
+// NewChildSpan returns a new child Span from a provided Parent Span. It can be
```
peterbourgon: From a parent span extracted from the passed context?

tracing/zipkin/span.go
```diff
@@ -197,14 +219,80 @@ func (s *Span) AnnotateString(key, value string) {
...
+// used to annotate resources like databases, caches, etc. and treat them as if
+// they are a regular service. For tracing client endpoints use AnnotateClient
+// instead.
```
peterbourgon: A note on the usage would be helpful. I guess something like

```go
func myHandler(ctx context.Context, ...) {
    // Basic usage
    _, collect := zipkin.NewChildSpan(ctx, ...)
    db.LongQuery()
    collect()

    // Advanced usage
    span, collect := zipkin.NewChildSpan(ctx, ...)
    span.AnnotateBinary("key", value)
    foo.Bar()
    collect()
}
```

peterbourgon:
With those 2 nits addressed, LGTM! Thanks so much for bearing with me on this. Zipkin is a beast.

basvanbeek:
as per our discussion on slack I have separated the zipkin specific part of the tracing readme and placed this in the zipkin directory.

Added an example of NewChildSpan usage and included an improved zipkin docker-compose file for people to run a development zipkin environment on Linux.

tracing/zipkin/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+If running on Linux `HOSTNAME` can be set to `localhost`. If running on Mac OS X
+or Windows you probably need to set the hostname environment variable to the
+hostname of the VM running the docker containers.
+
+```sh
+git clone https://github.com/basvanbeek/docker-zipkin.git
+cd docker-zipkin
+HOSTNAME=localhost docker-compose -f docker-compose-zipkin.yml up
+```
```
peterbourgon: This is now different, right?

```sh
cd tracing/zipkin/_docker
HOSTNAME=localhost docker-compose -f docker-compose-zipkin.yml up
```

basvanbeek:
Updated the Zipkin readme per peterbourgon-linter :)

peterbourgon:
Thanks so much for the effort!

Issue #239, opened 2016-04-19
etcd service discovery package has low test coverage


peterbourgon:
The client has 0% coverage, and the other component has no tests for the `<-responseChan` block in the main loop. It should be very straightforward to write two new tests to add this functionality.

Bigguy34:
So is this Issue fixed? Or can someone still work on it?

peterbourgon:
@AlexPikalov is working on it in #242. How's that going?

jstordeur:
tentative PR: [#344](https://github.com/go-kit/kit/pull/344)

jstordeur:
PR updated following code review and extra tests added for registrar

peterbourgon:
Fixed! 🎉

PR #241, opened 2016-04-23
Etcd client tests

AlexPikalov:
It contains tests for `loadbalancer/etcd` which are related to the issue https://github.com/go-kit/kit/issues/239.

To test successful branch of etcd Client test certificates were generated. Apart of this successful branch there a tests for following cases:
* no machines provided
* some of certificates is corrupted

There is an extra test for publisher's main loop


AlexPikalov:
hmm, something failed

PR #243, opened 2016-04-23
EndpointCache return sorted Endpoints

### LINKED ISSUE #159: EndpointCache returns sorted Endpoints
Hello,
I would be great that `loadbalancer.EndpointCache` will return the list of sorted Endpoints(alphanumeric).
My underlying usage is that I want to try as much as possible to redirect the same type of request (the type can be calculated with hash of specific attribute of request or something else) on the same endpoint with my own loadbalancer. 
By the way, I think with defined order will allow that the `loadbalancer.RoundRobin` to  be more consistency as it returns Endpoint sequentially.

If you think this feature can be accepted, I can create PR.

Many thx

jerome-laforge:
This closes #159 "EndpointCache returns sorted Endpoints" with no mutex and no memory allocation on hot path (i.e. Endpoints() ([]endpoint.Endpoint, error))
```
Benchmark new EndpointCache.Endpoints():
BenchmarkEndpoints-4	200000000	         7.27 ns/op	       0 B/op	       0 allocs/op

Benchmark old EndpointCache.Endpoints():
BenchmarkEndpoints-4	20000000	        98.3 ns/op	      16 B/op	       1 allocs/op
```

### DIFF
--- a/loadbalancer/endpoint_cache.go
+++ b/loadbalancer/endpoint_cache.go
@@ -2,7 +2,9 @@ package loadbalancer
 
 import (
 	"io"
+	"sort"
 	"sync"
+	"sync/atomic"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


peterbourgon:
Cool! Looks good. Out of curiosity, did you do this just for the fun of it, or were you hitting a performance problem?

jerome-laforge:
Hello,
No performance problem, my first goal was to return the Endpoint as sorted array.
The no mutex and no allocation memory is the icing on the cake.

PR #245, opened 2016-04-26
transport/http: change error and encoder definitions

peterbourgon:
- ErrorEndoer takes a context.Context as a first parameter
- Encoders and decoders also take context.Context as first parameter
- Refactor BadRequestError as a TransportError
- Rename TransportError to Error, to reduce stuttering
- Improve defaultErrorEncoder behavior

@thomshutt does this look good for you?

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -170,7 +170,7 @@ func main() {
 	log.Fatal(http.ListenAndServe(":8080", nil))
 }
 
-func decodeUppercaseRequest(r *http.Request) (interface{}, error) {
+func decodeUppercaseRequest(_ context.Context, r *http.Request) (interface{}, error) {
 	var request uppercaseRequest
 	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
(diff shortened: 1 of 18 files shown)


thomshutt:
New signatures look good to me, the only thing that threw what I currently had off is that you've started wrapping endpoint errors inside `TransportError` objects, which my errorEncoder implementation wasn't expecting

peterbourgon:
Cool. Yeah, that's a sort of standardization. Too much trouble to deal with?

thomshutt:
Nope, standardisation is good and no trouble to deal with now that I know what it's going to be. 

AFAICT it's _only_ now called with a `TransportError` though, so you could potentially change the ErrorEncoder signature to take one, rather than `error`. Either way, this does exactly what I need.

peterbourgon:
> AFAICT it's only now called with a `TransportError` though, so you could potentially change the ErrorEncoder signature to take one, rather than `error`

Hmm, excellent point!

Issue #37, opened 2015-04-29
Investigate Appdash support in package tracing


peterbourgon:
[Appdash, from SourceGraph](https://sourcegraph.com/blog/117580140734).

peterbourgon:
Work started by @anarcher in #102 — thanks!

peterbourgon:
This is now fully supported via OpenTracing, woohoo!

Issue #237, opened 2016-04-16
Support OpenTracing API


basvanbeek:
Next to our native implementation of Zipkin it could be very beneficial to support the OpenTracing API.

http://opentracing.io/

We can probably get it done quickly by using this package: https://github.com/opentracing/opentracing-go

peterbourgon:
Indeed Zipkin is becoming OpenTracing, on a long-enough timeline. Last I checked, a few months ago, the Go implementation was just a sketch, but it looks a little further along now. In any case, +1!

bensigelman:
@peterbourgon I just saw this... I'm happy to do the work if there's interest here.

peterbourgon:
@bensigelman Please! Let me know what you need.

bensigelman:
@peterbourgon if you have any initial guidance about the go-kit side and/or specific requirements, that would be welcome. If not, I will just do my best to make sure that existing code continues to work and that new code can use OpenTracing with go-kit.

peterbourgon:
Woohoo!

bensigelman:
@peterbourgon thanks for your help :)

PR #247, opened 2016-05-02
OpenTracing integration

bensigelman:
(See #237 for motivation and context)

# "Demo" (sic)
The `addsvc` in this PR uses the generic OpenTracing bindings for the httpjson and grpc transports, and in its `main()` it binds to either Appdash or LightStep (both of which are trivial to set up and have implementations of the Go OpenTracing API). As such, without changing any of the actual instrumentation beyond the `main()` function, both Appdash and LightStep are able to trace addsvc:

## Appdash

```
$ addsvc -appdash_hostport=localhost:7701 &
$ client -transport=grpc -appdash_hostport=localhost:7701 concat open tracing
```
image

## LightStep

```
$ addsvc -lightstep_access_token=XXX &
$ client -transport=grpc -lightstep_access_token=XXX concat open tracing
```
image

# Misc notes

The default implementation of OpenTracing is a noop. As such, rather than forcing users to install these various pieces of middleware/etc, some of this instrumentation could be opt-out, or perhaps downright unavoidable. Again, if the programmer didn't bother binding to OpenTracing, there would be near-zero overhead to support the noop implementation.

Also, the existing Zipkin instrumentation is worth thinking through in more detail. There are major companies that have adopted OpenTracing and bound it to Zipkin internally, so we know it's possible to go that route. If there's interest, we could try to port the `tracing/zipkin` package to be a pure OpenTracing implementation. I don't know how many depend on its API already, though.

# Next steps

I'd like to get basic feedback about the direction. If there's enthusiasm, I can finish up the other transports, add real tests, and merge this in. If there are problems, I'd rather go through them now before finishing up all of the odds and ends.

- - - - - - - - -

Thanks for reviewing!

---
This change is [Reviewable](https://reviewable.io/reviews/go-kit/kit/247)

### DIFF
--- a/examples/addsvc/client/grpc/factory.go
+++ b/examples/addsvc/client/grpc/factory.go
@@ -13,9 +13,10 @@ import (
 	"github.com/opentracing/opentracing-go"
 )
 
-// SumEndpointFactory transforms GRPC host:port strings into Endpoints that call the Sum method on a GRPC server
+// MakeSumEndpointFactory returns a loadbalancer.Factory that transforms GRPC
+// host:port strings into Endpoints that call the Sum method on a GRPC server
 // at that address.
-func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		cc, err := grpc.Dial(instance, grpc.WithInsecure())
 		return grpctransport.NewClient(
@@ -30,9 +31,10 @@ func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 	}
 }
 
-// ConcatEndpointFactory transforms GRPC host:port strings into Endpoints that call the Concat method on a GRPC server
-// at that address.
-func NewConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+// MakeConcatEndpointFactory returns a loadbalancer.Factory that transforms
+// GRPC host:port strings into Endpoints that call the Concat method on a GRPC
+// server at that address.
+func MakeConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		cc, err := grpc.Dial(instance, grpc.WithInsecure())
 		return grpctransport.NewClient(
--- a/examples/addsvc/client/grpc/factory.go
+++ b/examples/addsvc/client/grpc/factory.go
@@ -3,15 +3,15 @@ package grpc
 import (
 	"io"
 
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	"github.com/opentracing/opentracing-go"
 	"google.golang.org/grpc"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/pb"
 	"github.com/go-kit/kit/loadbalancer"
 	"github.com/go-kit/kit/log"
-	kitot "github.com/go-kit/kit/tracing/opentracing"
 	grpctransport "github.com/go-kit/kit/transport/grpc"
-	"github.com/opentracing/opentracing-go"
 )
 
 // MakeSumEndpointFactory returns a loadbalancer.Factory that transforms GRPC
--- a/examples/addsvc/client/grpc/factory.go
+++ b/examples/addsvc/client/grpc/factory.go
@@ -8,6 +8,7 @@ import (
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/pb"
 	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
 	kitot "github.com/go-kit/kit/tracing/opentracing"
 	grpctransport "github.com/go-kit/kit/transport/grpc"
 	"github.com/opentracing/opentracing-go"
@@ -16,7 +17,7 @@ import (
 // MakeSumEndpointFactory returns a loadbalancer.Factory that transforms GRPC
 // host:port strings into Endpoints that call the Sum method on a GRPC server
 // at that address.
-func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+func MakeSumEndpointFactory(tracer opentracing.Tracer, tracingLogger log.Logger) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		cc, err := grpc.Dial(instance, grpc.WithInsecure())
 		return grpctransport.NewClient(
@@ -26,7 +27,7 @@ func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 			encodeSumRequest,
 			decodeSumResponse,
 			pb.SumReply{},
-			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer)),
+			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer, tracingLogger)),
 		).Endpoint(), cc, err
 	}
 }
@@ -34,7 +35,7 @@ func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 // MakeConcatEndpointFactory returns a loadbalancer.Factory that transforms
 // GRPC host:port strings into Endpoints that call the Concat method on a GRPC
 // server at that address.
-func MakeConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+func MakeConcatEndpointFactory(tracer opentracing.Tracer, tracingLogger log.Logger) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		cc, err := grpc.Dial(instance, grpc.WithInsecure())
 		return grpctransport.NewClient(
@@ -44,7 +45,7 @@ func MakeConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 			encodeConcatRequest,
 			decodeConcatResponse,
 			pb.ConcatReply{},
-			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer)),
+			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer, tracingLogger)),
 		).Endpoint(), cc, err
 	}
 }
--- a/examples/addsvc/client/grpc/factory.go
+++ b/examples/addsvc/client/grpc/factory.go
@@ -7,33 +7,42 @@ import (
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/pb"
+	"github.com/go-kit/kit/loadbalancer"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
 	grpctransport "github.com/go-kit/kit/transport/grpc"
+	"github.com/opentracing/opentracing-go"
 )
 
 // SumEndpointFactory transforms GRPC host:port strings into Endpoints that call the Sum method on a GRPC server
 // at that address.
-func SumEndpointFactory(instance string) (endpoint.Endpoint, io.Closer, error) {
-	cc, err := grpc.Dial(instance, grpc.WithInsecure())
-	return grpctransport.NewClient(
-		cc,
-		"Add",
-		"Sum",
-		encodeSumRequest,
-		decodeSumResponse,
-		pb.SumReply{},
-	).Endpoint(), cc, err
+func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
+		cc, err := grpc.Dial(instance, grpc.WithInsecure())
+		return grpctransport.NewClient(
+			cc,
+			"Add",
+			"Sum",
+			encodeSumRequest,
+			decodeSumResponse,
+			pb.SumReply{},
+			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer)),
+		).Endpoint(), cc, err
+	}
 }
 
 // ConcatEndpointFactory transforms GRPC host:port strings into Endpoints that call the Concat method on a GRPC server
 // at that address.
-func ConcatEndpointFactory(instance string) (endpoint.Endpoint, io.Closer, error) {
-	cc, err := grpc.Dial(instance, grpc.WithInsecure())
-	return grpctransport.NewClient(
-		cc,
-		"Add",
-		"Concat",
-		encodeConcatRequest,
-		decodeConcatResponse,
-		pb.ConcatReply{},
-	).Endpoint(), cc, err
+func NewConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
+		cc, err := grpc.Dial(instance, grpc.WithInsecure())
+		return grpctransport.NewClient(
+			cc,
+			"Add",
+			"Concat",
+			encodeConcatRequest,
+			decodeConcatResponse,
+			pb.ConcatReply{},
+			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer)),
+		).Endpoint(), cc, err
+	}
 }
--- a/examples/addsvc/client/httpjson/factory.go
+++ b/examples/addsvc/client/httpjson/factory.go
@@ -12,11 +12,11 @@ import (
 	"github.com/opentracing/opentracing-go"
 )
 
-// SumEndpointFactory generates a Factory that transforms an http url into an
-// Endpoint.
+// MakeSumEndpointFactory generates a Factory that transforms an http url into
+// an Endpoint.
 //
 // The path of the url is reset to /sum.
-func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		sumURL, err := url.Parse(instance)
 		if err != nil {
@@ -37,11 +37,11 @@ func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 	}
 }
 
-// NewConcatEndpointFactory generates a Factory that transforms an http url
+// MakeConcatEndpointFactory generates a Factory that transforms an http url
 // into an Endpoint.
 //
 // The path of the url is reset to /concat.
-func NewConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+func MakeConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		concatURL, err := url.Parse(instance)
 		if err != nil {
--- a/examples/addsvc/client/httpjson/factory.go
+++ b/examples/addsvc/client/httpjson/factory.go
@@ -4,13 +4,14 @@ import (
 	"io"
 	"net/url"
 
+	"github.com/opentracing/opentracing-go"
+
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/server"
 	"github.com/go-kit/kit/loadbalancer"
 	"github.com/go-kit/kit/log"
 	kitot "github.com/go-kit/kit/tracing/opentracing"
 	httptransport "github.com/go-kit/kit/transport/http"
-	"github.com/opentracing/opentracing-go"
 )
 
 // MakeSumEndpointFactory generates a Factory that transforms an http url into
--- a/examples/addsvc/client/httpjson/factory.go
+++ b/examples/addsvc/client/httpjson/factory.go
@@ -6,45 +6,58 @@ import (
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/server"
+	"github.com/go-kit/kit/loadbalancer"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
 	httptransport "github.com/go-kit/kit/transport/http"
+	"github.com/opentracing/opentracing-go"
 )
 
-// SumEndpointFactory transforms a http url into an Endpoint.
+// SumEndpointFactory generates a Factory that transforms an http url into an
+// Endpoint.
+//
 // The path of the url is reset to /sum.
-func SumEndpointFactory(instance string) (endpoint.Endpoint, io.Closer, error) {
-	sumURL, err := url.Parse(instance)
-	if err != nil {
-		return nil, nil, err
+func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
+		sumURL, err := url.Parse(instance)
+		if err != nil {
+			return nil, nil, err
+		}
+		sumURL.Path = "/sum"
+
+		client := httptransport.NewClient(
+			"GET",
+			sumURL,
+			server.EncodeSumRequest,
+			server.DecodeSumResponse,
+			httptransport.SetClient(nil),
+			httptransport.SetClientBefore(kitot.ToHTTPRequest(tracer)),
+		)
+
+		return client.Endpoint(), nil, nil
 	}
-	sumURL.Path = "/sum"
-
-	client := httptransport.NewClient(
-		"GET",
-		sumURL,
-		server.EncodeSumRequest,
-		server.DecodeSumResponse,
-		httptransport.SetClient(nil),
-	)
-
-	return client.Endpoint(), nil, nil
 }
 
-// ConcatEndpointFactory transforms a http url into an Endpoint.
+// NewConcatEndpointFactory generates a Factory that transforms an http url
+// into an Endpoint.
+//
 // The path of the url is reset to /concat.
-func ConcatEndpointFactory(instance string) (endpoint.Endpoint, io.Closer, error) {
-	concatURL, err := url.Parse(instance)
-	if err != nil {
-		return nil, nil, err
+func NewConcatEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
+		concatURL, err := url.Parse(instance)
+		if err != nil {
+			return nil, nil, err
+		}
+		concatURL.Path = "/concat"
+
+		client := httptransport.NewClient(
+			"GET",
+			concatURL,
+			server.EncodeConcatRequest,
+			server.DecodeConcatResponse,
+			httptransport.SetClient(nil),
+			httptransport.SetClientBefore(kitot.ToHTTPRequest(tracer)),
+		)
+
+		return client.Endpoint(), nil, nil
 	}
-	concatURL.Path = "/concat"
-
-	client := httptransport.NewClient(
-		"GET",
-		concatURL,
-		server.EncodeConcatRequest,
-		server.DecodeConcatResponse,
-		httptransport.SetClient(nil),
-	)
-
-	return client.Endpoint(), nil, nil
 }
--- a/examples/addsvc/client/httpjson/factory.go
+++ b/examples/addsvc/client/httpjson/factory.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/server"
 	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
 	kitot "github.com/go-kit/kit/tracing/opentracing"
 	httptransport "github.com/go-kit/kit/transport/http"
 	"github.com/opentracing/opentracing-go"
@@ -16,7 +17,7 @@ import (
 // an Endpoint.
 //
 // The path of the url is reset to /sum.
-func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
+func MakeSumEndpointFactory(tracer opentracing.Tracer, tracingLogger log.Logger) loadbalancer.Factory {
 	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
 		sumURL, err := url.Parse(instance)
 		if err != nil {
@@ -30,7 +31,7 @@ func MakeSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
 			server.EncodeSumRequest,
 			server.DecodeSumResponse,
 			httptransport.SetClient(nil),
-			httptransport.SetClientBefore(kitot.ToHTTPRequest(tracer)),
+			httptransport.SetClientBefore(kitot.ToHTTPRequest(tracer, tracingLogger)),
 		)
 
 		return client.Endpoint(), nil, nil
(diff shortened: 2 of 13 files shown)


examples/addsvc/client/grpc/factory.go
```diff
@@ -7,33 +7,42 @@ import (
...
-	cc, err := grpc.Dial(instance, grpc.WithInsecure())
-	return grpctransport.NewClient(
-		cc,
-		"Add",
-		"Sum",
-		encodeSumRequest,
-		decodeSumResponse,
-		pb.SumReply{},
-	).Endpoint(), cc, err
+func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
```
bensigelman: btw, there is an `opentracing.GlobalTracer()` singleton; if we embrace it, the indirection here would not be necessary.

examples/addsvc/client/grpc/factory.go
```diff
@@ -7,33 +7,42 @@ import (
...
+	return func(instance string) (endpoint.Endpoint, io.Closer, error) {
+		cc, err := grpc.Dial(instance, grpc.WithInsecure())
+		return grpctransport.NewClient(
+			cc,
+			"Add",
+			"Sum",
+			encodeSumRequest,
+			decodeSumResponse,
+			pb.SumReply{},
+			grpctransport.SetClientBefore(kitot.ToGRPCRequest(tracer)),
```
bensigelman: the only "real" changes in this file are these before funcs... the rest is just indenting.

examples/addsvc/client/httpjson/factory.go
```diff
@@ -6,45 +6,58 @@ import (
...
+		}
+		sumURL.Path = "/sum"
+
+		client := httptransport.NewClient(
+			"GET",
+			sumURL,
+			server.EncodeSumRequest,
+			server.DecodeSumResponse,
+			httptransport.SetClient(nil),
+			httptransport.SetClientBefore(kitot.ToHTTPRequest(tracer)),
```
bensigelman: (same story here)

examples/addsvc/client/main.go
```diff
@@ -31,6 +36,10 @@ func main() {
 		thriftProtocol   = flag.String("thrift.protocol", "binary", "binary, compact, json, simplejson")
 		thriftBufferSize = flag.Int("thrift.buffer.size", 0, "0 for unbuffered")
 		thriftFramed     = flag.Bool("thrift.framed", false, "true to enable framing")
+
+		// Two OpenTracing backends (to demonstrate how they can be interchanged):
+		appdashHostport      = flag.String("appdash_hostport", "", "Enable Appdash tracing via an Appdash server host:port")
```
bensigelman: These next two lines are not intended for merge... they're just here to illustrate the generality of OpenTracing.

examples/addsvc/grpc_binding.go
```diff
@@ -6,17 +6,31 @@ import (
...
+func newGRPCBinding(ctx context.Context, tracer opentracing.Tracer, svc server.AddService) grpcBinding {
 	return grpcBinding{
-		sum:    grpc.NewServer(ctx, makeSumEndpoint(svc), servergrpc.DecodeSumRequest, servergrpc.EncodeSumResponse),
-		concat: grpc.NewServer(ctx, makeConcatEndpoint(svc), servergrpc.DecodeConcatRequest, servergrpc.EncodeConcatResponse),
+		sum: grpc.NewServer(
+			ctx,
+			kitot.TraceServer(tracer, "sum")(makeSumEndpoint(svc)),
+			servergrpc.DecodeSumRequest,
+			servergrpc.EncodeSumResponse,
+			grpc.ServerBefore(kitot.FromGRPCRequest(tracer, "")),
```
bensigelman: the only real change in this file... the rest is just line breaks.

examples/apigateway/main.go
```diff
@@ -77,8 +78,8 @@ func main() {
 		factory loadbalancer.Factory
 	}{
 		"addsvc": {
-			{path: "/api/addsvc/concat", factory: grpc.ConcatEndpointFactory},
-			{path: "/api/addsvc/sum", factory: grpc.SumEndpointFactory},
+			{path: "/api/addsvc/concat", factory: grpc.NewConcatEndpointFactory(opentracing.GlobalTracer())},
```
bensigelman: (just keeping the build intact for this PR)

examples/addsvc/main.go
```diff
@@ -123,35 +131,30 @@ func main() {
 	go func() {
 		var (
 			transportLogger = log.NewContext(logger).With("transport", "HTTP/JSON")
-			tracingLogger   = log.NewContext(transportLogger).With("component", "tracing")
-			newSumSpan      = zipkin.MakeNewSpanFunc(*zipkinHostPort, *zipkinServiceName, "sum")
```
bensigelman: note that OpenTracing doesn't need these sorts of `new*Span` functions... that's all part of the `Tracer` impl.

examples/addsvc/main.go
```diff
@@ -78,20 +84,22 @@ func main() {
 		))
 	}
 
-	// package tracing
-	var collector zipkin.Collector
+	// Set up OpenTracing
+	var tracer opentracing.Tracer
 	{
-		zipkinLogger := log.NewContext(logger).With("component", "zipkin")
```
bensigelman: I didn't thread a logger through to record errors... let me know if that's desired.

bensigelman:
Oh, this would also resolve https://github.com/go-kit/kit/issues/37 since Appdash supports OpenTracing (thanks to @bg451 / https://github.com/sourcegraph/appdash/pull/110).

tracing/opentracing/endpoint.go
```diff
@@ -0,0 +1,48 @@
...
+// TraceServer returns a Middleware that wraps the `next` Endpoint in an
+// OpenTracing Span called `operationName`.
+//
+// If `ctx` already has a Span, it is re-used and the operation name is
+// overwritten. If `ctx` does not yet have a Span, one is created here.
+func TraceServer(tracer opentracing.Tracer, operationName string) endpoint.Middleware {
+	return func(next endpoint.Endpoint) endpoint.Endpoint {
+		return func(ctx context.Context, request interface{}) (interface{}, error) {
+			serverSpan := opentracing.SpanFromContext(ctx)
+			if serverSpan == nil {
```
bg451: Is there any reason why `opentracing.StartSpanFromContext` isn't used instead?

examples/addsvc/client/grpc/factory.go
peterbourgon: I like the explicitness of the parameter much better.

examples/addsvc/client/httpjson/factory.go
```diff
@@ -6,45 +6,58 @@ import (
...
-// SumEndpointFactory transforms a http url into an Endpoint.
+// SumEndpointFactory generates a Factory that transforms an http url into an
+// Endpoint.
+//
 // The path of the url is reset to /sum.
-func SumEndpointFactory(instance string) (endpoint.Endpoint, io.Closer, error) {
-	sumURL, err := url.Parse(instance)
-	if err != nil {
-		return nil, nil, err
+func NewSumEndpointFactory(tracer opentracing.Tracer) loadbalancer.Factory {
```
peterbourgon: I guess it would be MakeSumEndpointFactory rather than New, and there's some discontinuity between the function name and the comment, but I'll leave this class of review off until the golden master, if that's OK.

examples/addsvc/client/main.go
```diff
@@ -31,6 +36,10 @@ func main() {
...
+		lightstepAccessToken = flag.String("lightstep_access_token", "", "Enable LightStep tracing via a LightStep access token")
```
peterbourgon: Got it. For the final version, please use `"appdash.addr"` and "lightstep.access.token"`.

examples/addsvc/client/main.go
```diff
@@ -49,6 +58,25 @@ func main() {
...
+				panic("Attempted to configure multiple OpenTracing implementations")
+			}
+			tracer = lightstep.NewTracer(lightstep.Options{
+				AccessToken: *lightstepAccessToken,
+			})
+		}
+		if tracer == nil {
+			tracer = opentracing.GlobalTracer() // the noop tracer
+		}
+	}
```
peterbourgon:
```go
switch {
case *appdashAddr != "" && *lightstepAccessToken == "":
    tracer = appdashot.NewTracer(...)
case *appdashAddr == "" && *lightstepAccessToken != "":
    tracer = lightstep.NewTracer(...)
case *appdashAddr == "" && *lightstepAccessToken == "":
    tracer = opentracing.GlobalTracer() // no-op
default:
    log.Fatal("specify either -appdash.addr or -lightstep.access.token, not both")
}
```

examples/addsvc/client/main.go
```diff
@@ -108,10 +136,15 @@ func main() {
 		logger.Log("err", "invalid method "+method)
 		os.Exit(1)
 	}
+
+	if len(*lightstepAccessToken) > 0 {
+		lightstep.FlushLightStepTracer(tracer)
+	}
```
peterbourgon: Can remove this if you add a `defer lightstep.FlushLightStepTracer(tracer)` after the creation in the switch block above.

examples/addsvc/client/main.go
```diff
@@ -108,10 +136,15 @@ func main() {
...
 }
 
-func buildEndpoint(instances []string, factory loadbalancer.Factory, seed int64, logger log.Logger) endpoint.Endpoint {
+func buildEndpoint(tracer opentracing.Tracer, operationName string, instances []string, factory loadbalancer.Factory, seed int64, logger log.Logger) endpoint.Endpoint {
 	publisher := static.NewPublisher(instances, factory, logger)
 	random := loadbalancer.NewRandom(publisher, seed)
-	return loadbalancer.Retry(10, 10*time.Second, random)
+	endpoint := loadbalancer.Retry(10, 10*time.Second, random)
+	return kitot.TraceClient(tracer, operationName)(endpoint)
```
peterbourgon: Should be TraceServer, right?

examples/addsvc/main.go
```diff
@@ -36,17 +41,18 @@ func main() {
...
+		grpcAddr         = fs.String("grpc.addr", ":8002", "Address for gRPC server")
+		netrpcAddr       = fs.String("netrpc.addr", ":8003", "Address for net/rpc server")
+		thriftAddr       = fs.String("thrift.addr", ":8004", "Address for Thrift server")
+		thriftProtocol   = fs.String("thrift.protocol", "binary", "binary, compact, json, simplejson")
+		thriftBufferSize = fs.Int("thrift.buffer.size", 0, "0 for unbuffered")
+		thriftFramed     = fs.Bool("thrift.framed", false, "true to enable framing")
+
+		// Supported OpenTracing backends
+		appdashHostport      = fs.String("appdash_hostport", "", "Enable Appdash tracing via an Appdash server host:port")
+		lightstepAccessToken = fs.String("lightstep_access_token", "", "Enable LightStep tracing via a LightStep access token")
```
peterbourgon: Naming likewise in the addsvc/client.

examples/addsvc/main.go
```diff
@@ -78,20 +84,22 @@ func main() {
...
+		if len(*lightstepAccessToken) > 0 {
+			if tracer != nil {
+				panic("Attempted to configure multiple OpenTracing implementations")
 			}
+			tracer = lightstep.NewTracer(lightstep.Options{
+				AccessToken: *lightstepAccessToken,
+			})
+		}
+		if tracer == nil {
+			tracer = opentracing.GlobalTracer() // the noop tracer
```
peterbourgon: This block restructured likewise in addsvc/client.

examples/apigateway/main.go
peterbourgon: +1

tracing/opentracing/endpoint.go
```diff
@@ -0,0 +1,48 @@
...
+				// All we can do is create a new root span.
+				serverSpan = tracer.StartSpan(operationName)
+			} else {
+				serverSpan.SetOperationName(operationName)
+			}
+			serverSpan.SetTag("span.kind", "server")
```
peterbourgon: Is `span.kind` part of a spec, i.e. should this be a constant?

tracing/opentracing/endpoint.go
```diff
@@ -0,0 +1,48 @@
...
+// OpenTracing Span called `operationName`.
+func TraceClient(tracer opentracing.Tracer, operationName string) endpoint.Middleware {
+	return func(next endpoint.Endpoint) endpoint.Endpoint {
+		return func(ctx context.Context, request interface{}) (interface{}, error) {
+			parentSpan := opentracing.SpanFromContext(ctx)
+			clientSpan := tracer.StartSpanWithOptions(opentracing.StartSpanOptions{
+				OperationName: operationName,
+				Parent:        parentSpan, // may be nil
+			})
+			clientSpan.SetTag("span.kind", "client")
```
peterbourgon: Likewise above?

tracing/opentracing/grpc.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// ToGRPCRequest returns a grpc RequestFunc that injects an OpenTracing Span
+// found in `ctx` into the grpc Metadata. If no such Span can be found, the
+// RequestFunc is a noop.
+func ToGRPCRequest(tracer opentracing.Tracer) func(ctx context.Context, md *metadata.MD) context.Context {
+	return func(ctx context.Context, md *metadata.MD) context.Context {
+		if span := opentracing.SpanFromContext(ctx); span != nil {
+			// There's nothing we can do with an error here.
+			_ = tracer.Inject(span, opentracing.TextMap, metadataReaderWriter{md})
```
peterbourgon: Why _ = ?

tracing/opentracing/grpc.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		return ctx
+	}
+}
```
peterbourgon: I'm curious, as I'm sure you have more tracing experience than I do: is the to/from gRPC methodology sane? Is there a better or different, standardized way of passing tracing context thru gRPC? (Or other transports, for that matter?)

peterbourgon:
Effectively all of my comments were nitpicking. This is a perfect approach.

> If there's interest, we could try to port the tracing/zipkin package to be a pure OpenTracing implementation. I don't know how many depend on its API already, though.

Yes, please! I'm all for standardization here. (It can be a separate PR if you think it might be big.) And for the record Go kit has no problem with API breakage; all of our consumers are vendoring.

basvanbeek:
No problem for me provided we have a proper OpenTracing driver for Zipkin which allow us to have all the features and functionality we currently have with the native implementation.

tracing/opentracing/endpoint.go
bensigelman: @bg451 `opentracing.StartSpanFromContext` relies on the global/singleton Tracer

examples/addsvc/client/main.go
bensigelman: @peterbourgon if we intend to keep more than one of these bindings, I'll definitely clean this up and change the flag names, etc... I just included both LightStep and Appdash to demonstrate how easy it is to switch them for this RFC. Preferences/thoughts on this?

tracing/opentracing/endpoint.go
bensigelman: TL;DR yes they should be. There's a long story that's not worth going into. There's something reasonable I can do in the interim...

tracing/opentracing/grpc.go
bensigelman: See the line above... since I didn't thread a logger through to report internal errors like this, there's nothing useful to be done with the `error`. There's some tradeoff to the simplicity of the API and how much data we can extract about various error cases.

tracing/opentracing/endpoint.go
yurishkuro: @bensigelman - they are in ext.tags

tracing/opentracing/grpc.go
peterbourgon: Right, that's fine, but why not just write `tracer.Inject`? Why do you need the explicit `_ =`?

tracing/opentracing/grpc.go
bensigelman: Each transport is its own beast... certain "transports" (sic) like the memcached protocol or SQL statements don't have a clean place to put true end-to-end tracing ids. The gRPC metadata approach seems perfectly reasonable to me... I have nits/etc, but it's mostly just matters of personal preference.

tracing/opentracing/grpc.go
yurishkuro: @bensigelman we usually emit metrics on these errors. Although they are more useful in the extract path.

tracing/opentracing/grpc.go
peterbourgon: Got it. Yeah I'm concerned because the strength of tracing relies so much on standardization of mechanism i.e. the way you shuttle the IDs between process boundaries... I'm so hungry for a spec here :)

bensigelman:
@peterbourgon @basvanbeek I am going to get coverage in for other transports, add tests, then ping this PR for another (closer) look.

I don't think it makes sense to port the Zipkin client over to OpenTracing as part of this PR... though I do think it's feasible and desirable.

tracing/opentracing/http.go
```diff
@@ -0,0 +1,40 @@
...
+// can be retrieved with opentracing.SpanFromContext(ctx).
+func FromHTTPRequest(tracer opentracing.Tracer, operationName string) kithttp.RequestFunc {
+	return func(ctx context.Context, req *http.Request) context.Context {
+		// Try to join to a trace propagated in `req`. There's nothing we can
+		// do with any errors here, so we ignore them.
+		span, _ := tracer.Join(operationName, opentracing.TextMap, opentracing.HTTPHeaderTextMapCarrier(req.Header))
+		if span == nil {
+			span = opentracing.StartSpan(operationName)
+		}
+		return opentracing.ContextWithSpan(ctx, span)
```
yurishkuro: we usually capture `http.url` tag here

examples/addsvc/main.go
peterbourgon: All else equal, yes.

tracing/opentracing/endpoint.go
bensigelman: @yurishkuro the values there are not up to spec... but I made that same change in my local client since it's an improvement.

tracing/opentracing/http.go
```diff
@@ -0,0 +1,40 @@
...
+
+// ToHTTPRequest returns an http RequestFunc that injects an OpenTracing Span
+// found in `ctx` into the http headers. If no such Span can be found, the
+// RequestFunc is a noop.
+func ToHTTPRequest(tracer opentracing.Tracer) kithttp.RequestFunc {
+	return func(ctx context.Context, req *http.Request) context.Context {
+		// Try to find a Span in the Context.
+		if span := opentracing.SpanFromContext(ctx); span != nil {
+			// There's nothing we can do with any errors here.
+			_ = tracer.Inject(span, opentracing.TextMap, opentracing.HTTPHeaderTextMapCarrier(req.Header))
```
yurishkuro: useful to capture `http.url` and peer tags (unless can be done in middleware funcs)

tracing/opentracing/grpc.go
bensigelman: @peterbourgon it's a matter of personal preference, but I like documenting ignored return values in this way... I think it helps a code-reader understand that something fishy is going on (which, in this case, is absolutely the case!).

peterbourgon: Hmm. Do you do it for fmt.Sprintf, too? ;)

tracing/opentracing/grpc.go
bensigelman: @peterbourgon I think a cross-tracing-system wire-format spec is difficult to achieve for fundamental reasons... That said, I am pleased with the approach in OpenTracing where we standardize on the concepts and basic formats (i.e., text maps and binary blobs) for propagation but leave the actual field choices to the tracing implementations... it provides interop at the instrumentation layer without forcing tracing systems to use precisely the same data model (because they don't!).

tracing/opentracing/endpoint.go
yurishkuro: you mean "s" vs. "server"? Strictly speaking, it doesn't matter much. As long as the tracer is using the same constants as the OT lib, it can remap them any way it wants. In our Zipkin-like tracer we don't store them anyway, but use them to drive Zipkin's cs/cr/ss/sr annotations instead.

but yeah we should fix them. I think the values were there way before the spec was agreed on.

tracing/opentracing/http.go
yurishkuro: and peer tags too - in Zipkin they drive SA/CA annotations.

tracing/opentracing/grpc.go
bensigelman: If you feel strongly I can remove it... as I said, it's a matter of personal preference. And no, I don't do this for fmt.Sprintf because I assume the reader understands the contract there already :)

tracing/opentracing/grpc.go
peterbourgon: I certainly couldn't hope for a cross-tracing-_system_ spec. But I'd love it if things were well defined for X-over-Y, X ∋ {Zipkin, Lightstep, Appdash, ...}, Y ∋ {HTTP, gRPC, Thrift, ...}.

basvanbeek: We might have to make an active contribution in this place. It would be good to help set or refine the standard if none or a half finished one exists.

When I added the gRPC propagation for Zipkin there was nothing defined yet. Now we have lowercased B3 keys over gRPC metadata supported in both Go kit and Java Brave making them interoperable. It has not been documented on the Zipkin site yet, but I think it should.

bensigelman: @peterbourgon @basvanbeek one nice thing about OpenTracing is that it takes the "Y" out of your set-theory statement above... by defining how a given impl gets in and out of a string map and binary blob, OpenTracing instrumentation (that's tracer/"X"-agnostic) deals with the individual protocols.

bensigelman:
Current status: blocked on https://github.com/opentracing/opentracing-go/pull/78

Once that's in, I still need to:
- add a `Logger` for debugging... @peterbourgon that will also obviate the controversial `_ := ...` thing since I'll be able to do something with the error :)
- add the various HTTP annotations per Yuri's suggestion
- same with peer identity info (mainly to make it feasible for an OT-zipkin bridge to function properly down the line)

peterbourgon:
Cool! Just a future nit: if you'll add an optional debug logger, I'd want to see the constructor moved over to a functional options idiom.

bensigelman:
Sure thing!

On Thursday, May 5, 2016, Peter Bourgon <notifications@github.com> wrote:

> Cool! Just a future nit: if you'll add an optional debug logger, I'd want
> to see the constructor moved over to a functional options idiom.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/go-kit/kit/pull/247#issuecomment-217289822>
>


-- 
(Sent via mobile)

tracing/opentracing/http.go
bensigelman: @yurishkuro I'm not clear on how you capture peer tags from the server-side of an `http.Request`... from the client it seems easier, of course. Pro-tips?

yurishkuro: hm, you're right, I just checked and for Go I didn't do much on the server side. In Python it looked like this:

```python
    span.set_tag(tags.SPAN_KIND, tags.SPAN_KIND_RPC_SERVER)
    span.set_tag('http.url', request.full_url)

    remote_ip = request.remote_ip
    if remote_ip:
        span.set_tag(tags.PEER_HOST_IPV4, remote_ip)

    caller_name = request.caller_name
    if caller_name:
        span.set_tag(tags.PEER_SERVICE, caller_name)

    remote_port = request.remote_port
    if remote_port:
        span.set_tag(tags.PEER_PORT, remote_port)
```

The `caller_name` was a custom behavior based on a special HTTP tag that our services send to identify themselves, and the instrumentation is configured with the name of that header. And Python libs usually expose remote addr to the server, similar to Java servlets. Go's `Request` also has `RemoteAddr` field, but I was lazy to parse it (note to self: don't be lazy, fix it).

bensigelman:
@peterbourgon I updated this PR per review comments. A few things to know:
- there are tests now :)
- per the original form of the PR, I started with the bare http and grpc transports; little did I know that the other transports would be essentially impossible to instrument without modifying application protocols. :-/
- I added a few OpenTracing Span "tags" here and there... I mentioned this in a particular comment to @yurishkuro, but I wasn't sure how to gain access to some of the peer info that Zipkin benefits from. I glanced at the `kit/tracing/zipkin` package, and it doesn't seem to have a way to access it, either.
- you asked for a functional options idiom: I don't see how this would work without making some pretty fundamental changes to thread a `Logger` (or some callback mechanism) through OpenTracing's APIs. I compromised and added `logger` parameters to various functions in the `kit/tracing/opentracing` package... Did you have something better in mind? I'm open to suggestions but the the functional options pattern (as I understand it) requires some object for the functional options to modify... in this case, I wasn't clear what that object would be.

Thanks for the review!

examples/addsvc/client/grpc/factory.go
```diff
@@ -7,33 +7,45 @@ import (
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/pb"
+	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
 	grpctransport "github.com/go-kit/kit/transport/grpc"
+	"github.com/opentracing/opentracing-go"
```
peterbourgon: Bump this one up with the other third-party import statement, grpc.

examples/addsvc/client/httpjson/factory.go
```diff
@@ -6,45 +6,59 @@ import (
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/server"
+	"github.com/go-kit/kit/loadbalancer"
+	"github.com/go-kit/kit/log"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
 	httptransport "github.com/go-kit/kit/transport/http"
+	"github.com/opentracing/opentracing-go"
```
peterbourgon: Likewise

examples/addsvc/client/grpc/factory.go
```diff
@@ -7,33 +7,45 @@ import (
...
-	cc, err := grpc.Dial(instance, grpc.WithInsecure())
-	return grpctransport.NewClient(
-		cc,
-		"Add",
-		"Sum",
-		encodeSumRequest,
-		decodeSumResponse,
-		pb.SumReply{},
-	).Endpoint(), cc, err
+func MakeSumEndpointFactory(tracer opentracing.Tracer, tracingLogger log.Logger) loadbalancer.Factory {
```
peterbourgon: Yeah, I think this is a reasonable-enough idiom for threading the logger through. My only other thought would be to define a Go kit tracer struct with the opentracing.Tracer + log.Logger together, but I think that's overkill at this point.

examples/addsvc/client/main.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/go-kit/kit/loadbalancer"
 	"github.com/go-kit/kit/loadbalancer/static"
 	"github.com/go-kit/kit/log"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	"github.com/lightstep/lightstep-tracer-go"
+	"github.com/opentracing/opentracing-go"
+	appdashot "github.com/sourcegraph/appdash/opentracing"
+	"sourcegraph.com/sourcegraph/appdash"
```
peterbourgon: Import grouping likewise above

examples/addsvc/client/main.go
```diff
@@ -48,6 +57,25 @@ func main() {
...
+			tracer = lightstep.NewTracer(lightstep.Options{
+				AccessToken: *lightstepAccessToken,
+			})
+			defer lightstep.FlushLightStepTracer(tracer)
+		case *appdashAddr == "" && *lightstepAccessToken == "":
+			tracer = opentracing.GlobalTracer() // no-op
+		default:
+			panic("specify either -appdash.addr or -lightstep.access.token, not both")
+		}
+	}
```
peterbourgon: 👍

examples/addsvc/grpc_binding.go
```diff
@@ -6,17 +6,32 @@ import (
 	"github.com/go-kit/kit/examples/addsvc/pb"
 	"github.com/go-kit/kit/examples/addsvc/server"
 	servergrpc "github.com/go-kit/kit/examples/addsvc/server/grpc"
+	"github.com/go-kit/kit/log"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
 	"github.com/go-kit/kit/transport/grpc"
+	"github.com/opentracing/opentracing-go"
```
peterbourgon: Import grouping likewise above

examples/addsvc/main.go
```diff
@@ -15,9 +15,14 @@ import (
...
+	"github.com/lightstep/lightstep-tracer-go"
+	"github.com/opentracing/opentracing-go"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
+	appdashot "github.com/sourcegraph/appdash/opentracing"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"sourcegraph.com/sourcegraph/appdash"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/examples/addsvc/pb"
```
peterbourgon: Import grouping for this whole block likewise above.

tracing/opentracing/endpoint_test.go
```diff
@@ -0,0 +1,95 @@
+package opentracing_test
+
+import (
+	"testing"
+
+	"github.com/go-kit/kit/endpoint"
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/mocktracer"
+	"golang.org/x/net/context"
```
peterbourgon: Import grouping likewise above

tracing/opentracing/endpoint_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func TestTraceServer(t *testing.T) {
+	tracer := mocktracer.New()
```
peterbourgon: Cool that this is exported!

tracing/opentracing/grpc_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"testing"
+
+	"google.golang.org/grpc/metadata"
+
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	"github.com/go-kit/kit/transport/grpc"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/mocktracer"
+	"golang.org/x/net/context"
```
peterbourgon: Import grouping likewise above

tracing/opentracing/http.go
```diff
@@ -0,0 +1,74 @@
...
+	"net"
+	"net/http"
+	"strconv"
+
+	"github.com/go-kit/kit/log"
+	kithttp "github.com/go-kit/kit/transport/http"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+	"golang.org/x/net/context"
+)
```
peterbourgon: Import grouping likewise above

tracing/opentracing/http_test.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"net/http"
+	"testing"
+
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	kithttp "github.com/go-kit/kit/transport/http"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/mocktracer"
+	"golang.org/x/net/context"
+)
```
peterbourgon: Import grouping likewise above

tracing/opentracing/grpc.go
```diff
@@ -0,0 +1,67 @@
+package opentracing
+
+import (
+	"github.com/go-kit/kit/log"
+	"github.com/opentracing/opentracing-go"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc/metadata"
+)
```
peterbourgon: Import grouping likewise above

peterbourgon:
The same nit repeated a billion times :) Thanks a ton for this! Fix those little things up and we'll get this merged straight away.

tracing/opentracing/endpoint_test.go
bensigelman: yes, I like it, too :)

bensigelman:
@peterbourgon I adjusted the PR per your suggestions and imports are now segmented per `go-kit` conventions. PTAL... I think this should be good to merge.

I would like to do some work adding documentation for this change and there's still the matter of making the `go-kit` zipkin impl interface-compatible, but I think those can happen separately.

peterbourgon:
Looks great! I can't thank you enough. And can finally close that Appdash ticket from a thousand years ago. I've filed #256 and #260 for those follow-up issues.

PR #253, opened 2016-05-06
log: Reuse encoders and buffers for Logfmt logger

nuss-justin:
This uses sync.Pool to avoid always allocating new encoders and buffers.

Reduces the number of allocations percall to Log by 2, if a value exists
in the pool.

Benchstat (Go 1.6.2, 5 runs each):

```
name                      old time/op    new time/op    delta
LogfmtLoggerSimple-8         869ns ± 1%     864ns ± 2%     ~     (p=0.175 n=5+5)
LogfmtLoggerContextual-8    1.28µs ± 0%    1.28µs ± 1%     ~     (p=0.603 n=5+5)

name                      old alloc/op   new alloc/op   delta
LogfmtLoggerSimple-8          384B ± 0%      128B ± 0%  -66.67%  (p=0.008 n=5+5)
LogfmtLoggerContextual-8      560B ± 0%      304B ± 0%  -45.71%  (p=0.008 n=5+5)

name                      old allocs/op  new allocs/op  delta
LogfmtLoggerSimple-8          6.00 ± 0%      4.00 ± 0%  -33.33%  (p=0.008 n=5+5)
LogfmtLoggerContextual-8      9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
```

Same for Go TIp (devel +0b6e5e3):

```
name                      old time/op    new time/op    delta
LogfmtLoggerSimple-8         825ns ± 1%     796ns ± 0%   -3.51%  (p=0.008 n=5+5)
LogfmtLoggerContextual-8    1.12µs ± 1%    1.11µs ± 0%   -1.18%  (p=0.008 n=5+5)

name                      old alloc/op   new alloc/op   delta
LogfmtLoggerSimple-8          384B ± 0%      128B ± 0%  -66.67%  (p=0.008 n=5+5)
LogfmtLoggerContextual-8      560B ± 0%      304B ± 0%  -45.71%  (p=0.008 n=5+5)

name                      old allocs/op  new allocs/op  delta
LogfmtLoggerSimple-8          6.00 ± 0%      4.00 ± 0%  -33.33%  (p=0.008 n=5+5)
LogfmtLoggerContextual-8      9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
```

Command: `go test -bench Logfmt -benchmem -count 5`
Tested on a linux/amd64

### DIFF
--- a/log/logfmt_logger.go
+++ b/log/logfmt_logger.go
@@ -1,10 +1,30 @@
 package log
 
 import (
+	"bytes"
 	"io"
+	"sync"
 
 	"github.com/go-logfmt/logfmt"
 )
+
+type logfmtEncoder struct {
+	*logfmt.Encoder
+	buf bytes.Buffer
+}
+
+func (l *logfmtEncoder) Reset() {
+	l.Encoder.Reset()
+	l.buf.Reset()
+}
+
+var logfmtEncoderPool = sync.Pool{
+	New: func() interface{} {
+		var enc logfmtEncoder
+		enc.Encoder = logfmt.NewEncoder(&enc.buf)
+		return &enc
+	},
+}
 
 type logfmtLogger struct {
 	w io.Writer
@@ -18,16 +38,23 @@ func NewLogfmtLogger(w io.Writer) Logger {
 }
 
 func (l logfmtLogger) Log(keyvals ...interface{}) error {
-	// The Logger interface requires implementations to be safe for concurrent
-	// use by multiple goroutines. For this implementation that means making
-	// only one call to l.w.Write() for each call to Log. We first collect all
-	// of the bytes into b, and then call l.w.Write(b).
-	b, err := logfmt.MarshalKeyvals(keyvals...)
-	if err != nil {
+	enc := logfmtEncoderPool.Get().(*logfmtEncoder)
+	enc.Reset()
+	defer logfmtEncoderPool.Put(enc)
+
+	if err := enc.EncodeKeyvals(keyvals...); err != nil {
 		return err
 	}
-	b = append(b, '\n')
-	if _, err := l.w.Write(b); err != nil {
+
+	// Add newline to the end of the buffer
+	if err := enc.EndRecord(); err != nil {
+		return err
+	}
+
+	// The Logger interface requires implementations to be safe for concurrent
+	// use by multiple goroutines. For this implementation that means making
+	// only one call to l.w.Write() for each call to Log.
+	if _, err := l.w.Write(enc.buf.Bytes()); err != nil {
 		return err
 	}
 	return nil


ChrisHines:
Thanks! After a quick read this looks nice. I'll take a careful look soon.

log/logfmt_logger.go
```diff
@@ -18,16 +38,45 @@ func NewLogfmtLogger(w io.Writer) Logger {
...
 func (l logfmtLogger) Log(keyvals ...interface{}) error {
-	// The Logger interface requires implementations to be safe for concurrent
-	// use by multiple goroutines. For this implementation that means making
-	// only one call to l.w.Write() for each call to Log. We first collect all
-	// of the bytes into b, and then call l.w.Write(b).
-	b, err := logfmt.MarshalKeyvals(keyvals...)
-	if err != nil {
+	// Code copied from logfmt.MarshalKeyvals and modified
+	// to reuse encoders and buffers with sync.Pool and
+	// call EndRecord
```
nuss-justin: @ChrisHines Do you think it would make sense to have a `MarshalKeyvalsTo` (name for debate) in the logfmt package that takes an *Encoder instead of allocating a new one? This way `MarshalKeyvals` could be implemented by calling `MarshalKeyvalsTo` with a bytes.Buffer and we could avoid having to duplicate the code here.

Something like this 

```go
func MarshalKeyvalsTo(enc *Encoder, keyvals ...interface{}) error
```

ChrisHines: Yes, although I think adding a method to `logfmt.Encoder` makes more sense.

```go
func (enc *Encoder) EncodeKeyvals(keyvals ...interface{}) error
```

See: https://github.com/go-logfmt/logfmt/commit/a0ff333161fe5c2daed0ba52d1792bd3d2531b94.

ChrisHines:
@nuss-justin: Please update for the new `Encoder.EncodeKeyvals` method in the logfmt package.

Also note that I pushed another speed improvement to the logfmt package earlier. See: https://github.com/go-logfmt/logfmt/commit/a5fe813d6880436150bdf16a66974092c5343f68.

nuss-justin:
Updated commit to use the new EncodeKeyvals method on *logfmt.Encoder. Also updated the benchmarks.

log/logfmt_logger.go
```diff
@@ -18,16 +38,25 @@ func NewLogfmtLogger(w io.Writer) Logger {
...
 	}
-	b = append(b, '\n')
-	if _, err := l.w.Write(b); err != nil {
+
+	// The Logger interface requires implementations to be safe for concurrent
+	// use by multiple goroutines. For this implementation that means making
+	// only one call to l.w.Write() for each call to Log. We call Write ourself
+	// with buf.Bytes() instead of using buf.WriteTo(l.w) as WriteTo can call
+	// Write multiple times.
+	if _, err := l.w.Write(enc.buf.Bytes()); err != nil {
```
ChrisHines: The last sentence of this comment is incorrect. Please remove.

The current implementation for `bytes.Buffer.WriteTo` will only make one call to `Write`. We use `buf.Bytes` in favor of `buf.WriteTo` because `buf.Bytes` is faster. `buf.WriteTo` updates the read position of `buf` based on the number of bytes written, while `buf.Bytes` does not. Since we discard unread data in `buf` after returning from this method, we don't need to pay the overhead of updating its read position.

ChrisHines:
I'm surprised your new CPU benchmark shows no improvement with the new code. Is that right? On my machine (go1.6.2 windows/amd64) I get:

```
name                      old time/op    new time/op    delta
LogfmtLoggerSimple-4        2.07µs ± 1%    1.80µs ± 1%  -13.12%  (p=0.000 n=10+10)
LogfmtLoggerContextual-4    3.04µs ± 1%    2.74µs ± 1%  -10.00%   (p=0.000 n=9+10)

name                      old alloc/op   new alloc/op   delta
LogfmtLoggerSimple-4          384B ± 0%      128B ± 0%  -66.67%  (p=0.000 n=10+10)
LogfmtLoggerContextual-4      560B ± 0%      304B ± 0%  -45.71%  (p=0.000 n=10+10)

name                      old allocs/op  new allocs/op  delta
LogfmtLoggerSimple-4          6.00 ± 0%      4.00 ± 0%  -33.33%  (p=0.000 n=10+10)
LogfmtLoggerContextual-4      9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.000 n=10+10)
```

But that is good news. A 10% improvement is nice. When combined with the recent improvement in the logfmt package the total improvement is very nice:

```
name                      old time/op    new time/op    delta
LogfmtLoggerSimple-4        2.42µs ± 2%    1.80µs ± 1%  -25.68%  (p=0.000 n=10+10)
LogfmtLoggerContextual-4    3.55µs ± 3%    2.74µs ± 1%  -22.86%  (p=0.000 n=10+10)
```

nuss-justin:
@ChrisHines Removed the last sentence and updated my benchmarks again. Also added results for current Go Tip. On my machine there's no change in time/op for 1.6.2.

ChrisHines:
@nuss-justin Thanks. Good work.

LGTM.

ChrisHines:
@ax-nathan I haven't tagged a new release of go-logfmt/logfmt yet, that could be why glide isn't pulling it in for you. Let me know. I will tag a release when I get a chance (with 24h).

ax-nathan:
This is the workaround for the bug in `glide` that trigged my issue: https://github.com/Masterminds/glide/issues/360#issuecomment-218226872

Good now. Thanks again!

PR #254, opened 2016-05-08
Added lb.Retry function

basvanbeek:
This function allows for calling an endpoint with load balanced retry logic using the new sd package in the same way as we had available in the old loadbalancer package.


peterbourgon:
Thank you! Missed this in the port.

Issue #251, opened 2016-05-05
Create a new `kit/metrics/graphite` package for aggregating and emitting metrics to Graphite


jprobinson:
To help transition users of go-metrics & go-metrics-graphite to kit's metrics package, I would like a new package for aggregating counters, timers and gauges and periodically emitting the stats to graphite directly.

jprobinson:
I'm going to spend some time over the few couple of days taking a whack at this one.

So far, I'm playing with creating an interface to tie the different metric types together and periodically push the data to a graphite host via a shared TCP connection (similar to go-metrics's Registry + go-metrics-graphite's Graphite type).

At this point, the interface is looking like:
```
type GraphiteEmitter interface {
    NewCounter(string) metrics.Counter
    NewHistogram(string) metrics.Histogram
    NewGauge(string) metrics.Gauge

    Start()
    Emit() error
}
```
The pattern of use would be:
  1. create a new 'emitter': `NewGrahpiteEmitter(addr *net.TCPAddr, prefix string, interval time.Duration) GraphiteEmitter`
  2. add any counters, gauges or histograms needed
  3. `Start()` the emitter to push the data every interval
  4. Users call `Emit()` in case the process needs to shut down and current stats need to be flushed to graphite first.

jprobinson:
So far, gauges and counters are no problem. The more I look at histograms, though, the more I'm thinking we might not to write _another_ implementation. 

On that note, any thoughts on bringing https://github.com/codahale/hdrhistogram in for just for this package?

peterbourgon:
@jprobinson That's what I did in metrics/expvar :) I think it's a fine idea. s/Emit/Flush?

jprobinson:
👍 awesome.

I went back and forth on Emit/Flush too. Flush is definitely a more recognizable method for this purpose. I guess I was trying to avoid creating a 'GraphiteFlusher' interface as it sounds a bit like metrics would be going down the toilet. 😜 I'll keep the `GraphiteEmitter` monicker, but use `Flush()`.

peterbourgon:
:tada:

PR #259, opened 2016-05-11
[PENDING] Adding package `metrics/graphite`

jprobinson:
This is the first whack at resolving issue #251.

With this new package, metrics will be managed and emitted to a Graphite host via a the new `Emitter`:

```
type Emitter interface {
    NewCounter(string) metrics.Counter
    NewHistogram(string, int64, int64, int, ...int) metrics.Histogram
    NewTimeHistogram(string, time.Duration, int64, int64, int, ...int) metrics.TimeHistogram
    NewGauge(string) metrics.Gauge

    Start(time.Duration)
    Flush() error
}
```

The normal pattern of use should be:
  1. Get a new emitter via `NewEmitter`
  2. Register any metrics with the emitter (`NewCounter`, `NewGauge`, etc.)
  3. Call `Start` on the emitter to kick off a goroutine to periodically emit metrics to Graphite.
  4. Call `Flush` on the emitter at shutdown to flush current metrics

### DIFF
--- a/metrics/graphite/graphite.go
+++ b/metrics/graphite/graphite.go
@@ -175,18 +175,48 @@ func (e *Emitter) Stop() error {
 	return err
 }
 
+var (
+	RetryMax        = 10
+	RetryWait       = 2 * time.Millisecond
+	RetryMultiplier = 2
+)
+
 // Flush will write the current metrics to the Emitter's
 // connection in the Graphite plaintext protocol.
-func (e *Emitter) Flush() { e.flush(e.conn) }
-
-func (e *Emitter) flush(conn io.Writer) {
+func (e *Emitter) Flush() error {
 	// only one flush at a time
 	e.mtx.Lock()
 	defer e.mtx.Unlock()
 
+	// set the system up to perform a retry loop
+	var err error
+	wait := RetryWait
+	for i := 1; i <= RetryMax; i++ {
+		err = e.flush(e.conn)
+		if err == nil {
+			return nil
+		}
+		e.logger.Log(
+			"err", err,
+			"msg", fmt.Sprintf("unable to flush metrics on attempt %d, waiting %s", i, wait),
+		)
+		time.Sleep(wait)
+		wait = wait * time.Duration(RetryMultiplier)
+	}
+	// log if we were unable to emit metrics
+	if err != nil {
+		e.logger.Log(
+			"err", err,
+			"msg", fmt.Sprintf("unable to flush metrics after %d attempts. giving up.", RetryMax),
+		)
+	}
+	return err
+}
+
+func (e *Emitter) flush(conn io.Writer) error {
+
 	// buffer the writer and make sure to flush it
 	w := bufio.NewWriter(conn)
-	defer w.Flush()
 
 	// emit counter stats
 	for _, c := range e.counters {
@@ -208,6 +238,9 @@ func (e *Emitter) flush(conn io.Writer) {
 	for _, g := range e.gauges {
 		fmt.Fprintf(w, "%s.%s %.2f %d\n", e.prefix, g.Name(), g.Get(), time.Now().Unix())
 	}
+
+	// check for error
+	return w.Flush()
 }
 
 type counter struct {
--- a/metrics/graphite/graphite.go
+++ b/metrics/graphite/graphite.go
@@ -186,7 +186,11 @@ func (e *emitter) Stop() error {
 	// get one last flush in
 	e.Flush()
 	// close the connection
-	return e.conn.Close()
+	err := e.conn.Close()
+	// nil the conn to avoid problems
+	// if Stop() is called more than once.
+	e.conn = nil
+	return err
 }
 
 // Flush will attempt to create a connection with the given address
--- a/metrics/graphite/graphite.go
+++ b/metrics/graphite/graphite.go
@@ -1,0 +1,296 @@
+// Package graphite implements a graphite backend for package metrics.
+//
+// The current implementation ignores fields.
+package graphite
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"log"
+	"math"
+	"net"
+	"sort"
+	"sync"
+	"time"
+
+	"sync/atomic"
+
+	"github.com/codahale/hdrhistogram"
+	"github.com/go-kit/kit/metrics"
+)
+
+// Emitter will keep track of all metrics and, once started,
+// will emit the metrics via the Flush method to the given io.Writer.
+type Emitter interface {
+	NewCounter(string) metrics.Counter
+	NewHistogram(string, int64, int64, int, ...int) metrics.Histogram
+	NewTimeHistogram(string, time.Duration, int64, int64, int, ...int) metrics.TimeHistogram
+	NewGauge(string) metrics.Gauge
+
+	Start(time.Duration)
+	Flush() error
+}
+
+type emitter struct {
+	addr   *net.TCPAddr
+	prefix string
+
+	metricMu   *sync.Mutex
+	counters   []*counter
+	histograms []*windowedHistogram
+	gauges     []*gauge
+}
+
+// NewEmitter will return an Emitter that will prefix all
+// metrics names with the given prefix. Once started, it will attempt to create
+// a TCP connection with the given address and most metrics to the connection
+// in a Graphite-compatible format.
+func NewEmitter(addr *net.TCPAddr, prefix string) Emitter {
+	e := &emitter{
+		addr, prefix, &sync.Mutex{},
+		[]*counter{}, []*windowedHistogram{}, []*gauge{},
+	}
+
+	return e
+}
+
+// NewCounter returns a Counter whose value will be periodically emitted in
+// a Graphite-compatible format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewCounter(name string) metrics.Counter {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+	c := &counter{name, 0}
+	e.counters = append(e.counters, c)
+	return c
+}
+
+// NewHistogram is taken from http://github.com/codahale/metrics. It returns a
+// windowed HDR histogram which drops data older than five minutes.
+//
+// The histogram exposes metrics for each passed quantile as gauges. Quantiles
+// should be integers in the range 1..99. The gauge names are assigned by
+// using the passed name as a prefix and appending "_pNN" e.g. "_p50".
+//
+// The values of this histogram will be periodically emitted in a Graphite-compatible
+// format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewHistogram(name string, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.Histogram {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+
+	gauges := map[int]metrics.Gauge{}
+	for _, quantile := range quantiles {
+		if quantile <= 0 || quantile >= 100 {
+			panic(fmt.Sprintf("invalid quantile %d", quantile))
+		}
+		gauges[quantile] = e.gauge(fmt.Sprintf("%s_p%02d", name, quantile))
+	}
+	h := newWindowedHistogram(name, minValue, maxValue, sigfigs, gauges)
+	e.histograms = append(e.histograms, h)
+	return h
+}
+
+// NewTimeHistogram returns a TimeHistogram wrapper around the windowed
+// HDR histrogram provided by this package.
+func (e *emitter) NewTimeHistogram(name string, unit time.Duration, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.TimeHistogram {
+	h := e.NewHistogram(name, minValue, maxValue, sigfigs, quantiles...)
+	return metrics.NewTimeHistogram(unit, h)
+}
+
+// NewGauge returns a Gauge whose value will be periodically emitted in
+// a Graphite-compatible format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewGauge(name string) metrics.Gauge {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+	return e.gauge(name)
+}
+
+func (e *emitter) gauge(name string) metrics.Gauge {
+	g := &gauge{name, 0}
+	e.gauges = append(e.gauges, g)
+	return g
+}
+
+// Start will kick off a background goroutine to
+// call Flush once every interval.
+func (e *emitter) Start(interval time.Duration) {
+	go func() {
+		t := time.Tick(interval)
+		for range t {
+			err := e.Flush()
+			if err != nil {
+				log.Print("error: could not dial graphite host: ", err)
+				continue
+			}
+		}
+	}()
+}
+
+// Flush will attempt to create a connection with the given address
+// and write the current metrics to it in a Graphite-compatible format.
+//
+// Users can call this method on process shutdown to ensure
+// the current metrics are pushed to Graphite.
+func (e *emitter) Flush() error {
+	// open connection
+	conn, err := net.DialTCP("tcp", nil, e.addr)
+	if err != nil {
+		return err
+	}
+
+	// flush stats to connection
+	e.flush(conn)
+
+	// close connection
+	conn.Close()
+	return nil
+}
+
+func (e *emitter) flush(conn io.Writer) {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+
+	// buffer the writer and make sure to flush it
+	w := bufio.NewWriter(conn)
+	defer w.Flush()
+
+	now := time.Now().Unix()
+
+	// emit counter stats
+	for _, c := range e.counters {
+		fmt.Fprintf(w, "%s.%s.count %d %d\n", e.prefix, c.Name(), c.count, now)
+	}
+
+	// emit histogram specific stats
+	for _, h := range e.histograms {
+		hist := h.hist.Merge()
+		fmt.Fprintf(w, "%s.%s.count %d %d\n", e.prefix, h.Name(), hist.TotalCount(), now)
+		fmt.Fprintf(w, "%s.%s.min %d %d\n", e.prefix, h.Name(), hist.Min(), now)
+		fmt.Fprintf(w, "%s.%s.max %d %d\n", e.prefix, h.Name(), hist.Max(), now)
+		fmt.Fprintf(w, "%s.%s.mean %.2f %d\n", e.prefix, h.Name(), hist.Mean(), now)
+		fmt.Fprintf(w, "%s.%s.std-dev %.2f %d\n", e.prefix, h.Name(), hist.StdDev(), now)
+	}
+
+	// emit gauge stats (which can include some histogram quantiles)
+	for _, g := range e.gauges {
+		fmt.Fprintf(w, "%s.%s %.2f %d\n", e.prefix, g.Name(), g.Get(), now)
+	}
+}
+
+type counter struct {
+	key   string
+	count uint64
+}
+
+func (c *counter) Name() string { return c.key }
+
+func (c *counter) With(metrics.Field) metrics.Counter { return c }
+
+func (c *counter) Add(delta uint64) { atomic.AddUint64(&c.count, delta) }
+
+type gauge struct {
+	key   string
+	value uint64 // math.Float64bits
+}
+
+func (g *gauge) Name() string { return g.key }
+
+func (g *gauge) With(metrics.Field) metrics.Gauge { return g }
+
+func (g *gauge) Add(delta float64) {
+	for {
+		old := atomic.LoadUint64(&g.value)
+		new := math.Float64bits(math.Float64frombits(old) + delta)
+		if atomic.CompareAndSwapUint64(&g.value, old, new) {
+			return
+		}
+	}
+}
+
+func (g *gauge) Set(value float64) {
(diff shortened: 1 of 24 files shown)


metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// Emitter will keep track of all metrics and, once started,
+// will emit the metrics via the Flush method to the given address.
+type Emitter interface {
+	NewCounter(string) metrics.Counter
+	NewHistogram(string, int64, int64, int, ...int) metrics.Histogram
+	NewTimeHistogram(string, time.Duration, int64, int64, int, ...int) metrics.TimeHistogram
+	NewGauge(string) metrics.Gauge
+
+	Start(time.Duration)
+	Flush() error
```
peterbourgon: Can we have names on these parameters, so we can understand the intent from the interface? e.g.

```go
NewHistogram(name string, min, max int64, sigfigs int, quantiles ...int)
```

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+type emitter struct {
+	addr   *net.TCPAddr
+	prefix string
+
+	metricMu   *sync.Mutex
```
peterbourgon: - Canonical Go kit style is to call this `mtx`
- The non-pointer type `sync.Mutex` has a usable default value and should probably be preferred — this is fine as you're (correctly) defining methods on `*emitter`

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+	counters   []*counter
+	histograms []*windowedHistogram
+	gauges     []*gauge
```
peterbourgon: Why are these pointers?

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// metrics names with the given prefix. Once started, it will attempt to create
+// a TCP connection with the given address and periodically post metrics to the
+// connection in a Graphite-compatible format.
+func NewEmitter(addr *net.TCPAddr, prefix string) Emitter {
+	e := &emitter{
+		addr, prefix, &sync.Mutex{},
+		[]*counter{}, []*windowedHistogram{}, []*gauge{},
+	}
+
+	return e
```
peterbourgon: If you go with the above suggestions, and leverage the One Neat Trick that the zero value of a slice type can be appended-to, then this constructor can reduce to

```go
return &emitter{
    addr:    addr,
    prefix: prefix,
}
```

(Always give key prefixes to struct members during construction...)

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// NewCounter returns a Counter whose value will be periodically emitted in
+// a Graphite-compatible format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewCounter(name string) metrics.Counter {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+	c := &counter{name, 0}
+	e.counters = append(e.counters, c)
+	return c
+}
```
peterbourgon: Ah, now I see why they're pointers. Disregard :)

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// The histogram exposes metrics for each passed quantile as gauges. Quantiles
+// should be integers in the range 1..99. The gauge names are assigned by
+// using the passed name as a prefix and appending "_pNN" e.g. "_p50".
+//
+// The values of this histogram will be periodically emitted in a Graphite-compatible
+// format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewHistogram(name string, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.Histogram {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
```
peterbourgon: You should be able to move this locking after the construction of `gauges`...

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// HDR histrogram provided by this package.
+func (e *emitter) NewTimeHistogram(name string, unit time.Duration, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.TimeHistogram {
+	h := e.NewHistogram(name, minValue, maxValue, sigfigs, quantiles...)
+	return metrics.NewTimeHistogram(unit, h)
+}
+
+// NewGauge returns a Gauge whose value will be periodically emitted in
+// a Graphite-compatible format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewGauge(name string) metrics.Gauge {
+	// only one flush at a time
```
peterbourgon: Comment seems to be copied from e.g. Flush but doesn't make sense here, or?

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+	e.gauges = append(e.gauges, g)
+	return g
+}
+
+// Start will kick off a background goroutine to
+// call Flush once every interval.
+func (e *emitter) Start(interval time.Duration) {
+	go func() {
+		t := time.Tick(interval)
+		for range t {
```
peterbourgon:
```go
for range time.Tick(interval) {
```

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
+// Package graphite implements a graphite backend for package metrics.
+//
+// The current implementation ignores fields.
+package graphite
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"log"
```
peterbourgon: stdlib package log is verboten ;) If your emitter will log, please take a logger in your constructor.

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// format once the Emitter is started. Fields are ignored.
+func (e *emitter) NewHistogram(name string, minValue, maxValue int64, sigfigs int, quantiles ...int) metrics.Histogram {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+
+	gauges := map[int]metrics.Gauge{}
+	for _, quantile := range quantiles {
+		if quantile <= 0 || quantile >= 100 {
+			panic(fmt.Sprintf("invalid quantile %d", quantile))
```
peterbourgon: Given this is a constructor, should it return an error instead?

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+
+// Start will kick off a background goroutine to
+// call Flush once every interval.
+func (e *emitter) Start(interval time.Duration) {
+	go func() {
+		t := time.Tick(interval)
+		for range t {
+			err := e.Flush()
+			if err != nil {
+				log.Print("error: could not dial graphite host: ", err)
```
peterbourgon: See above comment about log :)

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+// and write the current metrics to it in a Graphite-compatible format.
+//
+// Users can call this method on process shutdown to ensure
+// the current metrics are pushed to Graphite.
+func (e *emitter) Flush() error {
+	// open connection
+	conn, err := net.DialTCP("tcp", nil, e.addr)
+	if err != nil {
+		return err
+	}
```
peterbourgon: - We probably want to maintain a persistent connection to the Graphite server, rather than re-dialing with every Flush.
- Should we offer a UDP option?

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// buffer the writer and make sure to flush it
+	w := bufio.NewWriter(conn)
+	defer w.Flush()
+
+	now := time.Now().Unix()
+
+	// emit counter stats
+	for _, c := range e.counters {
+		fmt.Fprintf(w, "%s.%s.count %d %d\n", e.prefix, c.Name(), c.count, now)
```
peterbourgon: Can you drop a link to this format on e.g. the Graphite docs into a comment somewhere?

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+func (e *emitter) flush(conn io.Writer) {
+	// only one flush at a time
+	e.metricMu.Lock()
+	defer e.metricMu.Unlock()
+
+	// buffer the writer and make sure to flush it
+	w := bufio.NewWriter(conn)
+	defer w.Flush()
+
+	now := time.Now().Unix()
```
peterbourgon: Is `now` meant to be the time of the collection, or the time of the emission?

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+func (h *windowedHistogram) Name() string                         { return h.name }
+func (h *windowedHistogram) With(metrics.Field) metrics.Histogram { return h }
+
+func (h *windowedHistogram) Observe(value int64) {
+	h.mu.Lock()
+	err := h.hist.Current.RecordValue(value)
+	h.mu.Unlock()
+
+	if err != nil {
+		panic(err.Error())
```
peterbourgon: Is this truly panic-worthy? Perhaps an alternative would be to log an error and drop the observation?

metrics/graphite/graphite_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestCounter(t *testing.T) {
+	var (
+		prefix = "prefix"
+		name   = "m"
+		value  = 123
+		e      = NewEmitter(nil, prefix)
+		b      bytes.Buffer
+	)
+	e.NewCounter(name).With(metrics.Field{Key: "ignored", Value: "field"}).Add(uint64(value))
+	e.(*emitter).flush(&b)
```
peterbourgon: Maybe easier to do this type conversion above? e.g.

```go
e = NewEmitter(nil, prefix).(*emitter)
```

metrics/teststat/graphite.go
```diff
@@ -0,0 +1,61 @@
...
+func AssertGraphiteNormalHistogram(t *testing.T, prefix, metricName string, mean, stdev int64, quantiles []int, gPayload string) {
+	const tolerance int = 2
+
+	// check for hdr histo data
+	wants := map[string]int64{"count": 1234, "min": 15, "max": 83, "std-dev": stdev, "mean": mean}
+	for key, want := range wants {
+		re := regexp.MustCompile(fmt.Sprintf("%s.%s.%s (\\d*)", prefix, metricName, key))
+		if res := re.FindAllStringSubmatch(gPayload, 1); res != nil {
+			if len(res[0]) == 1 {
+				t.Errorf("bad regex found, please check the test scenario")
```
peterbourgon: panic?

metrics/teststat/graphite.go
```diff
@@ -0,0 +1,61 @@
...
+			have, err := strconv.ParseInt(res[0][1], 10, 64)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if int(math.Abs(float64(want-have))) > tolerance {
+				t.Errorf("key %s: want %d, have %d", key, want, have)
+			}
+		} else {
+			t.Error("did not find metrics log for", key, "in \n", gPayload)
```
peterbourgon:
```go
res := re.FindAllStringSubmatch(gPayload, 1)
if res == nil {
    t.Errorf("didn't find metrics log for %q in \n%s\n", key, gPayload)
    continue
}
// Happy path
```

metrics/teststat/graphite.go
```diff
@@ -0,0 +1,61 @@
...
+			have, err := strconv.ParseFloat(res[0][1], 64)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if int(math.Abs(float64(want)-have)) > tolerance {
+				t.Errorf("quantile %d: want %.2f, have %.2f", quantile, want, have)
+			}
+		} else {
+			t.Errorf("did not find metrics log for %d", quantile)
+		}
```
peterbourgon:
```go
res := re.FindAllStringSubmatch(gPayload, 1)
if res == nil {
    t.Errorf("didn't find metrics log for %d", quantile)
    continue
}
// Happy path
```

metrics/teststat/graphite.go
```diff
@@ -0,0 +1,61 @@
...
+		re := regexp.MustCompile(fmt.Sprintf("%s.%s.%s (\\d*)", prefix, metricName, key))
+		if res := re.FindAllStringSubmatch(gPayload, 1); res != nil {
+			if len(res[0]) == 1 {
+				t.Errorf("bad regex found, please check the test scenario")
+				continue
+			}
+
+			have, err := strconv.ParseInt(res[0][1], 10, 64)
+			if err != nil {
+				t.Fatal(err)
```
peterbourgon: Given this is in a loop, maybe

```go
if err != nil {
    t.Errorf("%q: %v", key, err)
    continue
}
```

metrics/teststat/graphite.go
```diff
@@ -0,0 +1,61 @@
...
+	"math"
+	"regexp"
+	"strconv"
+	"testing"
+)
+
+// AssertGraphiteNormalHistogram ensures the expvar Histogram referenced by
+// metricName abides a normal distribution.
+func AssertGraphiteNormalHistogram(t *testing.T, prefix, metricName string, mean, stdev int64, quantiles []int, gPayload string) {
+	const tolerance int = 2
```
peterbourgon: `int` is deduced, so

```go
const tolerance = 2
```

peterbourgon:
Lots of nits and relatively minor structural stuff, but the shape of it is good! Thanks very much!

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,295 @@
...
+
+func (e *emitter) gauge(name string) metrics.Gauge {
+	g := &gauge{name, 0}
+	e.gauges = append(e.gauges, g)
+	return g
+}
+
+// Start will kick off a background goroutine to
+// call Flush once every interval.
+func (e *emitter) Start(interval time.Duration) {
```
ChrisHines: What would happen if `Start` is called more than once? Should there be a `Stop` method?

jprobinson:
I've made all of the suggested changes including an option to use UDP. I've also changed the `now` value slightly when emitting metrics to reflect that it is metric collection time.

metrics/graphite/graphite.go
jprobinson: Good catch. `Start` called twice at this state would cause folks pain. I'll wrap the guts in a `sync.Once`.

`Stop` also sounds like a good idea as we could have it call `Flush` one last time and then close the current connection.

jprobinson: I've added a `sync.Once` to deal with multiple `Start`s, and a `Stop` method to stop the flush loop, flush the current metrics and close the existing connection.

metrics/graphite/graphite.go
```diff
@@ -150,16 +154,39 @@ func (e *emitter) dial() error {
...
+	var err error
+	e.start.Do(func() {
+		err = e.dial()
+		if err != nil {
+			return
 		}
-	}()
-	return nil
+		go func() {
+			t := time.Tick(interval)
```
ChrisHines: The underlying `time.Ticker` is leaked when this goroutine exits. Use `time.Newticker` and add `defer t.Stop` to ensure it isn't leaked.

metrics/graphite/graphite.go
```diff
@@ -150,16 +154,39 @@ func (e *emitter) dial() error {
...
+func (e *emitter) Stop() error {
+	if e.conn == nil {
+		return nil
+	}
+	// stop the ticking flush loop
+	e.stop <- true
+	// get one last flush in
+	e.Flush()
+	// close the connection
+	return e.conn.Close()
```
ChrisHines: This could race with the flush loop. `Stop` sends a stop signal, but doesn't wait on anything to confirm the flush loop has actually exited. `e.conn.Close()` could happen before the flush loop exits, which means the flush loop could call `e.Flush()` after the connection is closed.

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// Stop will flush the current metrics and close the
+// current Graphite connection, if it exists.
+func (e *emitter) Stop() error {
+	if e.conn == nil {
+		return nil
+	}
+	// stop the ticking flush loop
+	e.stop <- true
```
ChrisHines: This will block forever if `Stop` is called more than once.

metrics/graphite/graphite.go
jprobinson: Since `stop` is an unbuffered channel, I figured line 185 would block until the flush loop was able to hit the `stop` select case, meaning nothing else was currently flushing by the time we reached the second `Flush` on line 187, right?

ChrisHines: We only know that the flush loop has received a value after line 185 has completed, we cannot assume that it completed it's call to `e.Flush()`. The local code could get there first, and could also complete `e.conn.Close()` before the flusher goroutine gets a chance to do any of the select case after receiving the value.

buth: The `Stop` method does wait until the flush loop has read from the `stop` channel on line 185. And in that case, there is no possibility that the flush loop will call `e.Flush()` again, even if the loop has not yet exited. I don't see a race here.

metrics/graphite/graphite.go
jprobinson: Roger that. I've added logic to `nil` the connection after Stop. After the first call, this should just immediately return.

metrics/graphite/graphite.go
ChrisHines: Oh yes, I see it now. I was reading the sequence wrong. My appologies.

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,335 @@
...
+		err = e.dial()
+		if err != nil {
+			return
+		}
+		go func() {
+			t := time.Tick(interval)
+			for {
+				select {
+				case <-t:
+					e.Flush()
```
peterbourgon: In the case of TCP, Flush can fail if the network connection fails. We'd then need to initiate a reconnect process, with e.g. exponential backoff.

metrics/graphite/graphite.go
```diff
@@ -0,0 +1,335 @@
...
+
+	logger log.Logger
+}
+
+// NewEmitter will return an Emitter that will prefix all
+// metrics names with the given prefix. Once started, it will attempt to create
+// a TCP or a UDP connection with the given address and periodically post
+// metrics to the connection in the Graphite plaintext protocol.
+// If the provided `tcp` parameter is false, a UDP connection will be used.
+func NewEmitter(addr string, tcp bool, metricsPrefix string, logger log.Logger) *Emitter {
```
peterbourgon: I'd like to see this use the parameters from [net.Dial](https://golang.org/pkg/net/#Dial), namely `network, address string`, and drop `tcp bool`. Then you can have another constructor that parameterizes the dialer, which should really help with testing. 

```go
type Emitter struct {
    network string
    address string
    dial    Dialer
    // ....
}

type Dialer func(network, address string) (net.Conn, error)

func NewEmitter(network, address string, ...) *Emitter {
    return NewEmitterDial(network, address, net.Dial, ...)
}

func NewEmitterDial(network, address string, dial Dialer, ...) *Emitter {
    return &Emitter{
        network: network,
        address: address,
        dial:    dial,
        // ...
    }
}

func (e *Emitter) dial() {
    conn, err := e.dial(e.network, e.address)
    // ...
}
```

What do you think?

jprobinson: I like it. Adding now.

metrics/graphite/graphite.go
peterbourgon: I was thinking last night about this; surely there must be a small package which manages a net.Conn with exponential backoff via some kind of simple state machine. If not, a reasonably generic one could be built.

ChrisHines: Here are a couple options I found:

* https://godoc.org/github.com/bruston/try
* https://godoc.org/github.com/karrick/goperconn

Either could be used directly, or taken as inspiration for a Go kit provided implementation.

jprobinson: I've added some logic for retries with an exponential back off in Flush().

metrics/graphite/graphite.go
```diff
@@ -175,18 +175,48 @@ func (e *Emitter) Stop() error {
...
+	for i := 1; i <= RetryMax; i++ {
+		err = e.flush(e.conn)
+		if err == nil {
+			return nil
+		}
+		e.logger.Log(
+			"err", err,
+			"msg", fmt.Sprintf("unable to flush metrics on attempt %d, waiting %s", i, wait),
+		)
+		time.Sleep(wait)
```
adammck: Minor point, but the last (and longest) sleep will still occur even though we're not going to retry the flush. You might want to special-case that last one to break immediately after the failed flush.

jprobinson: Was thinking about this as well. I'll see if there's something a lil cleaner we can do.

jprobinson: The loop no longer sleeps after the last error 👍

metrics/graphite/graphite.go
peterbourgon: Flush cannot be allowed to block, because it interferes with stopping the emitter. It also probably _shouldn't_ block, as emission is best-effort.

I think the Emitter will have to have some kind of [connection manager state machine like this](https://gist.github.com/peterbourgon/4676d90de3afd25de46768200d009f62). Note I just threw this together, please give it a proper review and test :)

peterbourgon: Actually, if you don't mind, I'll make a PR to your PR?

jprobinson: Yo dawg, sounds good to me!

peterbourgon: @jprobinson https://github.com/jprobinson/kit/pull/1 :inception.jpg:

PR #265, opened 2016-05-16
Implementation of InfluxDB Metrics (resolves #215)

DSchalla:
This is a work in progress PR for the implementation of the InfluxDB metrics package. The first commit contains the implementation for the counter writing to a InfluxDB backend with tag and field support.

**To Do:**
- [x] Counter Support
- [x] Counter Tests
- [x] Gauge Support
- [x] Gauge Tests
- [x] Histogram Support
- [x] Histogram Tests
- [x] Code Cleanup

Feel free to provide feedback regarding structure, tests and quality overall, its highly appreciated!

PR #267, opened 2016-05-17
Remove naming prefixes from metric backends (closes #266)

### LINKED ISSUE #266: metrics: e.g. prometheusCounter should just be counter
The `prometheus` prefix is implicit in the package name. Same for package statsd. [See here](https://github.com/go-kit/kit/pull/265#issuecomment-219779275).

DSchalla:
As discussed in https://github.com/go-kit/kit/pull/265 this PR removes the backend prefixes from the packages, since they are not required.


peterbourgon:
Yay A+ #1! 👍

Issue #246, opened 2016-05-02
log: Provide a simple concurrent safe Logger wrapper.


ChrisHines:
#### Proposal

Add a new exported function to `Go-kit/log`.

```go
// NewMutexLogger returns a logger that uses a sync.Mutex to synchronize
// concurrent use of the wrapped logger.
func NewMutexLogger(logger Logger) Logger
```

#### Motivation

Go-kit/log should not require external code to use safely in a concurrent manner.

#### Status Quo

Go-kit/log carefully documents that `Logger` implementations must be safe for concurrent use. `NewLogfmtLogger` and `NewJSONLogger` document that they require an `io.Writer` safe for concurrent use *if the logger will be used concurrently*. That is to say that the Logfmt and JSON loggers are safe, but their side effect of writing data may not be safe. Unfortunately many implementations of `io.Writer` are not safe for concurrent use.

#### Alternatives

* A mutex based concurrent safe `io.Writer`, or a
* A channel based asynchronous Logger.

A MutexWriter allows the log event formatting to occur concurrently outside the scope of the mutex. A channel based AsyncLogger offloads log event formatting to a single background goroutine and also guarantees concurrent safe writing of the formatted data without additional locking, but it also complicates error handling and application shutdown.

peterbourgon:
Strong +1. This should be not much code, do you think?

ChrisHines:
The code is trivial. I am interested in opinions on the alternatives as well. I see advantages to each, and I have used an asynchronous Logger in production with log15 in the past. I can imagine a future where Go-kit/log included all three varieties, but I am concerned that doing so might introduce the paradox of choice. Thoughts? I can elaborate on the alternatives if that will help the discussion.

basvanbeek:
I think choice is not necessarily a bad thing if the pros/cons are diligently documented. By elaborating on the alternatives we start that process and might end up with the conclusion that either a single solution is the best or choice is needed. Either way we'll have the arguments for our future decision written down.

ChrisHines:
@peterbourgon & @basvanbeek Thanks for the feedback.

To move this forward I will create a PR that implements all three options. We can review and decide which to keep. It may be a few days before I have the time to put the PR together.

peterbourgon:
I reckon this can be closed?

PR #273, opened 2016-05-21
[WIP] log: Add asynchronous wrappers.

ChrisHines:
Various implementations as outlined in #246.

Please review and comment on new APIs.

### DIFF
--- a/log/sync.go
+++ b/log/sync.go
@@ -85,7 +85,7 @@ func (l *syncLogger) Log(keyvals ...interface{}) error {
 // another logger.
 //
 // Errors returned by the wrapped logger are ignored, therefore the wrapped
-// logger should must handle all errors appropriately.
+// logger must handle all errors appropriately.
 type AsyncLogger struct {
 	logger   Logger
 	keyvalsC chan []interface{}
@@ -132,16 +132,13 @@ func (l *AsyncLogger) Cap() int {
 // NonblockingLogger provides buffered asynchronous and concurrent safe
 // logging for another logger.
 //
-// If the wrapped logger's Log method ever returns an error, the
-// NonblockingLogger will stop processing log events and make the error
-// available via the Err method. Any unprocessed log events in the buffer will
-// be lost.
+// Errors returned by the wrapped logger are ignored, therefore the wrapped
+// logger must handle all errors appropriately.
 type NonblockingLogger struct {
 	logger   Logger
 	keyvalsC chan []interface{}
 
 	mu       sync.Mutex
-	err      error
 	stopping chan struct{} // must be closed before keyvalsC
 
 	stopped chan struct{} // closed when run loop exits
@@ -164,19 +161,15 @@ func NewNonblockingLogger(logger Logger, size int) *NonblockingLogger {
 func (l *NonblockingLogger) run() {
 	defer close(l.stopped)
 	for keyvals := range l.keyvalsC {
-		err := l.logger.Log(keyvals...)
-		if err != nil {
-			l.stop(err)
-			return
-		}
+		l.logger.Log(keyvals...)
 	}
 }
 
-func (l *NonblockingLogger) stop(err error) {
+// Stop stops the NonblockingLogger. After stop returns the logger will not
+// accept new log events. Log events queued prior to calling Stop will be
+// logged.
+func (l *NonblockingLogger) Stop() {
 	l.mu.Lock()
-	if err != nil && l.err == nil {
-		l.err = err
-	}
 	select {
 	case <-l.stopping:
 		// already stopping, do nothing
@@ -216,13 +209,6 @@ var (
 	ErrNonblockingLoggerOverflow = errors.New("aysnc logger: log buffer overflow")
 )
 
-// Stop stops the NonblockingLogger. After stop returns the logger will not
-// accept new log events. Log events queued prior to calling Stop will be
-// logged.
-func (l *NonblockingLogger) Stop() {
-	l.stop(nil)
-}
-
 // Stopping returns a channel that is closed after Stop is called.
 func (l *NonblockingLogger) Stopping() <-chan struct{} {
 	return l.stopping
@@ -232,14 +218,6 @@ func (l *NonblockingLogger) Stopping() <-chan struct{} {
 // events have been sent to the wrapped logger.
 func (l *NonblockingLogger) Stopped() <-chan struct{} {
 	return l.stopped
-}
-
-// Err returns the first error returned by the wrapped logger.
-func (l *NonblockingLogger) Err() error {
-	l.mu.Lock()
-	err := l.err
-	l.mu.Unlock()
-	return err
 }
 
 // Len returns a snapshot of the number of buffered log events. The returned
--- a/log/sync.go
+++ b/log/sync.go
@@ -1,6 +1,7 @@
 package log
 
 import (
+	"errors"
 	"io"
 	"sync"
 	"sync/atomic"
@@ -43,11 +44,9 @@ type SyncWriter struct {
 }
 
 // NewSyncWriter returns a new SyncWriter. The returned writer is safe for
-// concurrent use.
+// concurrent use by multiple goroutines.
 func NewSyncWriter(w io.Writer) *SyncWriter {
-	return &SyncWriter{
-		w: w,
-	}
+	return &SyncWriter{w: w}
 }
 
 // Write writes p to the underlying io.Writer. If another write is already in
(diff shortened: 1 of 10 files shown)


log/sync.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+}
+
+// Log queues keyvals for logging by the wrapped Logger. Log may be called
+// concurrently by multiple goroutines. If the the buffer is full, Log will
+// return ErrAsyncLoggerOverflow and the keyvals are not queued. If the
+// AsyncLogger is stopping, Log will return ErrAsyncLoggerStopping.
+func (l *AsyncLogger) Log(keyvals ...interface{}) error {
+	l.mu.Lock()
+	defer l.mu.Unlock()
```
peterbourgon: This is interesting. At first I thought it was an oversight, but now I see it's to prevent a race. Still, it seems a shame to require a mutex op with every Log; my intuition is that there's a clever way to avoid it...

peterbourgon: Ah, no. If AsyncLogger's contract is that all successful Log invocations make it to the writer, then you do indeed need some explicit synchronization.

peterbourgon:
Very happy with this. My only quasi-suggestion would be to place some kind of pithy comparison of the 3 new types somewhere, for easy review by new readers. Perhaps in the package-level documentation?

As an aside, it's a unique pleasure to read your code. The comments guide me precisely through the logic, and the code feels light to the touch; as much as any code I read these days, like [a quiet conversation](http://quotes.cat-v.org/programming) between programmer and mechanical colleague. Thanks for that.

ChrisHines:
> ... to place some kind of pithy comparison of the 3 new types somewhere, for easy review by new readers.

Indeed. I started on that but wanted to wait until we decided what to keep. That is one of the reasons this PR is marked WIP.

The other reason I consider this still a WIP is that I'd like to take some time to consider the contract of AsyncLogger very carefully.

Although `AsyncLogger` is rather complicated, it is valuable if you don't want latencies in the logging pipeline to impact clients. The cost we pay is the difficulty of signaling the application about errors encountered by the wrapped logger. In addition we introduce a new failure mode, buffer overflow.

##### Error Handling

I am pretty happy with the decision that `AsyncLogger` treats errors from the wrapped logger as fatal. I think this approach is far better than ignoring the errors as done by [log15.BufferedHandler](https://godoc.org/github.com/inconshreveable/log15#BufferedHandler). An application retains the option of introducing a more sophisticated error handling strategy in the wrapped logger so that non-fatal errors do not reach `AsyncLogger`.

`AsyncLogger` will stop due to either an explicit call to `Stop` or because of an error returned from the wrapped logger. Currently `Log` returns `ErrAsyncLoggerStopping` in either situation.

Should `Log` return the error from the wrapped logger if it is not `nil`?

##### Buffer Overflow

Regarding the buffer overflow semantics: I think there are three choices, return an error as this PR does, provide back pressure by blocking, or allow blocking for a max duration and then timeout with an error.

What do you think of the alternatives?
Should this behavior be configurable?

peterbourgon:
At first I struggled to understand the use cases for which AsyncLogger is intended. My intuition is that the strict error semantics and emphasis on throughput mean it is aimed at applications where logging is critical, e.g. event sourcing rather than application logging. Is that right? If so, it might be worth a little usage text.

In that light the current API design, especially making the error available via Err, makes a lot of sense. It apes bufio.Scanner. So,

> Should `Log` return the error from the wrapped logger if it is not nil?

Seems to me the current design is better, as it gives you more information about the state of the logger.

> What do you think of the alternatives? [backpressure-mode, dual-mode]
> Should this behavior be configurable?

Regarding backpressure-mode, I think it is a good idea and (without more context about likely usage) probably a better default than erroring out. But it seems like it would require a separate implementation. That is, it would be strange to have the Err method semantics on an instance that would never return an error; just looking at the type wouldn't be enough to tell you if you had to write different logic in your calling code to manage errors, you'd have to look at the constructor [options]. So I guess it would be better to be a different constructor + logger type altogether? Is that also your thinking?

I don't see _huge_ value in the dual-mode implementation. My intuition is that, practically, it would just delay inevitable problems, and reduce to the error-mode implementation. That is, if clients use it, they will still have to write their code to manage errors.

ChrisHines:
@peterbourgon Thanks for the feedback.

>My intuition is that the strict error semantics and emphasis on throughput mean it is aimed at applications where logging is critical, e.g. event sourcing rather than application logging.

That is partially right. I would say the emphasis is on low latency of the `Log` method rather than throughput. I think the fact that buffer overflow is considered a hard failure and the heavyweight synchronization make it a poor fit for high throughput use cases.

I have used something like `AsyncLogger` for critical application logs when logs were of moderate volume and delays caused by I/O latency were painful. The application was the node agent for a batch processing cluster orchestration system, so it often had noisy neighbors that could be saturating local system I/O and we didn't want the app to block while writing logs. At the same time we wanted logs flushed to disk ASAP because they were sometimes the best source of information to understand what led to a particular issue on a node. We needed strict error and overflow semantics because we preferred that the application abort when it encountered unanticipated errors (a.k.a. [crash-only software](https://lwn.net/Articles/191059/)).

>Regarding backpressure-mode, I think it is a good idea and (without more context about likely usage) probably a better default than erroring out. But it seems like it would require a separate implementation.

I agree but I'll take it a step farther. I think it *must* be a separate implementation. I made an attempt to add it as an option to the existing implementation yesterday and could not figure out how to easily support back-pressure along with the `Stop` method and the ability to drain the buffer without introducing possible deadlocks.

>I don't see huge value in the dual-mode implementation. My intuition is that, practically, it would just delay inevitable problems, and reduce to the error-mode implementation.

I agree with you here as well. I was already leaning toward that assessment, so thanks for confirming.

At this point I think the current semantics of `AsyncLogger` are probably too specialized for inclusion in package `log`. Do you think there is a variant that would be more widely useful to Go kit users, or should we just drop it altogether?

basvanbeek:
It might be too specialized for inclusion in package `log` but it would also be a waste not to have it available to potential users.

Maybe we should have a place where we can store or refer to "plugins" and "extensions" which have a smaller use case / audience.

ChrisHines:
>Maybe we should have a place where we can store or refer to "plugins" and "extensions" which have a smaller use case / audience.

Maybe, but log15 has an ext package along those lines and the choice of what went into that package and the main package just feels arbitrary, IMO. I would rather only create packages that have a well defined focus, or serve to isolate external dependencies.

So, my vote is that if we add an `AsyncLogger` I think it should be useful enough that we feel comfortable putting in package log next to `SyncLogger`.

ChrisHines:
I have renamed the original `AsyncLogger` to `NonblockingLogger` which I think clarifies its purpose quite a bit. I have add a new simpler `AsyncLogger` implementation that blocks when the buffer is full and ignores errors from the wrapped logger.

The next step is to consider changing `NonblockingLogger` to have the same error handling semantics as the new `AsyncLogger` with the hope that it simplifies the code.

ChrisHines:
I think it makes sense to force the wrapped logger to handle errors for both `AsyncLogger` and `NonblockingLogger` so I've removed the error trapping logic from the latter.

The remaining question is whether the ability to stop these loggers and wait for the buffer to drain is worthwhile. It has been important to me in the past to make a best effort that all log events were saved during a graceful shutdown. Note: A graceful shutdown may occur due to an unrecoverable error.

@peterbourgon @basvanbeek PTAL.

log/sync.go
```diff
@@ -0,0 +1,233 @@
...
+	return s.Log(keyvals...)
+}
+
+// Swap replaces the currently wrapped logger with logger. Swap may be called
+// concurrently with calls to Log from other goroutines.
+func (l *SwapLogger) Swap(logger Logger) {
+	l.logger.Store(loggerStruct{logger})
+}
+
+// SyncWriter synchronizes concurrent writes to an io.Writer.
```
peterbourgon: In what circumstances would you want to

```go
logger := log.NewLogfmtLogger(log.NewSyncWriter(w))
```

instead of

```go
logger := log.NewSyncLogger(log.NewLogfmtLogger(w))
```

? Could a little bit of guidance make its way into the comments on the types?

log/sync.go
```diff
@@ -0,0 +1,233 @@
...
+// Cap returns the maximum capacity of the buffer.
+func (l *AsyncLogger) Cap() int {
+	return cap(l.keyvalsC)
+}
+
+// NonblockingLogger provides buffered asynchronous and concurrent safe
+// logging for another logger.
+//
+// Errors returned by the wrapped logger are ignored, therefore the wrapped
+// logger must handle all errors appropriately.
```
peterbourgon: I don't have a great intuitive understanding of the circumstances where I would use these loggers. But it seems to me that more sophisticated operational requirements (async, nonblocking...) would go hand-in-hand with more sophisticated error management requirements. And that simply squelching the error seems heavy-handed, at first glance.

At the moment, if I want to respond to a non-nil error returned by a Log invocation, it seems I'll need to wire up that control flow hierarchy "underneath" {Nonblocking,Async}Logger, probably with my own e.g. ErrorSiphoningLogger — is that correct?

log/sync.go
ChrisHines: I plan to add docs like that before removing the WIP tag. Generally speaking you want to do the least amount of work you can within a mutex while making sure that each log event is handled completely while the lock is held. So the first form is preferred in this case because `NewLogfmtLogger` makes a single call to `Write` for each log event.

The second form is needed for loggers that perform multiple writes per log event. `NewJSONLogger` currently does that, and external implementations of `Logger` could as well.

So, crafting the above into some nice docs is still on the todo list for this PR.

log/sync.go
ChrisHines: Your first two paragraphs are spot on.

What value does having a `StopOnError` option provide? In what scenario would anyone enable that option and not want to also handle the error?

log/sync.go
peterbourgon: Gotcha.

log/sync.go
peterbourgon: The value is making your logging pipeline deterministic, in some sense. But playing it out a little further, I think you're right: simply stopping on error is a halfway solution.

I guess if you opt-in to async logging, you're on-board with CSP, and you're able to appropriately deal with something like a `func (logger *AsyncLogger) Errors() <-chan error`. Is that where you're driving the API, do you think?

ChrisHines:
Status update: #327 extracted SyncLogger and SyncWriter into a separate contribution. This PR remains as a place to discuss the merits and design of AsyncLogger and NonblockingLogger until we are happy with them or decide to abandon them.

peterbourgon:
So I'm walking through the outstanding PRs with an eye to either merging or abandoning, and I'm curious what the state of this one could be. I'm completely happy with the two implementations here, modulo final tweaks and usage documentation, but I'm also alright with abandoning them. WDYT?

ChrisHines:
I think the SyncLogger and SyncWriter we've already added to the package are sufficient and the additional channel based approaches here are not worth the trouble in this package. My vote is to abandon them.

peterbourgon:
Great. Thanks for the [historical] work!

PR #276, opened 2016-05-25
package sd (take 2)

peterbourgon:
Iteration on #235.

Package loadbalancer is deleted, replaced by package sd. What was previously loadbalancer.Publisher becomes sd.Subscriber, which I believe better reflects its purpose. That is, the sd/consul.Subscriber is subscribing to updates from Consul.

```go
// Subscriber listens to a service discovery system and yields a set of
// identical endpoints on demand. An error indicates a problem with connectivity
// to the service discovery system, or within the system itself; a subscriber
// may yield no endpoints without error.
type Subscriber interface {
	Endpoints() ([]endpoint.Endpoint, error)
}
```

Additionally, we introduce a new type, sd.Registrar, responsible for advertising the lifecycle of the Go kit service into the corresponding service discovery system. That is, sd/consul.Registrar registers instances to Consul. 

```go
// Registrar registers instance information to a service discovery system when
// an instance becomes alive and healthy, and deregisters that information when
// the service becomes unhealthy or goes away.
//
// Registrar implementations exist for various service discovery systems. Note
// that identifying instance information (e.g. host:port) must be given via the
// concrete constructor; this interface merely signals lifecycle changes.
type Registrar interface {
	Register()
	Deregister()
}
```

This is optional, as many users prefer to or must have their orchestration system manage entries in their service discovery system. But for those users who do self-registration, this will be useful. Addresses #167.

A few additional changes:

- The Consul implementation was refactored, and sd/consul.Publisher added
- The endpoint cache was simplified and moved to sd/cache
- Load balancers and helpers are moved to package sd/lb

The checklist before this PR is mergeable:

- [x] Refactor sd.Subscriber to yield individual endpoint.Endpoints
- [x] Port sd.Registrar + implementations 
- [x] Use new endpoint cache, Consul subscriber implementation, and other mechanical bits
- [x] Refactor addsvc example
- [x] Refactor profilesvc example, adding a client package
- [x] Refactor apigateway example
- [x] Make minimum changes necessary to shipping example
- [x] Update examples/README + refactor stringsvc examples
- [x] Delete package loadbalancer

Once this is merged I will file issues for:

- Add sd/etcd.Publisher
- Add sd/zk.Publisher


examples/addsvc2/service.go
```diff
@@ -0,0 +1,115 @@
...
+func NewBasicService() Service {
+	return basicService{}
+}
+
+type basicService struct{}
+
+// Sum implements Service.
+func (s basicService) Sum(_ context.Context, a, b int) (int, error) {
+	if a == 0 && b == 0 {
+		return 0, ErrTwoZeroes
```
basvanbeek: maybe we should test for a "real" error (overflow) to signal people that even the simplest method probably has real exceptions / error conditions so they don't think to lightly about proper error handling for every method

examples/addsvc2/service.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return a + b, nil
+}
+
+// Concat implements Service.
+func (s basicService) Concat(_ context.Context, a, b string) (string, error) {
+	return a + b, nil
```
basvanbeek: also here we could show sensibility by implementing a "size out of bounds" kind of error... not too much code to distract but enough to again highlight the importance of sanitizing and checking

basvanbeek:
something quick I came up with for the above mentioned comments...
```go
const (
	intMax = 1<<31 - 1
	intMin = -(intMax + 1)
	maxLen = 102400
)

// Service Method Errors
var (
	ErrIntOverflow     = errors.New("integer overflow")
	ErrMaxSizeExceeded = errors.New("result exceeds maximum size")
)

// Sum implements Service.
func (s basicService) Sum(_ context.Context, a, b int) (int, error) {
	if (b > 0 && a > (intMax-b)) || (b < 0 && a < (intMin-b)) {
		return 0, ErrIntOverflow
	}
	return a + b, nil
}

// Concat implements Service.
func (s basicService) Concat(_ context.Context, a, b string) (string, error) {
	if len(a)+len(b) > maxLen {
		return "", ErrMaxSizeExceeded
	}
	return a + b, nil
}
```

examples/addsvc/client/grpc/client.go
```diff
@@ -0,0 +1,75 @@
...
+			addsvc.DecodeGRPCSumResponse,
+			pb.SumReply{},
+			grpctransport.SetClientBefore(opentracing.FromGRPCRequest(tracer, "Sum", logger)),
+		).Endpoint()
+		sumEndpoint = opentracing.TraceClient(tracer, "Sum")(sumEndpoint)
+		sumEndpoint = limiter(sumEndpoint)
+		sumEndpoint = circuitbreaker.Gobreaker(gobreaker.NewCircuitBreaker(gobreaker.Settings{
+			Name:    "Sum",
+			Timeout: 30 * time.Second,
+		}))(sumEndpoint)
```
ChrisHines: Can you use `endpoint.Chain` to compose this sequence?

peterbourgon: Absolutely, but I find this way — working from the inside out, with multiple expressions — more intuitive. (I added endpoint.Chain by request, at some point.)

ChrisHines: I also find using multiple expressions more intuitive. I argued for simplifying our APIs in that direction way back in ancient history (https://github.com/go-kit/kit/pull/14#issuecomment-86338127). I wonder if the middleware abstraction has really proved worth it if the only real use of it (`endpoint.Chain`) is avoided because it is less intuitive.

*For reference: `endpoint.Chain` was designed in the discussion on #95 and implemented in it's current form in #96.*

peterbourgon:
This is now ready for review, if anyone would care to give it a look. Refactoring addsvc and profilesvc took much longer than anticipated; the upside is I'm _very_ happy with the state of those examples, now, and I think they will be very educating, especially to new users.

PR #278, opened 2016-05-27
shipping: Add prometheus instrumentation

marcusolsson:


examples/shipping/booking/instrumenting.go
```diff
@@ -0,0 +1,114 @@
...
+		Name:      "request_latency_microseconds",
+		Help:      "Total duration of requests in microseconds.",
+	}, fieldKeys))
+
+	return &instrumentingService{
+		requestCount:   requestCount,
+		requestLatency: requestLatency,
+		Service:        s,
+	}
+}
```
peterbourgon: Hmm. I tend to take the metrics as dependencies, rather than constructing them in the middleware, as you sometimes need to manage the registration (via the Prometheus HTTP handler) in unique ways. But I reckon that's a matter of opinion.

marcusolsson: I actually started off like that, but with multiple services, each with
multiple metrics, I felt my main.go got a little out of hand. I don't feel
too strongly about it though and I wouldn't have any problems with changing
it back. Didn't think of the registration management though, that's a good
point.

Den fre 27 maj 2016 16:59Peter Bourgon <notifications@github.com> skrev:

peterbourgon: Yeah, tradeoffs! Let's roll with this one for now :)

Issue #281, opened 2016-05-31
grpc.NewClient hard-codes proto buf package to "pb"


waigani:
https://github.com/go-kit/kit/blob/master/transport/grpc/client.go#L40 

This tripped me up when testing with https://github.com/grpc/grpc-go/tree/master/examples/helloworld which names the package "helloworld". This will be a common entry point for new users and, more to the point, hard-coding to "pb" prohibits using any other package name.

Either serviceName needs to include the package ("pb.Add") or NewClient needs to take a package parameter. Happy to make a PR for the one that makes most sense.

basvanbeek:
backward compatible fix has been merged into master

PR #282, opened 2016-06-01
gRPC client fix: allow package names other than "pb"

basvanbeek:
Previously the gRPC Client package automatically prefixed the "pb" package namespace to the gRPC service name. Since generated gRPC packages can have other names this caused problems.

This fix allows for setting service name with package prefix explicitly while remaining backward compatible to consumers omitting the package prefix.

This resolves issue #281


peterbourgon:
LGTM!

PR #289, opened 2016-06-07
After client option

geoffberger:
This work allows access to `http.Response` and update `context.Context` once a request has been made with `httptransport.NewClient`. Here's an example:

```go
u, _ := url.Parse("http://example.com")
afterFunc := func(ctx context.Context, r *http.Response) context.Context {
  return ctx
}
client := httptransport.NewClient(
  "GET",
  u,
  encodeFunc,
  decodeFunc,
  httptransport.SetClientAfter(afterFunc),
)
```

Let me know if any improvements can be made, i.e. code improvements, issues, naming, and so forth.


transport/http/client.go
```diff
@@ -47,6 +49,13 @@ func NewClient(
 // ClientOption sets an optional parameter for clients.
 type ClientOption func(*Client)
 
+// ClientResponseFunc takes information from an HTTP request to read from the
+// response and do something additional with the response or add/update
+// context. ClientResponseFuncs are only executed on the client immediately
+// after the request has been made but prior to decoding. This allows for
+// intervention in case there is an error from the request.
+type ClientResponseFunc func(context.Context, *http.Response) context.Context
+
```
peterbourgon: Please move this definition to request_response_funcs.go, rename the existing ResponseFunc to ServerResponseFunc, and make the comment blocks on {Client,Server}ResponseFunc roughly symmetrical :)

transport/http/client.go
```diff
@@ -85,6 +101,10 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		}
 
 		resp, err := ctxhttp.Do(ctx, c.client, req)
+		for _, f := range c.after {
+			ctx = f(ctx, resp)
+		}
+
```
peterbourgon: Why is this here and not after the err check?

transport/http/client.go
```diff
@@ -85,6 +101,10 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		}
 
 		resp, err := ctxhttp.Do(ctx, c.client, req)
+		for _, f := range c.after {
+			ctx = f(ctx, resp)
```
peterbourgon: After looking at this and comparing to Server.ServeHTTP, I'd like to see more symmetry — propose that ResponseFunc shall also return a context. Would you be willing to make that change in this PR too?

transport/http/client.go
```diff
@@ -59,6 +68,13 @@ func SetClientBefore(before ...RequestFunc) ClientOption {
 	return func(c *Client) { c.before = before }
 }
 
+// SetClientAfter sets the ClientResponseFunc applied to the incoming HTTP
```
peterbourgon: ...sets the ClientResponseFunc**s**

transport/http/client_test.go
```diff
@@ -62,6 +71,11 @@ func TestHTTPClient(t *testing.T) {
 		t.Errorf("want %q, have %q", want, have)
 	}
 
+	// Check that Response header set from server was received in SetClientAfter
+	if want, have := afterVal, afterHeaderVal; want != have {
```
peterbourgon: This is racy, isn't it?

transport/http/client.go
geoffberger: Nice, good call. This has been updated.

transport/http/client.go
geoffberger: So the reason to have this immediately after the request verses after the err check was to make sure any part of the `http.Response` can be captured, even if an error occurs. To your point above though, for consistency with `Server.ServeHTTP`, it might make sense to have it be mirrored. I'm open either way, what do you think?

transport/http/client.go
geoffberger: Awesome, good call. I've updated `http.ServerResponseFunc` to return a context.

transport/http/client.go
geoffberger: all set.

transport/http/client.go
geoffberger: My bad, you are correct, this should be after the error check. I looked into `ctxhttp.Do` and since no response is being passed back, this doesn't make sense. This now appears after the error, but before decoding, which is nice because it more closely mirrors `Server.ServeHTTP`. Comments have also been updated to reflect this change. Thanks!

transport/http/client.go
```diff
@@ -59,6 +61,13 @@ func SetClientBefore(before ...RequestFunc) ClientOption {
 	return func(c *Client) { c.before = before }
 }
 
+// SetClientAfter sets the ClientResponseFuncs applied to the incoming HTTP
+// request prior to it being decoded. This is useful for obtaining anything off
+// of the response and adding onto the context prior to decoding.
+func SetClientAfter(after ...ClientResponseFunc) ClientOption {
```
peterbourgon: Sorry for the nit :) just noticing this now. In transport/http.Server, none of the functional options begin with **Set**, but in Client, they all do. I guess this is an unintended consequence of the first functional option on the client being SetClient (to set the HTTP client). Can I ask you to go above-and-beyond one last time and make the following renames in http/transport.Client?

- SetClient → SetClient (no change)
- SetClientBefore → ClientBefore
- SetClientAfter → ClientAfter
- SetBufferedStream → BufferedStream

geoffberger: Yeah, definitely. This is all set. I also renamed SetClientBefore in transport/grpc.Client to match what is in transport/grpc.Client. If it's not a problem, at a later point in time I'll update one of the examples (probably addsvc) to use Client.ClientAfter, but let me know if that's something that should be part of this PR.

peterbourgon: No this is fine. Aces!

peterbourgon:
🎉 

Amaze. Very thank.

PR #291, opened 2016-06-07
lint

peterbourgon:
Eventually I would like to get this repo linting cleanly. For now I will just place the script here.


peterbourgon:
For the record the current lint settings aren't gospel, just a good starting point.

PR #300, opened 2016-06-19
metrics: clarify examples/documentation

fortytw2:
Makes the use of the metrics interfaces much clearer in the README examples - instead of just looking like normal use of the statsd/prometheus/expvar libraries, which can be confusing at first glance.

I also think the README information for `metrics` should also be duplicated in the doc comment, not 100% sure though (does godoc render doc comments as markdown?).


peterbourgon:
Cool, thanks for this. I get the gist of what you want here, and I agree, but I think I'd prefer a slightly different angle on the changes. Stay tuned, I'll make a separate PR.

peterbourgon:
This was done in the big metrics refactor. I took your examples almost verbatim. Thank you!

PR #309, opened 2016-07-03
Removed url.QueryEscape from the make*Endpoint functions

LennartOlsen:
Simple example of the issue here:
https://play.golang.org/p/irnijFBj60

### DIFF
--- a/examples/profilesvc/transport.go
+++ b/examples/profilesvc/transport.go
@@ -12,8 +12,6 @@ import (
 	"github.com/gorilla/mux"
 	"golang.org/x/net/context"
 
-	"net/url"
-
 	"github.com/go-kit/kit/log"
 	httptransport "github.com/go-kit/kit/transport/http"
 )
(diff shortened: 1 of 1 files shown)


examples/profilesvc/transport.go
```diff
@@ -227,63 +225,63 @@ func decodeDeleteAddressRequest(_ context.Context, r *http.Request) (request int
...
-	req.Method, req.URL.Path = "POST", url.QueryEscape("/profiles/")
+	req.Method, req.URL.Path = "POST", "/profiles/"
 	return encodeRequest(ctx, req, request)
 }
 
 func encodeGetProfileRequest(ctx context.Context, req *http.Request, request interface{}) error {
 	// r.Methods("GET").Path("/profiles/{id}")
 	r := request.(getProfileRequest)
-	req.Method, req.URL.Path = "GET", url.QueryEscape("/profiles/"+r.ID)
+	req.Method, req.URL.Path = "GET", "/profiles/"+r.ID
```
peterbourgon: Do we still want to QueryEscape the e.g. ID?

examples/profilesvc/transport.go
```diff
@@ -227,63 +225,63 @@ func decodeDeleteAddressRequest(_ context.Context, r *http.Request) (request int
...
-	req.Method, req.URL.Path = "GET", url.QueryEscape("/profiles/"+r.ProfileID+"/addresses/")
+	req.Method, req.URL.Path = "GET", "/profiles/"+r.ProfileID+"/addresses/"
 	return encodeRequest(ctx, req, request)
 }
 
 func encodeGetAddressRequest(ctx context.Context, req *http.Request, request interface{}) error {
 	// r.Methods("GET").Path("/profiles/{id}/addresses/{addressID}")
 	r := request.(getAddressRequest)
-	req.Method, req.URL.Path = "GET", url.QueryEscape("/profiles/"+r.ProfileID+"/addresses/"+r.AddressID)
+	req.Method, req.URL.Path = "GET", "/profiles/"+r.ProfileID+"/addresses/"+r.AddressID
```
peterbourgon: Likewise the two IDs here? (Other comments elided, but you get the same issue with other methods.)

LennartOlsen: You are right - but if we are escaping ids here, we should also be escaping them when saving.
I get that its user input and that should generally be escaped.

IMHO i wouldn't escape the path itself, only the query string eg. whats after in the url "?"

peterbourgon: For example if an ID is `"foo bar"` then it should be stored with a space literal but in the URL as `foo%20bar`.

LennartOlsen: You are right @peterbourgon ill do a new commit

peterbourgon:
Thank you!

PR #311, opened 2016-07-06
OpenTracing breaking API change: SpanContext

basvanbeek:
OpenTracing changed its API and deprecated the Inject method in favor of the new Extract method.

@bensigelman: can you comment if this conforms to an idiomatic OpenTracing solution

### DIFF
--- a/tracing/opentracing/grpc.go
+++ b/tracing/opentracing/grpc.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/metadata"
 
@@ -20,7 +21,7 @@ func ToGRPCRequest(tracer opentracing.Tracer, logger log.Logger) func(ctx contex
 	return func(ctx context.Context, md *metadata.MD) context.Context {
 		if span := opentracing.SpanFromContext(ctx); span != nil {
 			// There's nothing we can do with an error here.
-			if err := tracer.Inject(span, opentracing.TextMap, metadataReaderWriter{md}); err != nil {
+			if err := tracer.Inject(span.Context(), opentracing.TextMap, metadataReaderWriter{md}); err != nil {
 				logger.Log("err", err)
 			}
 		}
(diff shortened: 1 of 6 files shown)


tracing/opentracing/grpc.go
```diff
@@ -37,12 +38,15 @@ func ToGRPCRequest(tracer opentracing.Tracer, logger log.Logger) func(ctx contex
...
-		span, err := tracer.Join(operationName, opentracing.TextMap, metadataReaderWriter{md})
+		var span opentracing.Span
+		wireContext, err := tracer.Extract(opentracing.TextMap, metadataReaderWriter{md})
 		if err != nil {
+			logger.Log("err", err)
+		}
+		if wireContext != nil {
+			span = tracer.StartSpan(operationName, ext.RPCServerOption(wireContext))
+		} else {
 			span = tracer.StartSpan(operationName)
```
bensigelman: I guess (?) we could include the rpc server tag here even if there's no client SpanContext to reference...

tracing/opentracing/http.go
```diff
@@ -57,16 +57,18 @@ func ToHTTPRequest(tracer opentracing.Tracer, logger log.Logger) kithttp.Request
...
 			opentracing.TextMap,
 			opentracing.HTTPHeaderTextMapCarrier(req.Header),
 		)
 		if err != nil {
+			logger.Log("err", err)
+		}
+		if wireContext != nil {
+			span = tracer.StartSpan(operationName, ext.RPCServerOption(wireContext))
+		} else {
 			span = tracer.StartSpan(operationName)
```
bensigelman: (same)

bensigelman:
LGTM, thanks!

basvanbeek:
@bensigelman it looked very consumer unfriendly to add this:
```go
opentracing.Tags{string(ext.SpanKind): ext.SpanKindRPCServer}
```

if we update the Apply function of RPCServerOption to this:

```go
func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
	if r.clientContext != nil {
		opentracing.ChildOf(r.clientContext).Apply(o)
	}
	(opentracing.Tags{string(SpanKind): SpanKindRPCServer}).Apply(o)
}
```

we can change from this in Go kit and other consumers:
```go
if wireContext != nil {
	span = tracer.StartSpan(operationName, ext.RPCServerOption(wireContext))
} else {
	span = tracer.StartSpan(operationName, opentracing.Tags{string(ext.SpanKind): ext.SpanKindRPCServer})
}
```

to simply this:
```go
span = tracer.StartSpan(operationName, ext.RPCServerOption(wireContext))
```

I think it is a very common scenario for servers especially at the edge (API gateways and such) to start the trace with a RPC Server span without an upstream Client as the client often is not instrumented.

bensigelman:
@basvanbeek I think that makes sense, yeah. Let's do it.

peterbourgon:
We are still waiting for changes to be merged in openzipkin/zipkin-go-opentracing, is that right?

basvanbeek:
correct... that will be ready today

peterbourgon:
LGTM! You can merge, right?

PR #314, opened 2016-07-08
Fix: TraceClient crashed in absense of a parentSpan in Context

basvanbeek:
yeah that's unfortunate... we used to not care if parentSpan existed or not but since we now dereference the SpanContext we must test for it...


peterbourgon:
Is it possible to write a previously-failing, now-succeeding test for this?

basvanbeek:
This was a dumb oversight on my end while refactoring from old to new API.

The old api allowed you to drop in a span itself which was allowed to be nil. I now have to actually dereference the span (so it must exists) to get to its Context property. 

Not sure what an extra test would help with

basvanbeek:
Added test to safeguard from regressions...

peterbourgon:
Confirmed test fails on current master. LGTM!

PR #316, opened 2016-07-09
do not log "SpanContext not found in Extract carrier" 

basvanbeek:
If a server endpoint includes OpenTracing middleware and the client calling the endpoint is not instrumented, OpenTracing will emit a "SpanContext not found" error.  We should not log these as errors as it is common behavior.

Also removed the incorrect comments on logger being allowed to be nil.


peterbourgon:
LGTM

PR #317, opened 2016-07-10
Adding testable example documention

xmattstrongx:
Adding generic example as documentation for how to use this package.

### DIFF
--- a/sd/etcd/example_test.go
+++ b/sd/etcd/example_test.go
@@ -1,0 +1,69 @@
+package etcd
+
+import (
+	"fmt"
+	"time"
+	"io"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/endpoint"
+	"golang.org/x/net/context"
+)
+
+// Package sd/etcd provides a wrapper around the coroes/etcd key value store (https://github.com/coreos/etcd)
+// This example assumes the user has an instance of etcd installed and running locally on port 2379
+func Example() {
+
+	var (
+		prefix   = "/services/foosvc/" // known at compile time
+		instance = "1.2.3.4:8080"      // taken from runtime or platform, somehow
+		key      = prefix + instance
+		value    = "http://" + instance // based on our transport
+	)
+
+	client, err := NewClient(context.Background(), []string{"http://:2379"}, ClientOptions{
+		DialTimeout:             2 * time.Second,
+		DialKeepAlive:           2 * time.Second,
+		HeaderTimeoutPerRequest: 2 * time.Second,
+	})
+
+	// Instantiate new instance of *Registrar passing in test data
+	registrar := NewRegistrar(client, Service{
+		Key:   key,
+		Value: value,
+	}, log.NewNopLogger())
+	// Register new test data to etcd
+	registrar.Register()
+
+	//Retrieve entries from etcd
+	_, err = client.GetEntries(key)
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	factory := func(string) (endpoint.Endpoint, io.Closer, error) {
+		return endpoint.Nop, nil, nil
+	}
+	subscriber, _ := NewSubscriber(client, prefix, factory, log.NewNopLogger())
+
+	endpoints, err := subscriber.Endpoints()
+	if err != nil {
+		fmt.Printf("err: %v", err)
+	}
+	fmt.Println(len(endpoints)) // hopefully 1
+
+	// Deregister first instance of test data
+	registrar.Deregister()
+
+	endpoints, err = subscriber.Endpoints()
+	if err != nil {
+		fmt.Printf("err: %v", err)
+	}
+	fmt.Println(len(endpoints)) // hopefully 0
+
+	// Verify test data no longer exists in etcd
+	_, err = client.GetEntries(key)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
--- a/sd/etcd/client.go
+++ b/sd/etcd/client.go
@@ -44,7 +44,7 @@ type ClientOptions struct {
 	Key                     string
 	CaCert                  string
 	DialTimeout             time.Duration
-	DialKeepAline           time.Duration
+	DialKeepAlive           time.Duration
 	HeaderTimeoutPerRequest time.Duration
 }
 
@@ -83,7 +83,7 @@ func NewClient(ctx context.Context, machines []string, options ClientOptions) (C
 			Dial: func(network, addr string) (net.Conn, error) {
 				dial := &net.Dialer{
 					Timeout:   options.DialTimeout,
-					KeepAlive: options.DialKeepAline,
+					KeepAlive: options.DialKeepAlive,
 				}
 				return dial.Dial(network, addr)
 			},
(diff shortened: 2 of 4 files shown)


sd/etcd/example_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Package sd/etcd provides a wrapper around the coroes/etcd key value store (https://github.com/coreos/etcd) 
+// This example assumes the user has an instance of etcd installed and running locally on port 2379
+func Example() {
+
+	var logger log.Logger
+	{
+		logger = log.NewLogfmtLogger(os.Stdout)
+		logger = log.NewContext(logger).With("ts", log.DefaultTimestampUTC, "caller", log.DefaultCaller)
+	}
```
peterbourgon: The purpose of a test code is to show users how to wire things up. We don't need to be logging anything. Let's drop this, and all the debug logging below.

sd/etcd/example_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Creating random number too append to assure differentiation of test data keys 
+	rand.Seed(int64(time.Now().Nanosecond()))
+
+	value := rand.Int()
+	testData := fmt.Sprintf("foo%d", value)
+
+	value2 := rand.Int()
+	testData2 := fmt.Sprintf("foo%d", value2)
```
peterbourgon:
```go
rand.Seed(time.Now().Nanoseconds())
data1 := fmt.Sprintf("foo%d", rant.Int())
data2 := fmt.Sprintf("foo%d", rand.Int())
```

sd/etcd/example_test.go
```diff
@@ -0,0 +1,99 @@
...
+	kitClientOptions := ClientOptions{
+		Cert:                    "",
+		Key:                     "",
+		CaCert:                  "",
+		DialTimeout:             (2 * time.Second),
+		DialKeepAline:           (2 * time.Second),
+		HeaderTimeoutPerRequest: (2 * time.Second)}
+
+	// Instantiate NewClient on localhost port 2379. Change this to whichever port you host etcd from 
+	kitClient, err := NewClient(context.Background(), []string{"http://:2379"}, kitClientOptions)
```
peterbourgon: There is no regular `clientOptions` or `client`, so we can use shorter names and construction forms here. Also, if Cert/Key/CaCert are optional, better to elide them altogether than set them to empty strings. Also, `DialKeepAline` typo?

```go
client, err := NewClient(context.Background(), []string{"http://:2379"}, ClientOptions{
    DialTimeout:             2 * time.Second,
    DialKeepAlive:           2 * time.Second,
    HeaderTimeoutPerRequest: 2 * time.Second,
})
```

sd/etcd/example_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// Set up first instance of test data
+	data1 := &Service{
+		Key:   testData,
+		Value: "bar1",
+	}
+
+	// Instantiate new instance of *Registrar passing in test data
+	registrar := NewRegistrar(kitClient, *data1, log.NewNopLogger())
+	// Register new test data to etcd
+	registrar.Register()
```
peterbourgon:
```go
service1 := Service{
    Key:   "instance1",
    Value: "10.0.0.1:20001",
}
registrar := NewRegistrar(client, service1, log.NewNopLogger())
registrar.Register()
```

sd/etcd/example_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Deregister second instance of test data
+	registrar2.Deregister()
+
+	// Verify test data no longer exists in etcd
+	_, err = kitClient.GetEntries(testData2)
+	if err != nil {
+		logger.Log("err", err)
+	}
```
peterbourgon: (Comment pending discussion)

xmattstrongx: Sure. We can discuss here or slack. Whichever you prefer

xmattstrongx:
Any other thoughts/comments on this? 

I am thinking/wondering if there might be value in adding some subscriber examples to this before merging.

peterbourgon:
So the idea with this example is to demonstrate how to use sd/etcd.Registrar, but as you note it's meant to be used together with sd/etcd.Subscriber. So, let's say that we have foosvc and barsvc, and barsvc depends on foosvc. Then barsvc will use a Subscriber on a key prefix like "/services/foosvc", and foosvc will register itself with keys like "/services/foosvc/1.2.3.4:8080".

This informs the example. The Registrar should include that prefix, something like

```go
var (
	prefix   = "/services/foosvc/"  // known at compile time
	instance = "1.2.3.4:8080"       // taken from runtime or platform, somehow
	endpoint = "http://" + instance // known at compile time, based on our transport
)

registrar := NewRegistrar(client, Service{
	Key:   prefix+instance,
	Value: endpoint,
}, log.NewNopLogger())

registrar.Register()
```

That's the write side. As you've done, I agree it's a good idea to use the client directly to demonstrate how things look in etcd,

```go
entries, err := client.GetEntries(prefix+instance)
fmt.Printf("%q (%v)\n", strings.Join(entries), err)
```

But we also have the opportunity to provide an end-to-end demonstration by constructing a Subscriber and seeing how many instances we get.

```go
subscriber := NewSubscriber(client, "/services/barsvc", someFactory, log.NewNopLogger())
fmt.Println(len(subscriber.Endpoints()) // hopefully 1
```

And now we can deregister and perform everything again, in reverse.

```go
registrar.Deregister()

fmt.Println(len(subscriber.Endpoints()) // hopefully 0 now

entries, err = client.GetEntries(prefix+instance)
fmt.Printf("%q (%v)\n", strings.Join(entries, ", "), err)
```

I think registering two distinct IPs is a bit of a distraction, and maybe a little confusing, because no single service would ever do that.

Does this make sense? What do you think?

xmattstrongx:
I do agree with your observations. I think adding the subscriber example provides definite value to the code base and Ill be implementing that soon/

However when registering two IPs I was thinking of more of a distributed system where multiple instances of a single service existed. This could all be masked as a single IP by a vip or load balancer though so perhaps the multi IP style example could easily be misconstrued.

xmattstrongx:
Strangely I am not getting the correct zero output when checkout len(subscriber.Endpoints) after `Deregister`ing my instance (from code below). 

```go
registrar.Deregister()
fmt.Println(len(subscriber.Endpoints()) // hopefully 0 now
```


Although when I `Deregister` I am correctly seeing no value for given key when called `GetEntries`. I am thinking my fake factory implementation is incorrect.

xmattstrongx:
Well the code is complete as per the PR comments above. I am still not quite happy with it as I was unable to verify the subscriber functionality with my own client separate from this tests.

I do believe the example code to be accurate and my issues to be stemmed from my own errors when trying to get the factory in NewSubscriber implemented correctly.

peterbourgon:
I can reproduce your error on an actual etcd server, which leads me to believe the DeleteOptions are somehow incorrect. I'm working on this in a separate branch.

peterbourgon:
Nope, it's a bug in Subscriber GetEntries. Fixing.

xmattstrongx:
Ok cool. Ill be on the lookout for the PR. Glad this exercise was able to sniff out a bug.

peterbourgon:
Ha, it was a weird race condition. FML. PR incoming.

Issue #269, opened 2016-05-19
kit/log/levels: support filtering


AlmogBaku:
### Intro

I believe for some scenarios it's a good practice to "throw" a lot of logs(i.e. debugging/info), and in case of need to "expose them" in order to track of bugs.

However, you doesn't necessary want to send all those bugs to the "external" logs management(i.e. debug a single connection details, especially for real-time use cases). Mechanism of filtering by log-levels can be very useful for this scenarios.
### Suggestion

Adding "just a layer of filtering" to the current structure is feasible, however- it will be based on string comparisons instead of mathematical comparisons (from DEBUG logs and above)

In order to achieve a verbosity level, I suggest to use a new type `type LogLevel int`, and to compare base on it.
Example: https://gist.github.com/AlmogBaku/61f183350e10173377f85ee940a84741

If it make sense, i'd love to work on a PR to the existing kit/log/levels API, that contains this enhancement.

peterbourgon:
Dynamic inspection of log level with every Log invocation imposes a significant performance cost that we (ideally) want to avoid, or (at worst) make it explicitly opt-in. Let me propose this alternative to the existing log/levels package, which includes a concept of numeric level.

```go
type Level int

var (
    LevelNone Level = iota
    LevelError
    LevelInfo
    LevelDebug
)

type Levels struct {
    Error log.Logger
    Info  log.Logger
    Debug log.Logger
}

func NewLevels(logger log.Logger, level Level) Levels {
    levels := Levels{
        Error: log.NewNopLogger(),
        Info:  log.NewNopLogger(),
        Debug: log.NewNopLogger(),
    )
    if level >= LevelError {
        levels.Error = log.NewContext(logger).With("level", "error")
    }
    if level >= LevelInfo {
        levels.Info = log.NewContext(logger).With("level", "info")
    }
    if level >= LevelDebug {
        levels.Debug = log.NewContext(logger).With("level", "debug")
    }
    return levels
}

func Example() {
    logger := log.NewLogfmtLogger(os.Stdout)
    levels := levels.New(logger, LevelInfo)

    levels.Debug.Log("msg", "this will NOT get printed")
    levels.Info.Log("msg", "this will get printed")
    levels.Error.Log("msg", "this will also get printed")
}
```

### Alternative

It is also possible to opt-in to paying a runtime cost to get a slightly easier API.

```go
func NewKeyBasedLeveledLogger(logger log.Logger, level Level) log.Logger {
[...]

PR #322, opened 2016-07-16
kit/log/levels: Add enabled level

hnakamur:
I read #252 and #269 and I think I prefer simpler approach for users to check a log level is enabled. So here it is.

The example code:

```
	logger, err := enabled.New(levels.New(log.NewLogfmtLogger(os.Stdout)), "warn")
	if err != nil {
		// This happens only when the level is invalid.
		// In this example, this never happens as the valid level "warn" is used.
		// In a real usage like reading a level from a command line option or a
		// config file, you should check this error.
		panic(err)
	}
	if logger.DebugEnabled() {
		logger.Debug().Log("msg", "hello")
	}
	if logger.WarnEnabled() {
		logger.With("context", "foo").Warn().Log("err", "error")
	}
```

The con is that it takes two more lines for `if logger.XxxEnabled() {` and `}`.
The pro is that it avoids the execution cost of the block for `if` when the level is disabled. I think this is important.

### DIFF
--- a/log/levels/enabled/enabled.go
+++ b/log/levels/enabled/enabled.go
@@ -1,0 +1,89 @@
+package enabled
+
+import "github.com/go-kit/kit/log/levels"
+
+// LevelEnabledLogger provides a wrapper around a logger with a feature
+// for checking a level is enabled.  It has five
+// levels: debug, info, warning (warn), error, and critical (crit).
+//
+// It is users' responsibility not to call the Log method of a logger.
+type LevelEnabledLogger struct {
+	levels.Levels
+	debugEnabled bool
+	infoEnabled  bool
+	warnEnabled  bool
+	errorEnabled bool
+	critEnabled  bool
+}
+
+// New creates a new level enabled logger, wrapping the passed logger.
+func New(logger levels.Levels, option Option) LevelEnabledLogger {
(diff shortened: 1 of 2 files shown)


peterbourgon:
Thanks for the contribution. Doing explicit checks before every error invocation is very verbose, I think a non-starter from a usability perspective. @ChrisHines any other thought?

ChrisHines:
I agree that this approach is not what I want to put in the kit/log/levels package. As I explained in https://github.com/go-kit/kit/issues/269#issuecomment-220379858, the performance gain of the if statement is relatively small.

I do think that this is a viable approach for applications that require optimal performance when logging is disabled. There is no reason an application or organization can't use the technique in their own code, but I don't think the need is common enough to promote this pattern in Go kit.

peterbourgon:
Thanks, Chris. And thanks, Hiroaki, for the submission! We won't be accepting it in the current form but I really appreciate the attention and effort. If you wanted to contribute to Go kit logging, here are some open PRs and issues that need attention:

- PR #189 — add LTSV logger
- Issue #269 — kit/log/levels: support filtering
- Issue #250, PR #252 — Add filtered leveled logging and error promotion
- PR #273 — log: Add concurrent safe wrappers.

PR #323, opened 2016-07-17
kit/log: Add package for structured errors

hnakamur:
With this pull request, you can return a structured error and log the error with key value pairs
attached to it later at a higher call stack.

This is especially useful for libraries since library does not need to log errors and library
users can log errors in their application using their favorite logging package.

Could you take a look?
Thanks!

### DIFF
--- a/log/sterrors/sterrors.go
+++ b/log/sterrors/sterrors.go
@@ -1,0 +1,74 @@
+// Package sterrors provides a structured error.
+//
+// You can return a structured error and log the error with key value pairs
+// attached to it later at a higher call stack. This is especially useful
+// for libraries since library does not need to log errors and library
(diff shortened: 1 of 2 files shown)


peterbourgon:
Thank you for the contribution. But, I don't think this is in the domain of Go kit. An excellent errors package, much better than I could write or maintain, already exists at https://github.com/pkg/errors. It is totally composable with Go kit services.

peterbourgon:
I won't be accepting this PR, but I appreciate your thought and attention!

hnakamur:
I see. Thanks for your time and explanation.
Just for the record, an idea of adding fields to an error was rejected at
https://github.com/pkg/errors/issues/70 and https://github.com/pkg/errors/issues/34

PR #335, opened 2016-08-07
auth/jwt -- JWT service to service authentication

briankassouf:
Adds endpoint middleware for signing and parsing/validating JWT tokens. Also uses the RequestFunc & server/client before patterns to pass token data in and out of the context for remote calls and servers.

uses the [github.com/dgrijalva/jwt-go](https://github.com/dgrijalva/jwt-go) library

Please comment on the design choices

todo:
- [x] impliment endpoint middlewares
- [x] write tests for middlewares
- [x] grpc request funcs
- [x] write tests for gprc funcs
- [x] http request funcs
- [x] write test for http request funcs
- [x] add more documentation
- [ ] add JWT auth to addscv example?

### DIFF
--- a/auth/jwt/README.md
+++ b/auth/jwt/README.md
@@ -56,7 +56,7 @@ func main() {
 ```
 
 In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
-ToHTTPContext(), FromHTTPContext(), ToGRPCContext(), and FromGRPCContext() are given as helpers to do this.
+`ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
 These function impliment the correlating transport's RequestFunc interface and can be passes as ClientBefore or ServerBefore options.
 
 Example of use in a client:
--- a/auth/jwt/README.md
+++ b/auth/jwt/README.md
@@ -1,0 +1,122 @@
+# package auth/jwt
+
+`package auth/jwt` provides a set of interfaces for service authorization through [JSON Web Tokens](https://jwt.io/).
+
+## Usage
+
+NewParser takes a key function and an expected signing method and returns an `endpoint.Middleware`. 
+The middleware will parse a token passed into the context via the `jwt.JWTTokenContextKey`. 
+If the token is valid, any claims will be added to the context via the `jwt.JWTClaimsContextKey`.
+
+```go
+import (
+	stdjwt "github.com/dgrijalva/jwt-go"
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
+
+func main() {
+	var exampleEndpoint endpoint.Endpoint
+	{
+		keyFunc := func(token *stdjwt.Token) (interface{}, error) { return []byte("SigningString"), nil }
+		jwtParser := jwt.NewParser(keyFunc, stdjwt.SigningMethodHS256)
+        
+		exampleEndpoint = MakeExampleEndpoint(service)
+		exampleEndpoint = jwtParser(exampleEndpoint)
+	}
+}
+```
+
+NewSigner takes a JWT key id header, the signing key, signing method, and a claims object. It returns an `endpoint.Middleware`.
+The middleware will build the token string and add it to the context via the `jwt.JWTTokenContextKey`.
+
+```go
+import (
+	stdjwt "github.com/dgrijalva/jwt-go"
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
+
+func main() {
+	var exampleEndpoint endpoint.Endpoint
+	{
+		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
+        
+		exampleEndpoint = grpctransport.NewClient(
+        	. // build client endpoint here
+			.
+			.
+		).Endpoint()
+
+		exampleEndpoint = jwtSigner(exampleEndpoint)
+	}
+}
+```
+
+In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
+ToHTTPContext(), FromHTTPContext(), ToGRPCContext(), and FromGRPCContext() are given as helpers to do this.
+These function impliment the correlating transport's RequestFunc interface and can be passes as ClientBefore or ServerBefore options.
+
+Example of use in a client:
+
+```go
+import (
+	stdjwt "github.com/dgrijalva/jwt-go"
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
+
+func main() {
+
+    options := []httptransport.ClientOption{}
+	var exampleEndpoint endpoint.Endpoint
+	{
+		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
+       
+		options = append(options, httptransport.ClientBefore(jwt.FromGRPCContext()))
+		exampleEndpoint = grpctransport.NewClient(
+        	. // build client endpoint here
+			.
+			options....
+		).Endpoint()
+
+		exampleEndpoint = jwtSigner(exampleEndpoint)
+	}
+}
+```
+
+Example of use in a server:
+
+```go
+import (
+	"golang.org/x/net/context"
+
+	"github.com/go-kit/kit/auth/jwt"
+	"github.com/go-kit/kit/log"
+	grpctransport "github.com/go-kit/kit/transport/grpc"
+)
+
+func MakeGRPCServer(ctx context.Context, endpoints Endpoints, logger log.Logger) pb.ExampleServer {
+	options := []grpctransport.ServerOption{grpctransport.ServerErrorLogger(logger)}
+
+	return &grpcServer{
+		createUser: grpctransport.NewServer(
+			ctx,
+			endpoints.CreateUserEndpoint,
+			DecodeGRPCCreateUserRequest,
+			EncodeGRPCCreateUserResponse,
+			append(options, grpctransport.ServerBefore(jwt.ToGRPCContext()))...,
+		),
+		getUser: grpctransport.NewServer(
+			ctx,
+			endpoints.GetUserEndpoint,
+			DecodeGRPCGetUserRequest,
+			EncodeGRPCGetUserResponse,
+			options...,
+		),
+	}
+}
+```
--- a/auth/jwt/README.md
+++ b/auth/jwt/README.md
@@ -57,7 +57,7 @@ func main() {
 
 In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
 `ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
-These function impliment the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
+These function implement the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
 
 Example of use in a client:
 
--- a/auth/jwt/README.md
+++ b/auth/jwt/README.md
@@ -57,7 +57,7 @@ func main() {
 
 In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
 `ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
-These function implement the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
+These functions implement the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
 
 Example of use in a client:
 
--- a/auth/jwt/README.md
+++ b/auth/jwt/README.md
@@ -57,13 +57,14 @@ func main() {
 
 In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
 `ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
-These function impliment the correlating transport's RequestFunc interface and can be passes as ClientBefore or ServerBefore options.
+These function impliment the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
 
 Example of use in a client:
 
 ```go
 import (
-	stdjwt "github.com/dgrijalva/jwt-go"
+    stdjwt "github.com/dgrijalva/jwt-go"
+    grpctransport "github.com/go-kit/kit/transport/grpc"
     
     "github.com/go-kit/kit/auth/jwt"
     "github.com/go-kit/kit/endpoint"
@@ -76,7 +77,7 @@ func main() {
 	{
 		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
        
-		options = append(options, httptransport.ClientBefore(jwt.FromGRPCContext()))
+		options = append(options, grpctransport.ClientBefore(jwt.FromGRPCContext()))
 		exampleEndpoint = grpctransport.NewClient(
         	. // build client endpoint here
 			.
--- a/auth/jwt/README.md
+++ b/auth/jwt/README.md
@@ -1,90 +1,89 @@
 # package auth/jwt
 
-`package auth/jwt` provides a set of interfaces for service authorization through [JSON Web Tokens](https://jwt.io/).
+`package auth/jwt` provides a set of interfaces for service authorization
+through [JSON Web Tokens](https://jwt.io/).
 
 ## Usage
 
-NewParser takes a key function and an expected signing method and returns an `endpoint.Middleware`. 
-The middleware will parse a token passed into the context via the `jwt.JWTTokenContextKey`. 
-If the token is valid, any claims will be added to the context via the `jwt.JWTClaimsContextKey`.
+NewParser takes a key function and an expected signing method and returns an
+`endpoint.Middleware`. The middleware will parse a token passed into the
+context via the `jwt.JWTTokenContextKey`. If the token is valid, any claims
+will be added to the context via the `jwt.JWTClaimsContextKey`.
 
 ```go
 import (
 	stdjwt "github.com/dgrijalva/jwt-go"
-    
-    "github.com/go-kit/kit/auth/jwt"
-    "github.com/go-kit/kit/endpoint"
+	
+	"github.com/go-kit/kit/auth/jwt"
+	"github.com/go-kit/kit/endpoint"
 )
 
 func main() {
 	var exampleEndpoint endpoint.Endpoint
 	{
-		keyFunc := func(token *stdjwt.Token) (interface{}, error) { return []byte("SigningString"), nil }
-		jwtParser := jwt.NewParser(keyFunc, stdjwt.SigningMethodHS256)
-        
+		kf := func(token *stdjwt.Token) (interface{}, error) { return []byte("SigningString"), nil }
 		exampleEndpoint = MakeExampleEndpoint(service)
-		exampleEndpoint = jwtParser(exampleEndpoint)
+		exampleEndpoint = jwt.NewParser(kf, stdjwt.SigningMethodHS256)(exampleEndpoint)
 	}
 }
 ```
 
-NewSigner takes a JWT key id header, the signing key, signing method, and a claims object. It returns an `endpoint.Middleware`.
-The middleware will build the token string and add it to the context via the `jwt.JWTTokenContextKey`.
+NewSigner takes a JWT key ID header, the signing key, signing method, and a
+claims object. It returns an `endpoint.Middleware`. The middleware will build
+the token string and add it to the context via the `jwt.JWTTokenContextKey`.
 
 ```go
 import (
 	stdjwt "github.com/dgrijalva/jwt-go"
-    
-    "github.com/go-kit/kit/auth/jwt"
-    "github.com/go-kit/kit/endpoint"
+	
+	"github.com/go-kit/kit/auth/jwt"
+	"github.com/go-kit/kit/endpoint"
 )
 
 func main() {
 	var exampleEndpoint endpoint.Endpoint
 	{
-		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
-        
-		exampleEndpoint = grpctransport.NewClient(
-        	. // build client endpoint here
-			.
-			.
-		).Endpoint()
-
-		exampleEndpoint = jwtSigner(exampleEndpoint)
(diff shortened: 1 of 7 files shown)


peterbourgon:
Brilliant. The overall shape and API surface area is right on point. I have lots of little comments and suggestions, but I'll hold off until it's code complete. Looking great, and excited to see the final product!

auth/jwt/middleware.go
```diff
@@ -0,0 +1,107 @@
...
+
+			return next(ctx, request)
+		}
+	}
+}
+
+// Create a new JWT token parsing middleware, specifying a jwt.Keyfunc interface and the
+// signing method. Adds the resulting claims to endpoint context or returns error on invalid
+// token. Particularly useful for servers.
+func NewParser(keys KeySet) endpoint.Middleware {
```
cam-stitt: It seems like we are missing passing through a `jwt.Keyfunc` as per the comment suggests? We currently have a jwt implementation and we require a very different Keyfunc that what is being used in this method.

auth/jwt/transport.go
```diff
@@ -0,0 +1,83 @@
...
+			(*md)["authorization"] = []string{generateAuthHeaderFromToken(token)}
+		}
+
+		return ctx
+	}
+}
+
+// extractTokenFromAuthHeader returns the token from the value of the Authorzation header
+func extractTokenFromAuthHeader(val string) (token string, ok bool) {
+	if len(val) < 8 || !strings.EqualFold(val[0:7], "BEARER ") {
```
cam-stitt: Rather than the precise lengths, maybe a simpler solution could be to split on `" "` and use the parts? For reference: https://github.com/auth0/go-jwt-middleware/blob/master/jwtmiddleware.go#L130

peterbourgon: Is there a JWT spec that enumerates stuff like this, or is it totally up to the implementation?

auth/jwt/middleware.go
briankassouf: Hey @cam-stitt, what is included in your keyfunc? I was trying to build an interface that allowed for easy use of the `KID` (Key ID) header as means of identifying which key to use to verify a token. If it makes more sense to have added flexibility via a keyfunc then I can revert back to using that. (Comment hasn't been updated since i made this last change) 

See code at https://github.com/go-kit/kit/blob/1961299ceed1e99ddb9a868a51cd2177cf98583b/auth/jwt/middleware.go to see the keyfunc implementation.

auth/jwt/transport.go
briankassouf: The spec recommends sending it in the authorization header, but reading headers from the transport is up to the implementation. 

> Whenever the user wants to access a protected route or resource, the user agent should send the JWT, typically in the Authorization header using the Bearer schema. - https://jwt.io/introduction/#how-do-json-web-tokens-work-

auth/jwt/middleware.go
cam-stitt: Our keyfunc is extremely minimal. It doesn't perform validation at that point. We are using auth0 so follow a very similar pattern to them. I think it's important to allow for this customisation, as people using providers like auth0 or stormpath will not have the ability to modify the tokens (I don't believe they have a KID header either).

auth/jwt/transport.go
cam-stitt: As @briankassouf said, it's up to the implementation. But it would be nice to probably just move things like `BEARER` into a const. Also, splitting on `" "` would allow for future enhancements if we were to use different schemas. The current implementation is dependent on the schema word length.

auth/jwt/middleware.go
briankassouf: I could be wrong, as I have never used auth0 or stormpath, but my understanding is those services implement single sign on authentication for users to authenticate to an API. As I designed it (and again I could have misinterpreted the intention) this package implements service-to-service authentication. Meaning we are authenticating that a request is coming from another service inside our perimeter. These requests would be accompanied with JWT tokens that are signed inside a client from our own infrastructure (somehow sharing secret keys between the two services). This would allow us to pass a KID header with the shared Key ID and, if this is the correct intention of this package, then handling user authenticated tokens from a SSO provider would be additional functionality built into the service code.

cam-stitt: The situation with auth0 is no different to a standard auth0
implementation. We have a secret that gets returned in the key func that
decodes the token, you validate the signing method and the validity of the
token like any other jwt implementation. Take a look through the auth0 jwt
middleware package I posted in a previous comment. If you allow for custom
keyfuncs any implementation can be handled. In our case each service has
the secret key and decodes the token the same as any other.

This package can be used for both service to service as well as endpoint
protection from a ui etc.

auth/jwt/transport.go
peterbourgon: I agree that the current implementation seems needlessly fragile.

auth/jwt/middleware.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"golang.org/x/net/context"
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/go-kit/kit/endpoint"
+)
+
+const (
+	// JWTTokenContextKey holds the key used to store a JWT Token in the context
+	JWTTokenContextKey = "JWTToken"
```
nuss-justin: Shouldn't this (and JWTClaimsContextKey) have an unexported type to avoid collisions? Something like

```go
type key string

const (
	// JWTTokenContextKey holds the key used to store a JWT Token in the context
 	JWTTokenContextKey key = "JWTToken"
	// JWTClaimsContxtKey holds the key used to store the JWT Claims in the context
 	JWTClaimsContextKey key = "JWTClaims"
)
```

See https://blog.golang.org/context#TOC_3.2.

peterbourgon: 👍

peterbourgon:
Hey, what's the story here? I'd love to merge this in, and if necessary I'm happy to take it over the finish line, but I don't want to usurp your work :)

peterbourgon:
@briankassouf Any chance you'd be able to carry this over this finish line this week, or should I take over?

briankassouf:
@peterbourgon Sorry for the delay, haven't had much time to work on this the last few weeks. I'll try and clean it up a bit today and hopefully i'll get some time on Sunday as well.

auth/jwt/README.md
```diff
@@ -57,13 +57,14 @@ func main() {
 
 In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
 `ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
-These function impliment the correlating transport's RequestFunc interface and can be passes as ClientBefore or ServerBefore options.
+These function impliment the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
```
groob: s/impliment/implement

briankassouf: doh! good catch!

briankassouf:
@peterbourgon I believe this is ready to go. Mind giving it a review?

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
+package jwt
+
+import (
+	"errors"
+
+	"golang.org/x/net/context"
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/go-kit/kit/endpoint"
+)
```
peterbourgon: Please group third-party imports together, separate from Go kit imports, i.e.

```go
import (
    "errors"
  
    "golang.org/x/net/context"
    jwt "github.com/dgrijalva/jwt-go"

)
```

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+
+type contextKey string
+
+const (
+	// JWTTokenContextKey holds the key used to store a JWT Token in the context
```
peterbourgon: Please add a period to the end of the comment sentence, and wrap at 80 col if it isn't already.

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+	JWTTokenContextKey contextKey = "JWTToken"
+	// JWTClaimsContxtKey holds the key used to store the JWT Claims in the context
```
peterbourgon: Likewise above.

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+var (
+	ErrTokenContextMissing     = errors.New("Token up for parsing was not passed through the context")
+	ErrTokenInvalid            = errors.New("JWT Token was invalid")
+	ErrTokenExpired            = errors.New("JWT Token is expired")
+	ErrTokenMalformed          = errors.New("JWT Token is malformed")
+	ErrTokenNotActive          = errors.New("Token is not valid yet")
+	ErrUnexpectedSigningMethod = errors.New("Unexpected signing method")
+)
```
peterbourgon: These all need comments, and error strings should not begin with capital letters (excepting those beginning with JWT).

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+
+type Claims map[string]interface{}
```
peterbourgon: This type needs a comment. But why is it a separate type, versus just using jwt.MapClaims?

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Create a new JWT token generating middleware, specifying signing method and the claims
+// you would like it to contain. Particularly useful for clients.
+func NewSigner(kid string, key []byte, method jwt.SigningMethod, claims Claims) endpoint.Middleware {
```
peterbourgon: Comment should be of the form "NewSigner creates a new...". Also, can you mention what kid is?

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+
+			return next(ctx, request)
+		}
+	}
+}
+
+// Create a new JWT token parsing middleware, specifying a jwt.Keyfunc interface and the
+// signing method. Adds the resulting claims to endpoint context or returns error on invalid
+// token. Particularly useful for servers.
+func NewParser(keyFunc jwt.Keyfunc, method jwt.SigningMethod) endpoint.Middleware {
```
peterbourgon: Likewise above.

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+			// tokenString is stored in the context from the transport handlers
+			tokenString, ok := ctx.Value(JWTTokenContextKey).(string)
+			if !ok {
+				return nil, ErrTokenContextMissing
+			}
+
+			// Parse takes the token string and a function for looking up the key. The latter is especially
+			// useful if you use multiple keys for your application.  The standard is to use 'kid' in the
+			// head of the token to identify which key to use, but the parsed token (head and claims) is provided
+			// to the callback, providing flexibility.
```
peterbourgon: Please wrap the comment block at 80col

auth/jwt/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+					if e.Errors&jwt.ValidationErrorMalformed != 0 {
+						// Token is malformed
+						return nil, ErrTokenMalformed
+					} else if e.Errors&jwt.ValidationErrorExpired != 0 {
+						// Token is expired
+						return nil, ErrTokenExpired
+					} else if e.Errors&jwt.ValidationErrorNotValidYet != 0 {
+						// Token is not active yet
+						return nil, ErrTokenNotActive
+					}
```
peterbourgon: Are all of these mutually exclusive?

auth/jwt/transport.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"fmt"
+	stdhttp "net/http"
+	"strings"
+
+	"github.com/go-kit/kit/transport/grpc"
+	"github.com/go-kit/kit/transport/http"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc/metadata"
+)
```
peterbourgon: As above, please separate third-party imports away from Go kit imports.

auth/jwt/transport.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	BEARER        string = "bearer"
+	BEARER_FORMAT string = "Bearer %s"
+)
```
peterbourgon: If these should be exported they should be called Bearer and BearerFormat and get comments. If not, they should be called bearer and bearerFormat.

auth/jwt/transport.go
```diff
@@ -0,0 +1,89 @@
...
+
+// moves JWT token from request header to context
+// particularly useful for servers
+func ToHTTPContext() http.RequestFunc {
```
peterbourgon: Method comment should be of the form "ToHTTPContext moves the JWT token..." and be proper sentences.

auth/jwt/transport.go
```diff
@@ -0,0 +1,89 @@
...
+			return ctx
+		}
+
+		return context.WithValue(ctx, JWTTokenContextKey, token)
+	}
+}
+
+// moves JWT token from context to request header
+// particularly useful for clients
+func FromHTTPContext() http.RequestFunc {
```
peterbourgon: Likewise above re: comment. Further comments about function comments in this file elided.

auth/jwt/transport.go
```diff
@@ -0,0 +1,89 @@
...
+			// capital "Key" is illegal in HTTP/2.
+			(*md)["authorization"] = []string{generateAuthHeaderFromToken(token)}
+		}
+
+		return ctx
+	}
+}
+
+// extractTokenFromAuthHeader returns the token from the value of the Authorzation header
+func extractTokenFromAuthHeader(val string) (token string, ok bool) {
```
peterbourgon: Comment not strictly necessary on unexported functions. Either remove it, or make it a proper sentence and wrap it at 80 col.

auth/jwt/middleware_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	kid           = "kid"
+	key           = []byte("test_signing_key")
+	method        = jwt.SigningMethodHS256
+	invalidMethod = jwt.SigningMethodRS256
+	claims        = Claims{"user": "go-kit"}
+	signedKey     = "eyJhbGciOiJIUzI1NiIsImtpZCI6ImtpZCIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjoiZ28ta2l0In0.14M2VmYyApdSlV_LZ88ajjwuaLeIFplB8JpyNy0A19E"
+	invalidKey    = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.e30.vKVCKto-Wn6rgz3vBdaZaCBGfCBDTXOENSo_X2Gq7qA"
```
peterbourgon: Where did these strings come from? Can you add a comment indicating how to generate them?

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+The middleware will parse a token passed into the context via the `jwt.JWTTokenContextKey`. 
+If the token is valid, any claims will be added to the context via the `jwt.JWTClaimsContextKey`.
+
+```go
+import (
+	stdjwt "github.com/dgrijalva/jwt-go"
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
```
peterbourgon: Spaces/tabs

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+func main() {
+	var exampleEndpoint endpoint.Endpoint
+	{
+		keyFunc := func(token *stdjwt.Token) (interface{}, error) { return []byte("SigningString"), nil }
+		jwtParser := jwt.NewParser(keyFunc, stdjwt.SigningMethodHS256)
+        
+		exampleEndpoint = MakeExampleEndpoint(service)
+		exampleEndpoint = jwtParser(exampleEndpoint)
```
peterbourgon: How about

```go
exampleEndpoint = jwt.NewParser(kf, stdjwt.SigningMethodHS256)(exampleEndpoint)
```

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
+
+func main() {
+	var exampleEndpoint endpoint.Endpoint
+	{
+		keyFunc := func(token *stdjwt.Token) (interface{}, error) { return []byte("SigningString"), nil }
+		jwtParser := jwt.NewParser(keyFunc, stdjwt.SigningMethodHS256)
```
peterbourgon: If possible, it would be great to avoid the necessity of importing 2 jwt packages to use this in the common/default case. Unfortunately I don't really know enough about jwt to make strong recommendations :) Is there any way to produce a standard keyFunc, or will every user need to do something totally custom? Is stdjwt.SigningMethodHS256 "preferred" enough that it's worth treating it as a default, and having either another constructor or a functional option to specify something different?

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+        
+		exampleEndpoint = MakeExampleEndpoint(service)
+		exampleEndpoint = jwtParser(exampleEndpoint)
+	}
+}
+```
+
+NewSigner takes a JWT key id header, the signing key, signing method, and a claims object. It returns an `endpoint.Middleware`.
```
peterbourgon: key ID header

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+The middleware will build the token string and add it to the context via the `jwt.JWTTokenContextKey`.
+
+```go
+import (
+	stdjwt "github.com/dgrijalva/jwt-go"
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
```
peterbourgon: Tabs/spaces

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+func main() {
+	var exampleEndpoint endpoint.Endpoint
+	{
+		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
+        
+		exampleEndpoint = grpctransport.NewClient(
+        	. // build client endpoint here
+			.
+			.
+		).Endpoint()
```
peterbourgon: Tabs/spaces, and you can just as well do

```go
exampleEndpoint = grpctransport.NewClient(...).Endpoint()
```

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+
```
peterbourgon: Remove newline

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+		exampleEndpoint = jwtSigner(exampleEndpoint)
```
peterbourgon: Like above, what do you think about inlining the jwtSigner?

```go
exampleEndpoint = jwt.NewSigner(
    "kid-header",
    []byte("SigningString"),
    stdjwt.SigningMethodHS256,
    jwt.Claims{},
)(exampleEndpoint)
```

Maybe that's too long or worse than the alternative -- totally your call.

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+```
+
+In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
+`ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
+These functions implement the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
```
peterbourgon: Please wrap this and other longer paragraphs in the README at 80col.

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+    
+    "github.com/go-kit/kit/auth/jwt"
+    "github.com/go-kit/kit/endpoint"
+)
+
+func main() {
+
+    options := []httptransport.ClientOption{}
+	var exampleEndpoint endpoint.Endpoint
+	{
```
peterbourgon: Tabs/spaces

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+In order for the parser and the signer to work, the authorization headers need to be passed between the request and the context.
+`ToHTTPContext()`, `FromHTTPContext()`, `ToGRPCContext()`, and `FromGRPCContext()` are given as helpers to do this.
+These functions implement the correlating transport's RequestFunc interface and can be passed as ClientBefore or ServerBefore options.
+
+Example of use in a client:
+
+```go
+import (
+    stdjwt "github.com/dgrijalva/jwt-go"
+    grpctransport "github.com/go-kit/kit/transport/grpc"
```
peterbourgon: Move this import down with the other Go kit imports.

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+	{
+		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
+       
+		options = append(options, grpctransport.ClientBefore(jwt.FromGRPCContext()))
+		exampleEndpoint = grpctransport.NewClient(
+        	. // build client endpoint here
+			.
+			options....
+		).Endpoint()
+
```
peterbourgon: Extraneous newline

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+		exampleEndpoint = jwtSigner(exampleEndpoint)
```
peterbourgon: Same comment as above re: inlining.

auth/jwt/README.md
```diff
@@ -0,0 +1,123 @@
...
+	var exampleEndpoint endpoint.Endpoint
+	{
+		jwtSigner := jwt.NewSigner("kid-header", []byte("SigningString"), stdjwt.SigningMethodHS256, jwt.Claims{})
+       
+		options = append(options, grpctransport.ClientBefore(jwt.FromGRPCContext()))
+		exampleEndpoint = grpctransport.NewClient(
+        	. // build client endpoint here
+			.
+			options....
+		).Endpoint()
```
peterbourgon: Same as above, e.g.

```go
exampleEndpoint = grpctransport.NewClient(..., grpctransport.ClientBefore(jwt.FromGRPCContext()))
```

auth/jwt/middleware.go
briankassouf: I added it so that when pulling claim data out of the context (inside a service instance, or another endpoint middleware) the caller did not need to import both std and go-kit's JWT packages. If you think it's unnecessary I can remove it.

auth/jwt/middleware.go
briankassouf: Yes, they should all be mutually exclusive. Pulled the strategy from here: https://godoc.org/github.com/dgrijalva/jwt-go#example-Parse--ErrorChecking

auth/jwt/README.md
briankassouf: Hmm, I could add a jwt.NewDefaultParser() function. I also don't know off hand what the desired default would be. https://jwt.io defaults to `HS256` (`SHA-256`) but I could see `HS512` (`SHA-512`) being pretty common as well.

peterbourgon:
Hell yeah! 🌮

cam-stitt:
Great work getting this done @briankassouf. Excited to try it out :)

PR #342, opened 2016-08-16
Clarify metric comments

brian-brazil:
Gauges have different values at different times.

Quantiles and buckets are distinct things (Prometheus Summary vs Histogram).


peterbourgon:
Sure! Thanks.

PR #344, opened 2016-08-19
Add unit tests for etcd Client

jstordeur:


sd/etcd/client_test.go
```diff
@@ -62,3 +65,249 @@ func TestOptions(t *testing.T) {
...
+	}
+}
+
+// getfakeClient return a new etcd.Client built on top of the mocked interfaces
+func getFakeClient(event, err chan bool, getres *getResult) (Client, error) {
+	return &client{
+		keysAPI: &fakeKeysAPI{event, err, getres},
+		ctx:     context.Background(),
+	}, nil
+}
```
peterbourgon: No reason to have the return signature include an error that's always nil, unless I'm missing something?

```go
func newFakeClient(...) Client {
   return &client{...}
}
```

peterbourgon:
Other than that one nit, looks good! For the record: this increases coverage of package etcd from 35.1% to 68.9%. [Here is the coverage report](http://peter.bourgon.org/etcd.html). Would you be up to adding some tests to registrar, too? ;)

sd/etcd/client_test.go
jstordeur: True, I'll change that

jstordeur:
I'll have a look at registrar

peterbourgon:
78.4% 😎

peterbourgon:
Amazing! Thank you so much.

Issue #347, opened 2016-08-23
Example of ToHttpRequest tracing.


cam-stitt:
Hi there,

I'm looking for an example of ToHttpRequest usage. Currently, I have a similar setup to the addsvc that includes to seperate services communicating through a client. Within zipkin, I can see the spans for sending service and the client (which has the sending service name) but I can't see anything for the receiving service. I can't work out if it's something I'm missing, or if this is how it is supposed to work. Currently on phone, but can add screenshots of zipkin or anything else that may help.

Thanks,

Cameron

cam-stitt:
I should note that I tried adding ToHttpRequest as a ClientBefore func, but it didn't help.

peterbourgon:
Can you provide a minimal example which demonstrates the problem?

cam-stitt:
Right, that makes sense. The link you supplied for ToHTTPRequest is using FromHTTPRequest, though. I'll make that change and see if it fixes it.

peterbourgon:
Ha! You're right. D'oh.

cam-stitt:
That certainly seemed to tidy up some of the tracing. The only thing I can see now is that the client send/receive seems to not be quite right. I assume this is because the other service isn't picking up the sent trace?

<img width="577" alt="screen shot 2016-08-23 at 8 33 55 pm" src="https://cloud.githubusercontent.com/assets/186643/17888937/f5943470-6970-11e6-8d75-9c98340bad40.png">

cam-stitt:
Nvm. that was my own stupidity.

cam-stitt:
Btw, thanks for your help!

PR #348, opened 2016-08-23
Use To rather than From in tracing examples.

cam-stitt:
As discussed in #347, FromHTTPRequest was used on the client instead of ToHTTPRequest. It was also the case with GRPC. This updates the examples to use the correct RequestFuncs.


peterbourgon:
Brilliant! Thanks a ton.

PR #353, opened 2016-09-01
Add StartTimer/ObserveDuration to histograms.

brian-brazil:
This makes it easy to record latencies,
and follows the Prometheus conventions.

This is a first pass, I think I need to change HistogramTImer to taking a function rather than a Histogram so that the value can be changed to ms for statsd. If this library grows to support the pushgateway equivalents we'll also need it on gauges.

### DIFF
--- a/metrics/circonus/circonus.go
+++ b/metrics/circonus/circonus.go
@@ -83,3 +83,8 @@ func (h *Histogram) With(labelValues ...string) metrics.Histogram { return h }
 
 // Observe implements Histogram. No precision is lost.
 func (h *Histogram) Observe(value float64) { h.h.RecordValue(value) }
+
+// StartTimer implements Histogram.
+func (h *Histogram) StartTimer() metrics.HistogramTimer {
+  return metrics.NewHistogramTimer(h)
+}
--- a/metrics/discard/discard.go
+++ b/metrics/discard/discard.go
@@ -35,3 +35,8 @@ func (h histogram) With(labelValues ...string) metrics.Histogram { return h }
 
 // Observe implements histogram.
 func (h histogram) Observe(value float64) {}
+
+// StartTimer implements Histogram.
+func (h histogram) StartTimer() metrics.HistogramTimer {
+  return metrics.NewHistogramTimer(h)
+}
--- a/metrics/dogstatsd/dogstatsd.go
+++ b/metrics/dogstatsd/dogstatsd.go
@@ -283,6 +283,11 @@ func (t *Timing) Observe(value float64) {
 	t.obs(t.name, t.lvs, value)
 }
 
+// StartTimer implements metrics.Histogram.
+func (t *Timing) StartTimer() metrics.HistogramTimer {
+  return metrics.NewHistogramTimer(t)
+}
+
 // Histogram is a DogStatsD histrogram. Observations are forwarded to a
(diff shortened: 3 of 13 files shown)


peterbourgon:
Given

```go
var h metrics.Histogram
```

then it looks like

```go
t := h.StartTimer()
doStuff()
t.ObserveDuration()
```

is equivalent to

```go
t := time.Now()
doStuff()
h.Observe(time.Since(t))
```

is that correct?

brian-brazil:
Yes, that's correct.

peterbourgon:
Okay. Given it's something like a helper method, I'm not sure I see the value in modifying the interface definition across all of the implementations. Is there some knock-on benefit I'm not grokking? I would rather see this implemented as a free function, e.g.

```go
package metrics

func NewTimer(h Histogram) *Timer {
    return &Timer{
        h: h,
        t: time.Now(),
    }
}

func (t *Timer) ObserveDuration() {
    d := time.Since(t.t).Seconds()
    if d < 0 {
        d = 0 // can this actually happen?
    }
    t.h.Observe(d)
}
```

and used like

```go
t := metrics.NewTimer(h) // instead of h.StartTimer()
doStuff()
t.ObserveDuration()
```

Would you be OK with that?

brian-brazil:
Hmm, the problem with that is that if you're using just one of the libraries you then have to do an extra import in order to get to what should be a standard simple to use convenience function. This is particularly odd for the ones with a `Timer`, as it's not quite obvious how to use those to time things.

peterbourgon:
To your first point, given that some application code is operating on a metrics.Histogram, it's already importing package metrics, so that's not a problem. To your second point, I agree that Timer is not the perfect name... but it is at least unambiguous when fully-scoped i.e. metrics.Timer. Maybe... metrics.Stopwatch?

brian-brazil:
I'd expect users to act on just one of the sub-packages, as you'd be creating some metrics in every file. In addition if there's extra names you want to access (e.g. Inc/Dec on Gauge for Prometheus or the Timers for statsd) you have to import those and do some casting if you're combining with labels as the label functions return the interface.

The abstraction is already leaking a bit due to per-backend differences.

> I agree that Timer is not the perfect name

I think Timer is a fine name for this, I was referring to the existing `Timing` in statsd/dogstatsd.

peterbourgon:
> I'd expect users to act on just one of the sub-packages, as you'd be creating some metrics in every file.

Very much not. The Go kit idiom is to create all metrics in the `func main` and pass them as dependencies — and as metrics.{Counter, Gauge, Histogram} interface types — to the components that need them. There is no global state in Go kit :)

> In addition if there's extra names you want to access (e.g. Inc/Dec on Gauge for Prometheus or the Timers for statsd) you have to import those and do some casting if you're combining with labels as the label functions return the interface.

The idea is that the single methods provided in the interfaces in package metrics are enough to get you started on your instrumentation journey, and will carry you quite far. If and when you want more sophisticated access patterns, then you will likely eschew package metrics altogether and use the core types (e.g. prometheus.Summary) directly.

Thus,

> The abstraction is already leaking a bit due to per-backend differences.

Yes, the abstraction is imperfect, for the reasons above.

peterbourgon:
Closed in favor of #358. Thanks for the feedback and contribution!

PR #358, opened 2016-09-06
metrics: add Timer

peterbourgon:
As a counterproposal to #353, @brian-brazil, what do you think of this?

### DIFF
--- a/metrics/timer.go
+++ b/metrics/timer.go
@@ -9,7 +9,7 @@ type Timer struct {
 	t time.Time
 }
 
-// NewTimer wraps the given histogram and marks the current time.
+// NewTimer wraps the given histogram and records the current time.
 func NewTimer(h Histogram) *Timer {
 	return &Timer{
 		h: h,
--- a/metrics/timer.go
+++ b/metrics/timer.go
@@ -1,0 +1,28 @@
+package metrics
+
+import "time"
+
+// Timer acts as a stopwatch, sending observations to a wrapped histogram.
+// It's a bit of helpful syntax sugar for h.Observe(time.Since(x)).
+type Timer struct {
+	h Histogram
+	t time.Time
+}
+
+// NewTimer wraps the given histogram and marks the current time.
+func NewTimer(h Histogram) *Timer {
+	return &Timer{
+		h: h,
+		t: time.Now(),
+	}
+}
+
+// ObserveDuration captures the number of seconds since the timer was
+// constructed, and forwards that observation to the histogram.
+func (t *Timer) ObserveDuration() {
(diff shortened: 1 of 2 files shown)


metrics/timer.go
```diff
@@ -0,0 +1,28 @@
...
+import "time"
+
+// Timer acts as a stopwatch, sending observations to a wrapped histogram.
+// It's a bit of helpful syntax sugar for h.Observe(time.Since(x)).
+type Timer struct {
+	h Histogram
+	t time.Time
+}
+
+// NewTimer wraps the given histogram and marks the current time.
```
brian-brazil: I'd avoid the work "mark" as that's used by Dropwizard Meters and may cause confusion, 'records' or 'notes' should do.

metrics/timer_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/generic"
+)
+
+func TestTimerFast(t *testing.T) {
+	h := generic.NewSimpleHistogram()
+	metrics.NewTimer(h).ObserveDuration()
+
+	tolerance := 0.005
```
brian-brazil: We've seen 10ms not be enough in timing sensitive Prometheus tests. I'd use at least 50ms.

brian-brazil:
From a naming standpoint the Gauge will need to also support timing if batch job monitoring is added at any point.

I'm still not a fan of everyone having to import `metrics`, the approach suggested presumes that every user is happy with breaking changes anytime any of their potentially tens to hundreds of upstreams changes their metrics. I think consideration should be given to a user who wants to use just one backend, as that's a scalable approach.

metrics/timer_test.go
peterbourgon: For an elapsed time of zero?!

metrics/timer.go
peterbourgon: Will do!

peterbourgon:
Thanks for the feedback!

> From a naming standpoint the Gauge will need to also support timing if batch job monitoring is added at any point.

The Timer is syntax sugar for already-supported use cases. I can _just barely_ be convinced that the Histogram use case is prolific enough to warrant expanding the surface area of the package API with a new type. But I think the Gauge use case you describe falls well short of that mark, especially considering batch jobs are rather far outside of Go kit's purview. Still, thank you!

metrics/timer_test.go
ChrisHines: FWIW, with Go 1.6.x on Windows, time.Now has a resolution ~15ms.

Issue #341, opened 2016-08-15
Add custom error behavior for lb.Retry


kfir-stratoscale:
Currently there is no way to add custom error behavior in the Retry module.
For example we might want to stop after a single retry if the error that we got from the underlying http transport is NotFound error.

Allowing the user to provide a callback that will be used to check the errors can solve this in a clean way.

PR #360, opened 2016-09-08
Added RetryWithCallback allowing cb to terminate early with message

morganhein:
https://github.com/go-kit/kit/issues/341

Love them callbacks.


peterbourgon:
Your gofmt seems to be broken; it's inserting spaces instead of tabs?

sd/lb/retry_test.go
```diff
@@ -88,3 +88,44 @@ func TestRetryTimeout(t *testing.T) {
...
+        endpoints = sd.FixedSubscriber{} // no endpoints
+        lb        = loadbalancer.NewRoundRobin(endpoints)
+        retry    = loadbalancer.RetryWithCallback(999, time.Second, lb, cb) // lots of retries
+        ctx       = context.Background()
+    )
+     _, err := retry(ctx, struct{}{})
+     if err == nil {
+        t.Errorf("expected error, got none") // should fail
+    }
+    if err.Error() != "Aborting early" {
```
peterbourgon: Comparing strings is not a great way to compare errors. You'll want the callback to accept and return a proper `error` type.

sd/lb/retry.go
```diff
@@ -1,57 +1,83 @@
...
-	"github.com/go-kit/kit/endpoint"
+    "github.com/go-kit/kit/endpoint"
 )
 
+// Callback function that indicates the current attempt count and the error encountered.
+//  Should return whether the Retry function should continue trying, and a custom 
+//  error message if desired. The error message may be nil, but a true/false 
+//  is always expected. In all cases if the error message is supplied, the 
+//  current error will be replaced.
+type Callback func(int, string) (bool, *string)
```
peterbourgon: Hmm, a pointer to a string is not a great way to encode maybe-error. It would make a lot more sense if Callback had the signature

```go
type Callback func(n int, err error) (continue bool, err error)
```

morganhein: Yeah, I should've done that in the first place. Still learning obv. I'll fix spaces/tabs and the signature.

peterbourgon:
Ping? :)

rossmcf:
I agree with @peterbourgon. Allowing the callback to receive an error would make it easier to handle the case described in #341, where you want to check for a particular error type.

sd/lb/retry.go
rossmcf: I just gave this signature a try. I think the named return params help legibility, but `continue` is a keyword, so you'll probably have to go with `cont`.

peterbourgon:
Looks like this has been abandoned. I'm closing in favor of #370, but please correct me if I've got it wrong :) Thanks for the initiative! :)

Issue #354, opened 2016-09-03
metrics: add support for PCP


suyash:
I was [a part of this year's Google Summer of Code](https://summerofcode.withgoogle.com/projects/#6178186094182400) with the [Performance Co-Pilot organization](http://pcp.io/). PCP, although principally built for reporting low level application specific metrics, also supports instrumentation through a specially configured agent, mmv. My project was to build an instrumentation client library for that API, and it is currently at [performancecopilot/speed](https://github.com/performancecopilot/speed).

One of the goals of the project was to demonstrate real world usage of the implemented client library, and I am trying to fulfill that requirement by implementing PCP support for this library. My fork contains a pcp subpackage inside metrics at https://github.com/suyash/kit.

I have also modified 3 examples to report metrics to PCP instead of prometheus and they are implemented [in a separate tree](https://github.com/suyash/kit/tree/pcp-examples/examples). For stringsvc2, I have added [some screenshots](https://github.com/suyash/kit/tree/pcp-examples/examples/stringsvc2#stringsvc2).

Currently I cannot support `With` labels for metrics, because while we do have a similar concept (https://github.com/performancecopilot/speed#instancemetric), it requires all labels to be specified upfront, and reinitializing a new metric each time seemed overkill. Instead I am [trying to explore supporting mutating instances](https://github.com/performancecopilot/speed/issues/32)

Anyways, in line with the contributing guidelines, I am opening this issue for discussion

peterbourgon:
Neat! I'll take a look in the next few days.

> Currently I cannot support With labels for metrics, because while we do have a similar concept (https://github.com/performancecopilot/speed#instancemetric), it requires all labels to be specified upfront,

This is also true of Prometheus! Just make sure the With invocations are storing the label-values in a way that's retrievable by the observe method, and have the observe method pass the values to the underlying implementation. You can be permissive, and say that required unspecified dimensions get an "unknown" value, and non-required erroneously-specified dimensions are ignored. Or, you can be super strict, and panic on either condition. Either one is fine from my POV.

peterbourgon:
Global state is always a bad idea, and in Go kit it's strictly forbidden. So you'll need to find a way to arrange things so that you don't have a package global registry and client. I haven't dug too deeply, but one way might be by encapsulating them in a pcp.Reporter struct, and have the NewMetric constructors be methods on that type.

peterbourgon:
Great! The design of the new package looks reasonable from a high level.

[...]

PR #363, opened 2016-09-14
metrics: add support for PCP

suyash:
Based on the discussion in #354 

This is basically the work in my repo except for the readme changes

I have not squashed the commits as GitHub allows squashing all commits into one at merge step, however if it is important, I can squash all changes and submit just one commit


metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
+package pcp
+
+import (
+	"github.com/go-kit/kit/metrics"
+	"github.com/performancecopilot/speed"
+)
```
peterbourgon: Please group imports according to the [canonical style](https://github.com/golang/go/wiki/CodeReviewComments#imports), that is,

```go
import (
    // stdlib

    // third-party, not Go kit

    // Go kit 
)
```

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Reporter encapsulates a speed client
```
peterbourgon: Thanks for the method comments! Please use complete sentences, including proper punctuation and grammar, i.e.

```go
// Reporter encapsulates a speed client.
```

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+type Reporter struct {
+	c *speed.PCPClient
+}
+
+// NewReporter creates a new reporter instance
+func NewReporter(appname string) *Reporter {
+	c, err := speed.NewPCPClient(appname)
+	if err != nil {
+		panic(err)
```
peterbourgon: Is that necessary? Can it return `(*Reporter, error)` instead?

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/go-kit/kit/metrics"
+	"github.com/performancecopilot/speed"
+)
+
+// Reporter encapsulates a speed client
+type Reporter struct {
+	c *speed.PCPClient
+}
+
+// NewReporter creates a new reporter instance
```
peterbourgon: 1. Please use proper punctuation, as above.
2. Please describe a bit the meaning of the appname parameter, and the behavior if there is some error or misconfiguration.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+func NewReporter(appname string) *Reporter {
+	c, err := speed.NewPCPClient(appname)
+	if err != nil {
+		panic(err)
+	}
+	return &Reporter{c}
+}
+
+// Start starts reporting currently registered metrics to the backend
```
peterbourgon: 1. Please use proper punctuation (further comments of this type elided).
2. What does "backend" mean?

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+func (r *Reporter) Start() { r.c.MustStart() }
+
+// Stop stops reporting currently registered metrics to the backend
+func (r *Reporter) Stop() { r.c.MustStop() }
+
+////////////////////////////////////////////////////////////////////////////////////////
```
peterbourgon: Please remove separator comments like this one.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+
+// NewCounter creates a new Counter
+//
+// this requires a name parameter
+// and optionally takes a couple of string directions, which
+// are directly passed to speed
+func (r *Reporter) NewCounter(name string, desc ...string) *Counter {
+	c, err := speed.NewPCPCounter(0, name, desc...)
+	if err != nil {
+		panic(err)
```
peterbourgon: Again, can this return `(*Counter, error)` instead?

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+// Counter implements metrics.Counter via a single dimensional speed.Counter
+// for now, see https://github.com/performancecopilot/speed/issues/32
+type Counter struct {
+	c speed.Counter
+}
+
+// NewCounter creates a new Counter
+//
+// this requires a name parameter
+// and optionally takes a couple of string directions, which
```
peterbourgon: Directions? Do you mean dimensions? Does this have to do with instances? Can you give a bit more description here?

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+	return &Counter{c}
+}
+
+// With is a no-op.
+func (c *Counter) With(labelValues ...string) metrics.Counter { return c }
+
+// Add implements Counter.
+// speed Counters only take int64
+// if it is important, instead use speed.SingletonMetric with DoubleType, CounterSemantics and OneUnit
+// but that will mean this will need a mutex, to be safe
```
peterbourgon: I don't understand this comment; I don't see any way to use speed.SingletonMetric from this package's API. Can you give the user more concrete directions?

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+func (c *Counter) Add(delta float64) { c.c.Inc(int64(delta)) }
+
+////////////////////////////////////////////////////////////////////////////////////////
```
peterbourgon: Remove this separator.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+
+// NewGauge creates a new Gauge
+//
+// this requires a name parameter
+// and optionally takes a couple of string directions, which
+// are directly passed to speed
+func (r *Reporter) NewGauge(name string, desc ...string) *Gauge {
+	g, err := speed.NewPCPGauge(0, name, desc...)
+	if err != nil {
+		panic(err)
```
peterbourgon: Same comment re: panic vs. returning an error.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+////////////////////////////////////////////////////////////////////////////////////////
+
+// Histogram wraps a PCP Histogram
+type Histogram struct {
+	h speed.Histogram
+}
+
+// NewHistogram creates a new Histogram
+// minimum observeable value is 0
+// maximum observeable value is 3600000000
```
peterbourgon: Shouldn't these values be parameters?

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// this requires a name parameter
+// and optionally takes a couple of string directions, which
+// are directly passed to speed
+func (r *Reporter) NewHistogram(name string, desc ...string) *Histogram {
+	h, err := speed.NewPCPHistogram(name, 0, 3600000000, 5, desc...)
+	if err != nil {
+		panic(err)
```
peterbourgon: Similar comment re: panic vs. returning an error.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// With is a no-op.
+func (h *Histogram) With(labelValues ...string) metrics.Histogram { return h }
+
+// Observe observes a value
+//
+// this converts float64 value to int64, as the Histogram in speed
+// is backed using codahale/hdrhistogram, which only observes int64 values
+func (h *Histogram) Observe(value float64) { h.h.MustRecord(int64(value)) }
```
peterbourgon: Simply converting to an int64 is going to be problematic, because it is idiomatic to Observe in a histogram in units of seconds, which typically are well less than 1. How should this be supported, do you think?

metrics/pcp/pcp.go
suyash: done

metrics/pcp/pcp.go
suyash: done.

metrics/pcp/pcp.go
suyash: using 'PCP installation' instead of backend

metrics/pcp/pcp.go
suyash: done

metrics/pcp/pcp.go
suyash: Every PCP metric supports a couple of description strings, but they are optional and can be a maximum of 256 characters.

suyash: The description strings are required at the time of construction and are reported with the metric.

suyash: required meaning that only if you want to, not like mandatory.

metrics/pcp/pcp.go
suyash: done.

metrics/pcp/pcp.go
suyash: advising using metrics.Gauge for float64 additions, as counters only support int64, that is a design decision from the Java project speed itself is based on.

metrics/pcp/pcp.go
suyash: done.

metrics/pcp/pcp.go
suyash: The histogram implementation isn't native and simply wraps a PCP metric around codahale/hdrhistogram, which only records int64 (https://godoc.org/github.com/codahale/hdrhistogram#Histogram.RecordValue), which in turn is based on a java version, which itself records only long (http://hdrhistogram.github.io/HdrHistogram/JavaDoc/org/HdrHistogram/AbstractHistogram.html#recordValue(long)).

Raw PCP metrics allow setting a unit (https://github.com/performancecopilot/speed#singletonmetric), and I have shorted it to OneUnit at most places as it represents Count unit and most of the times we just want to count stuff. But I can provide a units parameter in histogram, similar to Timer (https://github.com/performancecopilot/speed#timer) so users can customize the unit to Milliseconds, Microseconds, Nanoseconds and report int64.

The other option is to replace the underlying histogram with another implementation. I like hdrhistogram for its ideas on precision and reduced memory footprint, but am open to ideas.

suyash: so the normal tests were taking 1.6s but tests with `-race` enabled were taking 29s on my local machine and 38s on travis, so did a small profile

screenshot from 2016-09-20 23 03 06

from what I can see, the iterator next method on hdrhistogram seems slow, or maybe I am doing something wrong

metrics/pcp/pcp.go
peterbourgon: I'm rather asking you to clarify the commentary in the source, not in the GitHub PR :)

metrics/pcp/pcp.go
peterbourgon: Can you instead fix the comment to be more clear for a user who may not have this background?

metrics/pcp/pcp.go
peterbourgon: I'm fine with keeping HDRHistogram. I like the Unit idea, it should be settable at metric construction time. Testing with -race will always be much slower than otherwise; I wouldn't worry about profiling it.

suyash:
@peterbourgon sorry for the late update on this

- have added an additional unit parameter to the histogram construction, both here and to speed (https://godoc.org/github.com/performancecopilot/speed#NewPCPHistogram)

- as you said I have added some mode comments to all custom metric construction methods for speed

- however, the test time for histogram is still a concern to me, mainly I want to understand why it would take that long. Still investigating that, as well as trying to find other histogram implementations, will update this thread once I finalize something.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Reporter encapsulates a speed client.
+type Reporter struct {
+	c *speed.PCPClient
+}
+
+// NewReporter creates a new Reporter instance.
+// The first parameter is the application name and is used to create the speed client.
+// Hence it should be a valid speed parameter name and should not contain spaces or the path separator
+// for your operating system.
```
peterbourgon: Please wrap this and all comment blocks at 80 columns.

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,127 @@
...
+	r.c.MustRegister(c)
+	return &Counter{c}, nil
+}
+
+// With is a no-op.
+func (c *Counter) With(labelValues ...string) metrics.Counter { return c }
+
+// Add implements Counter.
+// speed Counters only take int64.
+// speed Gauges can take float64 and Add(float64) is implemented by metrics.Gauge.
```
peterbourgon: Please fix the capitalization in this and all comment blocks, and clarify the behavior.

```
// Add implements Counter. Speed counters only take int64, so 
// the delta is converted to an int64 before observation.
```

metrics/pcp/pcp.go
```diff
@@ -0,0 +1,127 @@
...
+	return &Histogram{h}, nil
+}
+
+// With is a no-op.
+func (h *Histogram) With(labelValues ...string) metrics.Histogram { return h }
+
+// Observe observes a value.
+//
+// this converts float64 value to int64, as the Histogram in speed
+// is backed using codahale/hdrhistogram, which only observes int64 values.
```
peterbourgon: Please mention something about the value being interpreted as the unit passed to the constructor.

peterbourgon:
Please don't waste time finding out why tests take longer to run with the race detector enabled. The race detector adds a lot of overhead by design.

peterbourgon:
With those final cleanups I've requested, I'm happy to merge. Thanks for taking the time!

suyash:
@peterbourgon as requested I have updated the comment blocks to be a maximum of 80 columns in length.

Moreover, I have generated the godoc for the package on my fork, which can be seen at (https://godoc.org/github.com/suyash/kit/metrics/pcp).

I am using the package name in small characters everywhere (`speed` instead of `Speed`), because of the package name.

W.e.f the histogram test time issue, the final call is yours, even if I change something in speed, my goal will be to keep things 100% backwards compatible. Right now whenever I can find the time, I am trying to understand the underlying working of hdrhistogram. In case there arises a need for a change in anything in this package, I can always open another PR.

Thanks for being patient with me, with the code reviews. The generated godoc LGTM, please let me know if you need anything else. I am using atom to check the line width, and except for line 106, everything looks to be inside.

peterbourgon:
Thanks for all of the hard work. I'll make a few other minor cleanups post-merge but this is great. 💯

suyash:
Thanks :sparkles:, you forgot to close the issue, will close it manually.

PR #367, opened 2016-10-02
metrics/influx: set the counter labels as tags

kpacha:
The current implementation of the influxdb metrics package is sending the labels as fields, so influxdb overrides all the different values you send, storing just the last received for each counter.

With this simple test you'll see my point:
```
func TestCounter_multipleTags(t *testing.T) {
	in := New(map[string]string{"a": "b"}, influxdb.BatchPointsConfig{}, log.NewNopLogger())
	counter := in.NewCounter("influx_counter")
	counter.Add(10)
	counter.With("error", "true").Add(1)
	counter.With("error", "false").Add(2)
	counter.Add(50)

	client := &bufWriter{}
	in.WriteTo(client)
	fmt.Println(client.buf.String())
}
```
The output:
```
influx_counter,a=b count=60 1475403189384692499
influx_counter,a=b count=1,error="true" 1475403189384692499
influx_counter,a=b count=2,error="false" 1475403189384692499
```
You can test how influxdb handles that by sending a POST as described in the documentation (https://docs.influxdata.com/influxdb/v1.0/guides/writing_data/#writing-data-using-the-http-api).

```
$ curl -i -XPOST 'http://192.168.99.100:8086/write?db=gokit' --data-binary 'influx_counter,a=b count=60 1475403189384692499
influx_counter,a=b count=1,error="true" 1475403189384692499
influx_counter,a=b count=2,error="false" 1475403189384692499'
HTTP/1.1 204 No Content
Content-Type: application/json
Request-Id: a34916d1-888d-11e6-8003-000000000000
X-Influxdb-Version: 1.0.0
Date: Sun, 02 Oct 2016 10:47:39 GMT

$ curl -GET 'http://192.168.99.100:8086/query?pretty=true' --data-urlencode "db=gokit" --data-urlencode "q=SELECT * FROM \"influx_counter\""
{
    "results": [
        {
            "series": [
                {
                    "name": "influx_counter",
                    "columns": [
                        "time",
                        "a",
                        "count",
                        "error"
                    ],
                    "values": [
                        [
                            "2016-10-02T10:13:09.384692499Z",
                            "b",
                            2,
                            "false"
                        ]
                    ]
                }
            ]
        }
    ]
}
```

Sending the labels as tags, influxdb is able to process them properly

```
$ curl -i -XPOST 'http://192.168.99.100:8086/write?db=gokit' --data-binary 'influx_counter,a=b count=60 1475403189384692499
influx_counter,a=b,error=true count=1 1475403189384692499
[...]

PR #369, opened 2016-10-02
metrics/influx: fix the package implementation

kpacha:
As reviewing the #367 I've found all the implementations of the metrics used the labels as fields, so I generalized the hack for gauges and histograms too.

Also, the histogram implementation wasn't aggregating the results, so they couldn't be stored properly in the backend (same problem as in #367, because influx accepts just one combination of fields for a given combination of tags and timestamp).

Finally, I added 3 tests where you can see how the messages are being sent over the wire.

### DIFF
--- a/metrics/influx/influx.go
+++ b/metrics/influx/influx.go
@@ -26,8 +26,8 @@ import (
 //
 // Influx tags are attached to the Influx object, can be given to each
 // metric at construction and can be updated anytime via With function. Influx fields
-// are mapped to Go kit label values directly by this collector. Actual metric values are provided as
-// fields with specific names depending on the metric.
+// are mapped to Go kit label values directly by this collector. Actual metric
+// values are provided as fields with specific names depending on the metric.
 //
 // All observations are collected in memory locally, and flushed on demand.
 type Influx struct {
--- a/metrics/influx/influx.go
+++ b/metrics/influx/influx.go
@@ -21,13 +21,12 @@ import (
 // one data point per flush, with a "count" field that reflects all adds since
 // the last flush. Gauges are modeled as a timeseries with one data point per
 // flush, with a "value" field that reflects the current state of the gauge.
-// Histograms are modeled as a timeseries with one data point per observation,
-// with a "value" field that reflects each observation; use e.g. the HISTOGRAM
-// aggregate function to compute histograms.
+// Histograms are modeled as a timeseries with one data point per combination of tags,
+// with a set of quantile fields that reflects the p50, p90, p95 & p99.
 //
-// Influx tags are immutable, attached to the Influx object, and given to each
-// metric at construction. Influx fields are mapped to Go kit label values, and
-// may be mutated via With functions. Actual metric values are provided as
+// Influx tags are attached to the Influx object, can be given to each
+// metric at construction and can be updated anytime via With function. Influx fields
+// are mapped to Go kit label values directly by this collector. Actual metric values are provided as
 // fields with specific names depending on the metric.
 //
 // All observations are collected in memory locally, and flushed on demand.
--- a/metrics/influx/influx.go
+++ b/metrics/influx/influx.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/generic"
 	"github.com/go-kit/kit/metrics/internal/lv"
 )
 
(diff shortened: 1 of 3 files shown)


metrics/influx/influx.go
```diff
@@ -123,10 +124,10 @@ func (in *Influx) WriteTo(w BatchPointsWriter) (err error) {
 	}
 
 	in.gauges.Reset().Walk(func(name string, lvs lv.LabelValues, values []float64) bool {
-		fields := fieldsFrom(lvs)
-		fields["value"] = last(values)
+		tags := mergeTags(in.tags, lvs)
 		var p *influxdb.Point
-		p, err = influxdb.NewPoint(name, in.tags, fields, now)
+		fields := map[string]interface{}{"value": last(values)}
+		p, err = influxdb.NewPoint(name, tags, fields, now)
```
peterbourgon: Can you please update the comment on the Influx struct to reflect the new role of fields and tags?

kpacha: done.

Let's see what do you think about it...

BTW, very nice project!!!

peterbourgon:
> execute the new influx tests as examples

Why? What is the advantage?

kpacha:
aren't the examples automatically linked to the documentation? since those tests were added in order to give a better explanation about how the influx metrics work, I believe adding them in the doc will help a lot.

metrics/influx/influx.go
```diff
@@ -20,13 +21,12 @@ import (
...
+// Histograms are modeled as a timeseries with one data point per combination of tags,
+// with a set of quantile fields that reflects the p50, p90, p95 & p99.
 //
-// Influx tags are immutable, attached to the Influx object, and given to each
-// metric at construction. Influx fields are mapped to Go kit label values, and
-// may be mutated via With functions. Actual metric values are provided as
+// Influx tags are attached to the Influx object, can be given to each
+// metric at construction and can be updated anytime via With function. Influx fields
+// are mapped to Go kit label values directly by this collector. Actual metric values are provided as
 // fields with specific names depending on the metric.
```
peterbourgon: Please wrap comments at 80col

metrics/influx/influx_test.go
```diff
@@ -31,6 +30,33 @@ func TestCounter(t *testing.T) {
 	}
 }
 
+func ExampleCounter() {
```
peterbourgon: Best not to mix Examples and real Tests in the same file. In the past I've stuck all the examples in an example_test.go file, can we do the same here?

metrics/influx/influx_test.go
```diff
@@ -31,6 +30,33 @@ func TestCounter(t *testing.T) {
...
+
+	if err := extractAndPrintMessage(expectedLines, client.buf.String()); err != nil {
+		fmt.Println(err.Error())
+	}
+
+	// Output:
+	// influx_counter,a=b count=60
+	// influx_counter,a=b,error=true count=1
+	// influx_counter,a=b,error=false count=2
+}
```
peterbourgon: Since this is now an Example, the test runner will take care of validating the output; there's no need for a separate step with the regexes. So,

```go
in := New(...)
counter := in.NewCounter("...")
// etc.

client := &bufWriter{}
in.WriteTo(client)
io.Copy(os.Stdout, client.buf)

// Output:
// foo
// bar
// baz
```

should be sufficient.

metrics/influx/influx_test.go
```diff
@@ -91,3 +174,15 @@ func (w *bufWriter) Write(bp influxdb.BatchPoints) error {
...
+	for _, pattern := range expected {
+		re := regexp.MustCompile(pattern)
+		match := re.FindStringSubmatch(msg)
+		if len(match) != 2 {
+			return fmt.Errorf("Pattern not found! {%s} [%s]: %v\n", pattern, msg, match)
+		}
+		fmt.Println(match[1])
+	}
+	return nil
+}
```
peterbourgon: re: above comments, I think this whole function can now be removed.

metrics/influx/influx_test.go
kpacha: the result is not deterministic because it adds a timestamp at the end of every line. That's why I used the regex extraction.

metrics/influx/influx_test.go
kpacha: ok. I'll move them to the `influxexample_test.go` file

metrics/influx/influx_test.go
peterbourgon: Then why is the example not failing? Is it not being run?

metrics/influx/influx_test.go
peterbourgon: Please just `example_test.go` in the appropriate directory.

metrics/influx/influx_test.go
kpacha: It doesn't fail because it extracts the timestamp before printing each line. so, instead of printing `influx_counter,a=b count=60 1234567891234567890` (sample timestamp), it just prints `influx_counter,a=b count=60`.

If any of the given patterns doesn't match, it also prints an error message and the example fails

peterbourgon: Ah! That's a bit tricky, but I understand now. Thanks.

metrics/influx/influx_test.go
peterbourgon: With the clarifications in the previous comment, this can stay :)

kpacha:
The branch now is including the changes you requested:
- Comment line fixed
- Examples in a dedicated file

Let me know if there is something else we could improve

peterbourgon:
Nice! Thank you very much for the contribution.

PR #370, opened 2016-10-02
Retry callbacks passing errors.

### LINKED ISSUE #341: Add custom error behavior for lb.Retry

rossmcf:
Hopefully this should resolve #341.

I wasn't sure what the protocol should be for this, but I noticed things had gone a bit quiet on the other PR for this issue. I've pulled @morganhein's changes and made a few tweaks in line with @peterbourgon's feedback.

Callbacks now accept an `error`, and an additional test is added to demonstrate that errors are passed unchanged from the endpoint.

Please let me know if anything needs tweaked, or if I should submit this change differently.

### DIFF
--- a/sd/lb/retry.go
+++ b/sd/lb/retry.go
@@ -44,7 +44,9 @@ func maxRetries(max int) Callback {
 // whichever comes first.
 func RetryWithCallback(timeout time.Duration, b Balancer, cb Callback) endpoint.Endpoint {
 	if cb == nil {
-		panic("nil Callback")
+		cb = func(n int, err error) (bool, error) {
+			return true, nil
+		}
 	}
 	if b == nil {
 		panic("nil Balancer")
--- a/sd/lb/retry.go
+++ b/sd/lb/retry.go
@@ -14,8 +14,8 @@ import (
 // encountered. Should return whether the Retry function should continue trying,
 // and a custom error message if desired. The error message may be nil, but a
 // true/false is always expected. In all cases if the error message is supplied,
-// thecurrent error will be replaced.
-type Callback func(n int, err error) (cont bool, cbErr error)
+// the current error will be replaced.
+type Callback func(n int, received error) (keepTrying bool, cbErr error)
 
 // Retry wraps a service load balancer and returns an endpoint oriented load
 // balancer for the specified service method.
--- a/sd/lb/retry.go
+++ b/sd/lb/retry.go
@@ -10,7 +10,7 @@ import (
 	"github.com/go-kit/kit/endpoint"
 )
 
-// Callback is a function that indicates the current attempt count and the error
+// Callback is a function that is given the current attempt count and the error
 // encountered. Should return whether the Retry function should continue trying,
 // and a custom error message if desired. The error message may be nil, but a
 // true/false is always expected. In all cases if the error message is supplied,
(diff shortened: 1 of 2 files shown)


sd/lb/retry.go
```diff
@@ -10,12 +10,32 @@ import (
...
+// RetryWithCallback wraps a service load balancer and returns an endpoint oriented load
+// balancer for the specified service method.
+// Requests to the endpoint will be automatically load balanced via the load
+// balancer. Requests that return errors will be retried until they succeed,
+// up to max times, until the callback returns false, or until the timeout is elapsed,
+// whichever comes first.
+func RetryWithCallback(max int, timeout time.Duration, b Balancer, cb Callback) endpoint.Endpoint {
+	if cb == nil {
+		panic("nil Callback")
+	}
```
peterbourgon: We can be safer, here. If there's a nil callback, we can substitute a no-op or always-pass implementation instead.

sd/lb/retry.go
```diff
@@ -48,7 +68,18 @@ func Retry(max int, timeout time.Duration, b Balancer) endpoint.Endpoint {
...
+					if cbErr == nil {
+						return nil, fmt.Errorf("retry attempts exceeded (%s)", strings.Join(a, "; "))
+					}
+					return nil, cbErr
+				}
+				currentErr := err.Error()
+				if cbErr != nil {
+					currentErr = cbErr.Error()
+				}
+				a = append(a, currentErr)
```
peterbourgon: This is super messy, especially regarding the final error returned to the caller, which at this point I think deserves to be lifted into its own type, with a slice of underlying errors as a member. But, I think a lot of the messiness may go away if you're able to implement all of this logic in the callback as I wanted above.

sd/lb/retry.go
```diff
@@ -10,12 +10,32 @@ import (
 	"github.com/go-kit/kit/endpoint"
 )
 
+// Callback function that indicates the current attempt count and the error encountered.
+//  Should return whether the Retry function should continue trying, and a custom
+//  error message if desired. The error message may be nil, but a true/false
+//  is always expected. In all cases if the error message is supplied, the
+//  current error will be replaced.
+type Callback func(n int, err error) (cont bool, cbErr error)
```
rossmcf: Updated to make the intention clearer.

sd/lb/retry.go
rossmcf: Substituting a no-op now, with a test.

sd/lb/retry.go
```diff
@@ -10,12 +10,32 @@ import (
...
+
 // Retry wraps a service load balancer and returns an endpoint oriented load
 // balancer for the specified service method.
 // Requests to the endpoint will be automatically load balanced via the load
 // balancer. Requests that return errors will be retried until they succeed,
 // up to max times, or until the timeout is elapsed, whichever comes first.
 func Retry(max int, timeout time.Duration, b Balancer) endpoint.Endpoint {
+	return RetryWithCallback(max, timeout, b, func(c int, err error) (bool, error) { return true, nil })
+}
```
rossmcf: I've split the max retry logic out into the callback. There's a helper function for generating a max retries callback if that's all you want to do, and `Retry()` passes one of those on to `RetryWithCallback()` now.

I'm not sure if or how to tackle handling timeouts in the callback. I think it's reasonable to keep those two jobs separate. To me, I'm happy with the timeout in the `select` inside `RetryWithCallback()`. What do you reckon?

peterbourgon:
See https://github.com/rossmcf/kit/pull/1

peterbourgon:
Ace!

PR #371, opened 2016-10-04
Remove service leftovers in consul integration test

### LINKED ISSUE #366: File  sd/consul/integration_test.go references nonexistent import path
File [_sd/consul/integration_test.go_](https://github.com/go-kit/kit/blob/9a19822c46c494bdac0944b8e5da6fc8684536e7/sd/consul/integration_test.go#L12) makes reference to a nonexistent import path: ""github.com/go-kit/kit/service”. I noted the problem [here](https://github.com/go-kit/kit/commit/9a19822c46c494bdac0944b8e5da6fc8684536e7#commitcomment-19240903), but figured it warrants an issue.

The file is guarded from building with [the “integration” constraint](https://github.com/go-kit/kit/blob/9a19822c46c494bdac0944b8e5da6fc8684536e7/sd/consul/integration_test.go#L1).
Did this _service_ package move?

I noticed the problem when running _glide update_; it tries to follow all of these import paths, and it gets stuck here.

xla:
During a brief and forgotten period there was the Service abstraction which was used in the integration test of the Consul sd package. As it is guarded with a build tag it likely was missed, this change-set addresses this and removes all traces of the forgotten abstraction.

Fixes #366

### DIFF
--- a/sd/consul/integration_test.go
+++ b/sd/consul/integration_test.go
@@ -8,8 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
-	"github.com/go-kit/kit/service"
 	stdconsul "github.com/hashicorp/consul/api"
 )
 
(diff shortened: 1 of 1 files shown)


seh:
I confirmed that Glide can now scan the "github.com/go-kit/kit/sd/consul" import path without running into the previous error with not being able to find the "service" package. Thank you for fixing this.

peterbourgon:
😎

Issue #328, opened 2016-07-20
Actual integration testing on CI


peterbourgon:
Go kit includes some packages which are designed to interact with popular open source software. They include

- ~~sd/etcd — with CoreOS etcd~~ _DONE_
- ~~sd/consul — with HashiCorp Consul~~ _DONE_
- ~~sd/zk — with Apache ZooKeeper~~ _DONE_
- tracing/opentracing — with AppDash, Lightstep, and Zipkin

Many of those packages include proper integration tests that connect to the relevant software. (Some do not.) In this ticket we should

- Provide simple, reliable integration tests for the packages above that don't have them; and
- Configure one or more of our CI servers (Travis, Circle, and/or Drone) to run those tests.

Writing the tests should involve only basic connectivity and expected functionality. No need to get clever checking for obscure edge conditions, the big win here is making sure we don't regress. 

Running the tests should involve absolutely minimal changes to this repo: no complex setup or teardown scripts embedded into circle.yml, please! Leverage Docker images where appropriate, and go with standalone or single-node test-cluster setups wherever possible.

xmattstrongx:
I can work on some of these.

fvosberg:
Hi Peter,

I would love to help you with this issue, but I don't really get what I can do. As I've seen there are basic tests in the packages. Is there a specific case, which you are missing? Or should I configure running tests on a specific server? 

Thanks in advance

peterbourgon:
Hi @fvosberg, thanks for reaching out! You're right, the state of the actual project has accelerated away from this issue, at the moment we have good integration tests implemented for all of sd/{etcd,consul,zk} and [running on Circle](https://circleci.com/gh/go-kit/kit/1115). I'll update the original issue to reflect that.

The only thing left is the tracing stuff, which is tricky! It's slightly less necessary, because OpenTracing itself encapsulates the vast majority of the hard stuff, and it's already well-tested internally. It's also a bit less obvious, because a useful integration test from Go kit's perspective would involve quite a lot of work—though I think it could be quite illuminating!

Off the cuff, here's what I think would need to happen for the integration test, in rough strokes:

- Pick an OpenTracing-compatible system to test against
- Create 2 simple microservices, A and B, fully instrumented for that tracing system
- Make sure that A depends on B, i.e. A calls out to B in one of its methods
- In the integration test, boot up all 3 systems and ensure connectivity
- Make a request to A
- Wait some time, and then query the OpenTracing system for the trace
[...]

PR #372, opened 2016-10-06
Run integration tests on circleci

rodrigosaito:
This PR is to address the following issue: https://github.com/go-kit/kit/issues/328

This is still working in progress, I'd be happy to finish this if it's looking good.

There are some workarounds to fix the GOPATH in circle.yml, it might not be necessary on the original repo.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,18 @@
+machine:
+  pre:
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
+  services:
+    - docker
+
+dependencies:
+  pre:
+    - sudo pip install docker-compose
+    - docker-compose -f docker-compose-integration.yml up -d --force-recreate
+
+test:
+  override:
+    - go test -v -race -tags integration ./...:
+        environment:
+          ETCD_ADDR: http://localhost:2379
+          CONSUL_ADDR: localhost:8500
+          ZK_ADDR: localhost:2181
--- a/sd/consul/integration_test.go
+++ b/sd/consul/integration_test.go
@@ -14,11 +14,9 @@ import (
 )
 
 func TestIntegration(t *testing.T) {
-	// Connect to Consul.
-	// docker run -p 8500:8500 progrium/consul -server -bootstrap
-	consulAddr := os.Getenv("CONSUL_ADDRESS")
+	consulAddr := os.Getenv("CONSUL_ADDR")
 	if consulAddr == "" {
-		t.Fatal("CONSUL_ADDRESS is not set")
+		t.Fatal("CONSUL_ADDR is not set")
 	}
 	stdClient, err := stdconsul.NewClient(&stdconsul.Config{
 		Address: consulAddr,
--- a/sd/zk/integration_test.go
+++ b/sd/zk/integration_test.go
@@ -4,8 +4,7 @@ package zk
 
 import (
 	"bytes"
-	"flag"
-	"fmt"
+	"log"
 	"os"
 	"testing"
 	"time"
@@ -18,21 +17,11 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	flag.Parse()
-
-	fmt.Println("Starting ZooKeeper server...")
-
-	ts, err := stdzk.StartTestCluster(1, nil, nil)
-	if err != nil {
-		fmt.Printf("ZooKeeper server error: %v\n", err)
-		os.Exit(1)
+	zkAddr := os.Getenv("ZK_ADDR")
+	if zkAddr == "" {
+		log.Fatal("ZK_ADDR is not set")
 	}
-
-	host = []string{fmt.Sprintf("localhost:%d", ts.Servers[0].Port)}
-	code := m.Run()
-
-	ts.Stop()
-	os.Exit(code)
+	host = []string{zkAddr}
 }
 
 func TestCreateParentNodesOnServer(t *testing.T) {
--- a/docker-compose-integration.yml
+++ b/docker-compose-integration.yml
@@ -1,0 +1,16 @@
+version: '2'
+services:
+  etcd:
+    image: quay.io/coreos/etcd
+    ports:
+      - "2379:2379"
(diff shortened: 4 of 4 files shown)


sd/consul/integration_test.go
```diff
@@ -1,4 +1,4 @@
-// +build integration
+// +build integration-ignore
```
peterbourgon: What does this do?

circle.yml
```diff
@@ -0,0 +1,19 @@
...
+  pre:
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
+  services:
+    - docker
+
+dependencies:
+  pre:
+    - sudo pip install docker-compose
+    - docker-compose -f docker-compose-integration.yml up -d --force-recreate
+    - mkdir -p ${HOME}/.go_workspace/src/github.com/go-kit/ && ln -sf ${HOME}/kit ${HOME}/.go_workspace/src/github.com/go-kit/
```
peterbourgon: Why is this necessary?

circle.yml
```diff
@@ -0,0 +1,19 @@
...
+  override:
+    - cd ${HOME}/.go_workspace/src/github.com/go-kit/kit && go get -t -d -v ./...
```
peterbourgon: Shouldn't it be sufficient to `go get -d -t -v ./...` without the cd?

sd/consul/integration_test.go
rodrigosaito: just ignoring those tests for now, will remove that once everything is done.

circle.yml
rodrigosaito: yes, probably with go-kit/kit repo, not with my fork.

before sending the final PR I will remove all this workaround that I need to run builds on circle from my fork.

peterbourgon:
Ping? :)

rodrigosaito:
@peterbourgon sorry about the delay.

so should I keep going with this one?

peterbourgon:
No worries at all! Yes, please, if you can; otherwise I'll take it over :)

rodrigosaito:
Yes, I can and will be happy to help. Will try to finish this by the weekend.

peterbourgon:
Friendly ping :) Any update here?

rodrigosaito:
I am being busy lately, will try to find some time to work on this in the next few days.

rodrigosaito:
@peterbourgon done!

the tests are failing now, probably because circleci is configured to use the old image (12.04) instead of the new one (14.04). If you change that, then the build should pass. Another solution is to stop zookeeper on circle.yml, I can do that if you prefer this solution.

peterbourgon:
Well, the error is pretty explicit; the docker-compose up fails with

```
ERROR: for zk  Cannot start service zk: failed to create endpoint kit_zk_1 on network kit_default: Error starting userland proxy: listen tcp 0.0.0.0:2181: bind: address already in use
```

When I try to do the same thing locally, I get a similar error:

```
ERROR: for etcd  Cannot start service etcd: driver failed programming external connectivity on endpoint kit_etcd_1 (0577dd66f5552c29e6bf8fe0c882c9cfcf4515215a1d2bdf19650df0b44dd90e): Error starting userland proxy: listen tcp 0.0.0.0:2379: bind: address already in use
```

I think there must be some error in the way you've got the docker-compose file orchestrating things. Does this work for you locally?

rodrigosaito:
The error happens on circle ci, probably because it's running on 12.04 image which has the zookeeper up and running(https://circleci.com/docs/build-image-precise/#databases), conflicting the ports as the docker-compose is using default ports for every key store (consul, etcd and zk). I tested on circle ci using 14.04 (https://circleci.com/docs/build-image-trusty/#databases) which doesn't have a zookeeper running.

On your machine you might have a etcd running which will also conflict the ports.

Yes it works locally for me and when I run on circle ci from my repo.

I can change to use random ports if you prefer.

peterbourgon:
Ah, OK! Thanks for the more lengthy explanation. You're probably right; I was running that in a minikube (Kubernetes) VM which probably has etcd listening on the default port. I'll poke around a bit more.

rodrigosaito:
hey @peterbourgon any update on this? do you need some help?

peterbourgon:
Thanks for the poke :) I got it working locally. A few changes I'd like to see to the PR itself incoming...

circle.yml
```diff
@@ -0,0 +1,18 @@
...
+    - sudo pip install docker-compose
+    - docker-compose -f docker-compose-integration.yml up -d --force-recreate
+
+test:
+  override:
+    - go test -v -race -tags integration ./...:
+        environment:
+          ETCD_ADDR: http://localhost:2379
+          CONSUL_ADDR: localhost:8500
+          ZK_ADDR: localhost:2181
```
peterbourgon: Can you make whatever changes are necessary to get this working on Circle in the file, or do I need to change something in the repo settings?

sd/zk/integration_test.go
```diff
@@ -17,22 +16,12 @@ var (
...
-func TestMain(m *testing.M) {
-	flag.Parse()
-
-	fmt.Println("Starting ZooKeeper server...")
-
-	ts, err := stdzk.StartTestCluster(1, nil, nil)
-	if err != nil {
-		fmt.Printf("ZooKeeper server error: %v\n", err)
-		os.Exit(1)
+func init() {
```
peterbourgon: Why was this moved out of TestMain and into init?

rodrigosaito: removed the code that was starting zookeeper (this previously needed the zk jar file somewhere). Not really sure the difference of init and TestMain. Would you prefer to have this code on TestMain instead?

rodrigosaito: ok, got it. Will do that later today.

rodrigosaito: done!

circle.yml
peterbourgon: Done. Takes a fresh commit to test; we'll do it on master! ;)

peterbourgon:
Brilliant! Works locally and I'm happy with the delta. Thanks very much for the contribution! 👍

PR #374, opened 2016-10-10
fixed import path of appdash to official one

basvanbeek:
Minor mistake I stumbled on when creating a pull request for AppDash handling latest OpenAPI changes.


peterbourgon:
:\ Not a fan of these vanity URLs; really wish folks would stop using them. Of course that's not our call to make, and thank you for the patch!

basvanbeek:
Totally agree and it made it annoying to work. AppDash also has an internal package so working from the more obvious github.com package path made it fail to build / test

PR #385, opened 2016-11-03
JWT/auth typo

groob:
I fixed a typo I spotted while reading through the package docs. 
I also added punctuation in the doc comments to keep consistent with the rest of the code.


peterbourgon:
Yay! Thanks.

PR #388, opened 2016-11-07
shipping: Update with upstream changes

marcusolsson:
This updates the shipping examples with some changes made upstreams in [goddd](https://github.com/marcusolsson/goddd).

- Variable renaming: Some names smelled of Java and have been improved(?).
- `repository` package renamed to `inmem` to better reflect what it provides.
- The `Location` struct is a DDD Entity but was used as a Value Object. Other domain objects now refer to it by identity (UNLocode) instead of value. 
- Added `mock` package with mock implementations of the repositories. Questionable if this should be added to the example, since the tests have been excluded in the shipping example. I think it's still interesting to show how it compares to the `inmem` repository, but I have no problems leaving it out.

P.S. Bulk changes are bad and I should feel bad. Let me know if you want me to split it.


peterbourgon:
Can you rebase on master to get the tests green?

marcusolsson:
Done.

PR #392, opened 2016-11-14
fix(generic.go): add defer to ApproximateMovingAverage mutext

mishudark:
Currently there is a `h.mtx.RLock()` following by `h.mtx.RUnlock()`, the unlock should be a defer `defer h.mtx.RUnlock()`


peterbourgon:
Wow, thank you!

PR #396, opened 2016-11-17
Correct typ-o in addsvc for concatEndpoint

patterns:
There is a small typ-o in the addsvc example http client. The concatEndpoint block shows two lines with 'sumEndpoint'.


basvanbeek:
thanks!

PR #398, opened 2016-11-22
Reset bold attribute

serge-v:


log/term/colorlogger.go
```diff
@@ -42,15 +42,15 @@ const (
...
 	fgColorBytes    [][]byte
 	bgColorBytes    [][]byte
 )
 
 func init() {
 	// Default
-	fgColorBytes = append(fgColorBytes, []byte("\x1b[39m"))
-	bgColorBytes = append(bgColorBytes, []byte("\x1b[49m"))
+	fgColorBytes = append(fgColorBytes, []byte("\x1b[39;22m"))
+	bgColorBytes = append(bgColorBytes, []byte("\x1b[49;22m"))
```
ChrisHines: Is this change required? I believe we maintain the invariant that `resetColorBytes` is emitted at the end of each log record, which implies that no color codes are in effect at the point that these sequences are applied, making the additional `22m` code redundant with the change on line 45. Thoughts?

serge-v: you are right, it's not needed. Only in resetColorBytes.

serge-v:
will redo

Issue #397, opened 2016-11-22
Reset bold font


serge-v:
https://github.com/go-kit/kit/blob/master/log/term/colorlogger.go#L45 needs code 22 also to reset bold font `resetColorBytes = []byte("\x1b[39;49;22m")`

ChrisHines:
Thanks for the suggestion. The color logger (as currently written) doesn't have any code to enable bold font. Do you still think code 22 is necessary? Do you have example code that shows when this addition is needed?

serge-v:
Actually this loop https://github.com/go-kit/kit/blob/master/log/term/colorlogger.go#L61 enables bold attribute for fonts (";1m"). So it should be reset.

ChrisHines:
OK. Thanks for clarifying. Would you like to submit a PR for this fix?

PR #399, opened 2016-11-22
Reset bold attribute

serge-v:


ChrisHines:
@serge-v Thank you for the fix!

ChrisHines:
Fixes #397.

PR #406, opened 2016-12-09
Update the Histogram example to reflect API changes

Artmann:
`prometheus.NewSummary()` takes a ` *"github.com/prometheus/client_golang/prometheus".SummaryVec` as argument. `prometheus.NewSummaryFrom()` has the behaivour described in the example.

- Update the Histogram example so it uses `prometheus.NewSummaryFrom()` instead of `prometheus.NewSummary()`.

### DIFF
--- a/metrics/README.md
+++ b/metrics/README.md
@@ -49,7 +49,7 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


peterbourgon:
Ah, thank you! :)

PR #409, opened 2016-12-13
fix a example bug

shanshanpt:
Fix a example bug.  Then we can route the request like "curl -X POST --data "{\"A\":1, \"B\":2}" 127.0.0.1:8000/addsvc/sum".

### DIFF
--- a/examples/addsvc/transport_http.go
+++ b/examples/addsvc/transport_http.go
@@ -26,14 +26,14 @@ func MakeHTTPHandler(ctx context.Context, endpoints Endpoints, tracer stdopentra
 		httptransport.ServerErrorLogger(logger),
 	}
 	m := http.NewServeMux()
-	m.Handle("/sum", httptransport.NewServer(
+	m.Handle("/addsvc/sum", httptransport.NewServer(
 		ctx,
 		endpoints.SumEndpoint,
 		DecodeHTTPSumRequest,
 		EncodeHTTPGenericResponse,
 		append(options, httptransport.ServerBefore(opentracing.FromHTTPRequest(tracer, "Sum", logger)))...,
 	))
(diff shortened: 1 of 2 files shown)


peterbourgon:
No, sorry — the addsvc itself should not prefix its own paths with `addsvc/`. What bug were you fixing, exactly?

shanshanpt:
In old example code, when I execute "curl -X POST --data "{"A":1, "B":2}" 127.0.0.1:8000/addsvc/sum", the response is "404 not found".  Then it's OK after I updated the code. I don't know if I execute the wrong curl-command. Am I misunderstand the example?  Thanks.  :)

peterbourgon:
What happens if you make only the change in examples/apigateway/main.go, and throw away the other change?

shanshanpt:
Also "404". I try it just now.

peterbourgon:
Whoops! There was a bug. Here is the fix:

```diff
diff --git a/examples/apigateway/main.go b/examples/apigateway/main.go
index 01367ec..b357e74 100644
--- a/examples/apigateway/main.go
+++ b/examples/apigateway/main.go
@@ -104,7 +104,7 @@ func main() {
                // HTTP handler, and just install it under a particular path prefix in
                // our router.

-               r.PathPrefix("addsvc/").Handler(addsvc.MakeHTTPHandler(ctx, endpoints, tracer, logger))
+               r.PathPrefix("/addsvc").Handler(http.StripPrefix("/addsvc", addsvc.MakeHTTPHandler(ctx, endpoints, tracer, logger)))
        }

        // stringsvc routes.
```

I've made the fix in master. Can you pull and try again?

peterbourgon:
I'm pretty sure this works, so I'll close the PR. Let me know if it's still broken for you :)

Issue #343, opened 2016-08-17
SD - registration doesn't expire ?


pdaures:
Hi, 
A stupid question : I checked the sd/etcd client and if I understood correctly, the registration doesn't expire. So, if the machine goes down violently, we won't run Deregister() and the service will still be registered.

Wouldn't it be better to have a TTL on the registration, and call it regularly ? So if the service goes down and cannot call Deregister, etcd will remove the service by itself and it won't appear anymore to the other services.

peterbourgon:
Yes, your observation is correct. TTL is typically implemented by the service discovery system itself, in the form of regular health checks to registered services. If you think about it, this is strictly superior to heartbeating: less work overall, and a more accurate result.

Of course it's totally possible to do what you suggest. The right way, I think, would be to provide another implementation of Registrar for the systems you want to support, which performs this heartbeating under the hood. Which SD system do you have in mind? Would you be up for making a PR for that?

pdaures:
Thx for answering quickly.
I had in mind etcd.

I don't think etcd has any type of built-in service health checks. 
I know we can register health check endpoints to Consul, but I haven't seen it in go-kit.

So, I guess for ETCD we would have to create a Registrar that calls ETCD every X duration.
For Consul, I don't know what would be the correct way of doing this, providing an endpoint ?

I don't have time this week to do this, but I can probably find some time next week.

peterbourgon:
For etcd, I guess you will create a special registrar implementation that creates a key with a short TTL, and regularly refreshes the lease with a goroutine. For Consul, there's no single answer. Probably the best way would be to mount a /health endpoint into your service and configure Consul to ping it; that's more about configuring Consul than adding things to Go kit, though, I think.

PR #413, opened 2016-12-14
Add TTL for etcd sd

buptmiao:
Add TTL for registering keys to etcd, with backward compatibility。  

relevant issue: [#343](https://github.com/go-kit/kit/issues/343)

Users can set TTL options like this:
```go
       registrar := etcd.NewRegistrar(sdClient, etcd.Service{
		Key:   key,
		Value: value,
		TTL: etcd.NewTTLOption(time.Second, time.Second * 3),
	}, log.NewNopLogger())
```

### DIFF
--- a/sd/etcd/client.go
+++ b/sd/etcd/client.go
@@ -156,6 +156,10 @@ func (c *client) Register(s Service) error {
 	if s.Value == "" {
 		return ErrNoValue
 	}
+	if s.TTL != nil {
+		_, err := c.keysAPI.Set(c.ctx, s.Key, s.Value, &etcd.SetOptions{PrevExist: etcd.PrevIgnore, TTL: s.TTL.TTL})
+		return err
+	}
 	_, err := c.keysAPI.Create(c.ctx, s.Key, s.Value)
 	return err
 }
--- a/sd/etcd/client.go
+++ b/sd/etcd/client.go
@@ -156,11 +156,12 @@ func (c *client) Register(s Service) error {
 	if s.Value == "" {
 		return ErrNoValue
 	}
+	var err error
 	if s.TTL != nil {
-		_, err := c.keysAPI.Set(c.ctx, s.Key, s.Value, &etcd.SetOptions{PrevExist: etcd.PrevIgnore, TTL: s.TTL.TTL})
-		return err
+		_, err = c.keysAPI.Set(c.ctx, s.Key, s.Value, &etcd.SetOptions{PrevExist: etcd.PrevIgnore, TTL: s.TTL.ttl})
+	} else {
+		_, err = c.keysAPI.Create(c.ctx, s.Key, s.Value)
 	}
-	_, err := c.keysAPI.Create(c.ctx, s.Key, s.Value)
 	return err
 }
 
--- a/sd/etcd/registrar.go
+++ b/sd/etcd/registrar.go
@@ -4,6 +4,11 @@ import (
 	etcd "github.com/coreos/etcd/client"
 
 	"github.com/go-kit/kit/log"
+	"time"
+)
+
+const (
+	MinHeartBeatTime = time.Millisecond * 500
 )
 
 // Registrar registers service instance liveness information to etcd.
@@ -11,6 +16,7 @@ type Registrar struct {
 	client  Client
 	service Service
 	logger  log.Logger
+	quit    chan struct{}
 }
 
 // Service holds the instance identifying data you want to publish to etcd. Key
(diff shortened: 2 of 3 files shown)


sd/etcd/registrar.go
```diff
@@ -4,13 +4,19 @@ import (
 	etcd "github.com/coreos/etcd/client"
 
 	"github.com/go-kit/kit/log"
+	"time"
```
peterbourgon: This should be grouped with the other stdlib imports.

sd/etcd/registrar.go
```diff
@@ -4,13 +4,19 @@ import (
...
+)
+
+const (
+	MinHeartBeatTime = time.Millisecond * 500
 )
```
peterbourgon: This constant should either be made private (I think that's preferable) or given a doc comment.

sd/etcd/registrar.go
```diff
@@ -19,9 +25,27 @@ type Registrar struct {
 type Service struct {
 	Key           string // unique key, e.g. "/service/foobar/1.2.3.4:8080"
 	Value         string // returned to subscribers, e.g. "http://1.2.3.4:8080"
+	TTL           *TTLOption
 	DeleteOptions *etcd.DeleteOptions
 }
 
+// TTLOption allow setting a key with a TTL, and regularly refreshes the lease with a goroutine
```
peterbourgon: This comment describes the mechanics of what happens, but not what benefit it provides. Can you amend the comment to tell the user in which circumstances they would want to use this option? Also, please wrap the comment at 80 columns, and use proper punctuation (i.e. add a period to the end of the sentence).

sd/etcd/registrar.go
```diff
@@ -19,9 +25,27 @@ type Registrar struct {
...
+type TTLOption struct {
+	Heartbeat time.Duration
+	TTL       time.Duration
+}
```
peterbourgon: Given that a TTLOption is constructed with a constructor, I don't think either of these fields needs to be exported. Is that correct?

sd/etcd/registrar.go
```diff
@@ -19,9 +25,27 @@ type Registrar struct {
...
+
+// NewTTLOption returns a TTLOption
+func NewTTLOption(heartbeat, ttl time.Duration) *TTLOption {
```
peterbourgon: The doc comment here is a little sparse. Could it please describe what each of the parameters are, and some good default values? It would also be good to describe the bounds-checking logic implemented in the function body.

sd/etcd/registrar.go
```diff
@@ -19,9 +25,27 @@ type Registrar struct {
...
+	if heartbeat <= MinHeartBeatTime {
+		heartbeat = MinHeartBeatTime
+	}
+	if ttl <= heartbeat {
+		ttl = heartbeat * 3
+	}
+	return &TTLOption{heartbeat, ttl}
```
peterbourgon: Given that both heartbeat and ttl have the same type, it's easy to flip them around. It's a good idea to use struct literal initialization, e.g.

```go
return &TTLOption{
    heartbeat: heartbeat,
    ttl:       ttl,
}
```

sd/etcd/registrar.go
```diff
@@ -43,6 +67,22 @@ func (r *Registrar) Register() {
 	} else {
 		r.logger.Log("action", "register")
 	}
+	if r.service.TTL == nil {
+		return
+	}
```
peterbourgon: This says "bail early if there's no configured TTL" — but I think it's much more clear if the logic is inverted, i.e. "if there _is_ a configured TTL, create the quit chan and launch the maintenance loop".

sd/etcd/registrar.go
```diff
@@ -43,6 +67,22 @@ func (r *Registrar) Register() {
...
+			select {
+			case <-r.quit:
+				return
+			case <-time.After(r.service.TTL.Heartbeat):
+				if err := r.client.Register(r.service); err != nil {
+					r.logger.Log("err", err)
+				}
+			}
+		}
+	}()
```
peterbourgon: Please extract this to a loop method on the registrar type.

sd/etcd/registrar.go
```diff
@@ -43,6 +67,22 @@ func (r *Registrar) Register() {
...
+	if r.service.TTL == nil {
+		return
+	}
+	r.quit = make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-r.quit:
+				return
+			case <-time.After(r.service.TTL.Heartbeat):
```
peterbourgon: This will generate a lot of unnecessary garbage. It's better to use a time.Ticker, constructed outside of the for loop.

sd/etcd/registrar.go
```diff
@@ -43,6 +67,22 @@ func (r *Registrar) Register() {
 	} else {
 		r.logger.Log("action", "register")
 	}
+	if r.service.TTL == nil {
+		return
+	}
+	r.quit = make(chan struct{})
```
peterbourgon: This mutates registrar state, which makes it unsafe for concurrent access from multiple goroutines. I think the registrar needs its internal state protected with a mutex, which should be taken with every public method.

sd/etcd/registrar.go
```diff
@@ -53,4 +93,7 @@ func (r *Registrar) Deregister() {
 	} else {
 		r.logger.Log("action", "deregister")
 	}
+	if r.quit != nil {
+		close(r.quit)
+	}
```
peterbourgon: It looks like r.quit should also be reset to nil.

sd/etcd/client.go
```diff
@@ -156,6 +156,10 @@ func (c *client) Register(s Service) error {
 	if s.Value == "" {
 		return ErrNoValue
 	}
+	if s.TTL != nil {
+		_, err := c.keysAPI.Set(c.ctx, s.Key, s.Value, &etcd.SetOptions{PrevExist: etcd.PrevIgnore, TTL: s.TTL.TTL})
+		return err
+	}
 	_, err := c.keysAPI.Create(c.ctx, s.Key, s.Value)
 	return err
```
peterbourgon: Hmm. What do you think about

```go
var err error
if s.TTL != nil {
    _, err = c.keysAPI.Set(...)
} else {
    _, err = c.keysAPI.Create(...)
}
return err
```

buptmiao:
Thank you for your attention and advice.

peterbourgon:
Amazing, thank you! There are a couple of other small things, but I'll make the changes in master after the merge.

PR #417, opened 2016-12-17
metrics: improve README

### LINKED ISSUE #411: Counter/Gauge/Histogram.With not documented clearly
Coming to this documentation with no metrics exposure before this, it's not clear what label values are. They kinda seemed like a namespace prefix at first, but that turns out not to be correct. One or two lines would save newbies a lot of time.

peterbourgon:
Closes #411 — I hope!

### DIFF
--- a/metrics/doc.go
+++ b/metrics/doc.go
@@ -1,17 +1,35 @@
-// Package metrics provides a framework for application instrumentation. All
-// metrics are safe for concurrent use. Considerable design influence has been
-// taken from https://github.com/codahale/metrics and https://prometheus.io.
+// Package metrics provides a framework for application instrumentation. It's
+// primarily designed to help you get started with good and robust
+// instrumentation, and to help you migrate from a less-capable system like
+// Graphite to a more-capable system like Prometheus. If your organization has
+// already standardized on an instrumentation system like Prometheus, and has no
+// plans to change, it probably makes sense to use that system's instrumentation
+// library directly.
 //
-// This package contains the common interfaces. Your code should take these
-// interfaces as parameters. Implementations are provided for different
-// instrumentation systems in the various subdirectories.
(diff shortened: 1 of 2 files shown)


metrics/doc.go
```diff
@@ -1,17 +1,35 @@
...
 //
-// This package contains the common interfaces. Your code should take these
-// interfaces as parameters. Implementations are provided for different
-// instrumentation systems in the various subdirectories.
+// This package provides three core metric abstractions (Counter, Gauge, and
+// Histogram) and implementations for almost all common instrumentation
+// backends. Each metric has an observation method (Add, Set, or Observe,
+// respectively) used to record values, and a With method to "scope" the
+// observation by various parameters. For example, you might have a Histogram to
+// record request durations, parameterized by the method that's being called.
```
willfaught: The label names seem to be implementation-dependent as well (please correct me if I'm wrong!). It seems, IIUC, that for Prometheus, you should use underscores as separators, whereas for Statsd you should use periods. I imagine it's possible for another implementation that actually uses With to expect periods too. So for gokit, the label value format for With is implementation-specific, even though the interfaces are supposed to abstract away those details. Kinda defeats the purpose of the interfaces, IMO. Maybe the With impls should accept a common format and transform it if necessary to the underlying impl's format?

willfaught: The label names seem to be implementation-dependent as well (please correct me if I'm wrong!). It seems, IIUC, that for Prometheus, you should use underscores as separators, whereas for Statsd you should use periods. I imagine it's possible for another implementation that actually uses With to expect periods too. So for gokit, the label value format for With is implementation-specific, even though the interfaces are supposed to abstract away those details. Kinda defeats the purpose of the interfaces, IMO. Maybe the With impls should accept a common format and transform it if necessary to the underlying impl's format?

Issue #420, opened 2016-12-24
Move to Standard library Context package


GrimmKull:
Context is now a Standard Go library. It is the same code as `golang.org/x/net/context`.
The only thing that is missing is `ctxhttp` used in only for HTTP microservice clients.

I have a suggestion. To add a new branch to the main go-kit repository that uses the standard Context package and the `ctxhttp` implementation.

The code with these changes is available in [GrimmKull fork](https://github.com/GrimmKull/kit/tree/1.7context) in a separate **1.7context** branch.

What do you guys think?

P.S.:

If anyone wants to use this before it gets moved to the main repository you can use vendoring.

Example:

```bash
govendor fetch "github.com/go-kit/kit/endpoint::github.com/GrimmKull/kit/endpoint@1.7context"
```

peterbourgon:
Yep, I was waiting for 1.8 to drop and then make the switch in master. I'll start prepping for that now.

willfaught:
Can't wait!

ziyadparekh:
How do we use go-kit with go version 1.7.4 in the meantime ?

groob:
@ziyadparekh you keep importing `golang.org/x/net/context` in your projects.

ziyadparekh:
The main error it throws is that it cant find the context http package inside the http transport package. And when i try to `go get` the context http package, it throws an error because its not found? Any ideas on what I can do to circumvent these errors?

Apologies for all the questions, but getting the right environment set up is a real pain.

groob:
Can you post the actual error getting instead of describing it? 

The slack #go-kit channel might be a better place to debug this issue.

ziyadparekh:
@groob Sorry about that. Posted the actual error in an edited version of the comment. I'll get myself set up on the Slack channel as well. In case you didnt see it, here is the error:

```
../../go-kit/kit/transport/http/client.go:8:2: cannot find package "golang.org/x/net/context/ctxhttp" in any of:
	/usr/local/go/src/golang.org/x/net/context/ctxhttp (from $GOROOT)
	/Users/zparekh/Go/src/golang.org/x/net/context/ctxhttp (from $GOPATH)
```

willfaught:
`go get -u golang.org/x/net/context/ctxhttp` works for me. It doesn't work for you?

groob:
[...]

PR #421, opened 2016-12-25
Use stdlib context instead of golang.org/x/net/context

peterbourgon:
Addresses #420 (heyoh)

### DIFF
--- a/examples/addsvc/endpoints.go
+++ b/examples/addsvc/endpoints.go
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"time"
 
-	"golang.org/x/net/context"
+	"context"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
--- a/examples/addsvc/endpoints.go
+++ b/examples/addsvc/endpoints.go
@@ -6,10 +6,9 @@ package addsvc
 // formats. It also includes endpoint middlewares.
 
 import (
+	"context"
 	"fmt"
 	"time"
-
-	"context"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: go
 script: go test -race -v ./...
 
 go:
-    - 1.5.4
-    - 1.6.3
-    - 1.7.1
+    - 1.7.4
+    - 1.8beta2
     - tip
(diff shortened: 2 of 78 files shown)


peterbourgon:
Hmm, gRPC still uses x/net/context, which means it can't produce or use endpoints that use plain context. Unless someone has a coercion trick I think we're stuck until gRPC makes the switch.

examples/addsvc/endpoints.go
```diff
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"time"
```
harlow: can remove empty line-break

Ayiga:
Since grpc uses `golang.org/x/net/context.Context`, and it's an interface, the only real issue that we're having is grpc Server method signature don't being directly equivalent, right?

Would it be a decent solution, for the time being, to have a partial mix?

For example, I can get the addsvc working by changing one file in 3 places `transport_grpc.go`:

```go
// ...
import (
	"context"

	gcontext "golang.org/x/net/context"
// ...
)
// ...
func (s *grpcServer) Sum(ctx gcontext.Context, req *pb.SumRequest) (*pb.SumReply, error) {
// ...
}
// ...
func (s *grpcServer) Concat(ctx gcontext.Context, req *pb.ConcatRequest) (*pb.ConcatReply, error) {
// ...
}
// ...
```

Is this a simple enough solution?  I understand the need to try and make this as seamless and easy to use as possible for people.  But the only other possible solution is to create a Middleware for submitting to grpc itself.

```go

func MakeGRPCShimServer(server *grpcServer) pb.AddServer {
    return grpcShimServer {
        grpcServer: server,
   }
}

type grpcShimServer struct {
    grpcServer *grpcServer
}

// A shim Sum method for function matching.
func (s *grpcShimServer) Sum(ctx gcontext.Context, req *pb.SumRequest) (*pb.SumReply, error) {
   return  s.grpcServer.Sum(ctx, req)
}

func (s *grpcShimServer) Concat(ctx gcontext.Context, req *pb.ConcatRequest) (*pb.ConcatReply, error) {
    return s.grpcServer.Concat(ctx, req)
}
```
Then, of course, you'd have to either introduce a new interface or change the return type of the `MakeGRPCServer` function:

```go
// Server API for Add service
type AddServer interface {
	// Sums two integers.
	Sum(context.Context, *SumRequest) (*SumReply, error)
	// Concatenates two strings
	Concat(context.Context, *ConcatRequest) (*ConcatReply, error)
}

// MakeGRPCServer makes a set of endpoints available as a gRPC AddServer.
func MakeGRPCServer(ctx context.Context, endpoints Endpoints, tracer stdopentracing.Tracer, logger log.Logger) AddServer {
// ...
}
```

Issue #415, opened 2016-12-15
add a concurrent example to go-kit/log to avoid confusion in the documentation


hydroflame:

peterbourgon:
I don't understand. Can you clarify?

ChrisHines:
@peterbourgon I got this one. @hydroflame pointed out to me on Slack that when you read the package docs in godoc the last section is about concurrent safety. That is followed immediately by the package level examples, none of which demonstrate using the concurrent safety feature just discussed. The request is to add an example that shows the use of the concurrent safety features. I asked @hydroflame to file an issue to remind me about it.

ChrisHines:
Addressed by #429.

Issue #346, opened 2016-08-22
log: Document error handling philosophy.


ChrisHines:
Handling logging errors has become a FAQ. We should add some documentation to help guide users.

See #164 for previous conversation that should be distilled.

ChrisHines:
Addressed by #429.

PR #429, opened 2017-01-11
Enhance package log docs.

ChrisHines:
Addresses #346 and #415.

### DIFF
--- a/log/example_test.go
+++ b/log/example_test.go
@@ -1,7 +1,9 @@
 package log_test
 
 import (
+	"math/rand"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


log/example_test.go
```diff
@@ -96,7 +98,40 @@ func Example_debugInfo() {
...
+	}
+
+	wg.Add(2)
+
+	go RunTask(Task{ID: 1}, logger)
+	go RunTask(Task{ID: 2}, logger)
+
+	wg.Wait()
+
+	// Unordered output:
```
peterbourgon: Whoa, didn't know that was a thing.

ChrisHines: @peterbourgon It was added in Go 1.7 but I was surprised to discover it missing from the docs at golang.org/pkg/testing. There is an oblique reference in the release notes, but the best place to find it is by running `go help testfunc`.

ChrisHines: @peterbourgon See also: https://github.com/golang/go/issues/10149.

liquidgecka: Glad I could help.. I had basically the exact same issue with a library I wrote a while back so I figured it would be useful.. the deploy became complicated because godoc has to be backwards compatible so that kind of became a crap show for deployments. =(

PR #436, opened 2017-01-12
don't encode json response with http.StatusNoContent

### LINKED ISSUE #435: EncodeJSONResponse does not account for http.StatusNoContent
The [current](https://github.com/go-kit/kit/blob/44df43eae441b8197be13e2004ccdf535e24b277/transport/http/server.go#L149-L154) implementation of `EncodeJSONResponse` always writes an HTTP response body. 
When the status should be `http.StatusNoContent`, there shouldn't be any response body. 

In my own code I've done a check like this:

```
    if e, ok := response.(statusCoder); ok {
        w.WriteHeader(e.status())
        if e.status() == http.StatusNoContent {
            return nil
        }
    }
```

Should the [EncodeJSONResponse](https://github.com/go-kit/kit/blob/44df43eae441b8197be13e2004ccdf535e24b277/transport/http/server.go#L142) have a check for 204 responses? I can send a PR but wanted to make sure it's wanted first.

groob:
204 status cannot contain a message body:
https://tools.ietf.org/html/rfc7231#page-53

Closes #435

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -151,6 +151,9 @@ func EncodeJSONResponse(_ context.Context, w http.ResponseWriter, response inter
 		code = sc.StatusCode()
 	}
 	w.WriteHeader(code)
+	if code == http.StatusNoContent {
+		return nil
+	}
 	return json.NewEncoder(w).Encode(response)
 }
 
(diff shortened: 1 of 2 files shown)


peterbourgon:
🖖

PR #438, opened 2017-01-17
Convert StdlibWriterAdaptor times to RFC3339 with microseconds

### LINKED ISSUE #434: log: StdlibAdaptor timestamp value is weird in logfmt format
It's quoted and doesn't match the RFC3339 format of DefaultTimestamp/UTC:

```
timestamp=2017-01-11T12:58:52-08:00 caller=server.go:16 message=started
timestamp="2017/01/11 12:58:52.426762" caller=http.go:142 message="Serving [::]:8000 with pid 64404"
```

willfaught:
Parses stdlib logger dates and times into RFC3339 format with microsecond precision (since that's the stdlib log precision limit with `log.Lmicroseconds`). Assumes dates and times are in the local time zone unless the UTC option is used (the local time zone is the default because that's how it is for the stdlib logger; adding a UTC option is required because stdlib logger formatting doesn't specify time zones).

If no date is logged, the current date is used (according to the presence or absence of the UTC option). If no time is logged, the current time is used (same deal with the UTC option).

Fixes #434.

### DIFF
--- a/log/stdlib.go
+++ b/log/stdlib.go
@@ -5,6 +5,15 @@ import (
 	"log"
 	"regexp"
 	"strings"
+	"time"
+)
+
+const (
+	formatStdDate      = "2006/01/02"
+	formatStdMicro     = ".999999"
+	formatStdTime      = "15:04:05"
+	formatStdAll       = formatStdDate + " " + formatStdTime + formatStdMicro
+	formatRFC3339Micro = "2006-01-02T15:04:05.999999Z07:00"
 )
 
 // StdlibWriter implements io.Writer by invoking the stdlib log.Print. It's
@@ -29,6 +38,7 @@ type StdlibAdapter struct {
 	timestampKey string
 	fileKey      string
 	messageKey   string
+	utc          bool
 }
 
 // StdlibAdapterOption sets a parameter for the StdlibAdapter.
@@ -49,6 +59,11 @@ func MessageKey(key string) StdlibAdapterOption {
 	return func(a *StdlibAdapter) { a.messageKey = key }
 }
 
+// UTC parses times in the UTC time zone instead of the local time zone.
+func UTC() StdlibAdapterOption {
+	return func(a *StdlibAdapter) { a.utc = true }
+}
+
 // NewStdlibAdapter returns a new StdlibAdapter wrapper around the passed
 // logger. It's designed to be passed to log.SetOutput.
 func NewStdlibAdapter(logger Logger, options ...StdlibAdapterOption) io.Writer {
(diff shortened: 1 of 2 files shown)


willfaught:
@peterbourgon Looks like TestGauge and TestCounter fail intermittently? Is there a way I can rerun the build?

ChrisHines:
@willfaught Thanks for the PR! I restarted the build for you. I don't have time for a code review right now, but I'll try to get to it soon.

peterbourgon:
It's not you, it's Circonus. They write their lib so it's basically impossible to test, and change its behavior so the hacks I use stop working. I'll fix it in master and maybe you can rebase? Sorry for the noise.

willfaught:
Rebased

peterbourgon:
Conflict :( Plus, I'll defer to @ChrisHines for a code review.

willfaught:
Rebased again

willfaught:
Fixed test failures

willfaught:
Ready for review

willfaught:
A note on this line:

```go
timestamp, err = time.ParseInLocation(formatStdAll, d+" "+now.Format(formatStdTime), loc)
```

`formatStdTime` is used instead of `formatStdTime + formatStdMicro` because the microsecond resolution would make it untestable, whereas the second resolution is long enough for the tests to reliably pass. The tests were already relying on this trick.

ChrisHines:
As I started reviewing this PR I noticed some things that looked like mistakes in the tests, but then I realized that they are intentional and that this PR changes the behavior of `StdlibAdapter` more dramatically than I expected.

I was expecting a change that would add a feature to `StdlibAdapter` allowing it to optionally reformat time-stamps to RFC3339 from the information available in the output of stdlib.log. Parsing the intent from the code I gather that this PR forces all time-stamps emitted by `StdlibAdapter` into RFC3339 format. In addition it attempts to augment the received information with any missing components. For example it will add the date if only the time is received, or vice versa.

I have two primary concerns with this PR as it stands. First—and most critically—the attempt to augment the received time-stamp with additional information suffers from a fatal flaw. It is using data from two distinct calls to `time.Now`, one made by stdlib.log and another made by `StdlibAdapter.Write`. The case of adding the missing date information has the potential to cause the biggest error. Suppose the two calls to `time.Now` return the times:

- 2017-01-17T23:59:59.999999
- 2017-01-18T00:00:00.000010

When stdlib.log has flags set to `log.Ltime` it will emit only the time portion of the first time-stamp. This PR's code will then combine that with the date portion of the second time-stamp and output a time-stamp nearly 24 hours in the future.

- 2017-01-18T23:59:59.999999

My second concern is simply that this PR changes the behavior of existing programs and removes some choices previously available.

I think the solution presented as a "workaround" in #434 is preferable to this implementation because it produces all of the time-stamp data from a single call to `time.Now` and does not change existing programs or remove existing options.

Can we find a less ambitious approach?

willfaught:
Thanks for the thoughtful review, Chris. You make excellent points, and I
agree with all of them.

What do you think of this approach:

- Option for time parse format
- Option for time parse location
- Transform the parsed stdlib time value with the options in Write
- Default to time.RFC3339Nano and time.UTC

willfaught:
Better:

- Option to transform the stdlib date/time from a format/location to another format/location
- If the option is given, do it in Write

willfaught:
Another thought: Why not remove the stdlib time parsing altogether and let the user use that workaround if they need a timestamp? This way they're not stuck having the stdlib timestamp passed on to the kit logger if they don't want/need it.

ChrisHines:
>Another thought: Why not remove the stdlib time parsing altogether and let the user use that workaround if they need a timestamp? This way they're not stuck having the stdlib timestamp passed on to the kit logger if they don't want/need it.

Configuring the stdlib logger to leave out the time-stamp has the same effect. Maybe we just need to add a suggestion along those lines to the docs for StdlibAdapter.

willfaught:
What if you don't have control over the flags?

ChrisHines:
>What if you don't have control over the flags?

That is certainly possible. I speculate, however, that if you cannot set the flags you also cannot redirect the output to a StdlibAdapter. It seems fair to me to assume that either you have complete control over the stdlib.Logger used, or you have no control, and I think that the latter case is out of scope for StdlibAdapter. Would you agree?

willfaught:
I agree. That makes sense.

I guess the only remaining issue I have is that parsing the stdlib timestamp seems pointless unless you have a way to tell Adaptor how to interpret it, since there's no time zone info, and, to a lesser extent, the format isn't a common one. If you want it to be used like other timestamps in your architecture, so that, for example, you could search for events by timestamp in Kibana, it won't conform and thus be useful.

Would you ever use the parsed time as-is without ever transforming it in a production environment?

ChrisHines:
This package doesn't have a strong opinion about time-stamps. It doesn't even require them at all, and if you don't use time-stamps it won't make any calls to `time.Now`, which is more (or perhaps less) than most logging packages can say. This package provides the building blocks to tailor a broad range of logging solutions with the primary opinion being that structured logging is better than `Printf` based log messages.

This package provides some helpers for including good time-stamps in log events because it is obviously a **_very_** common requirement. 

We provide StdlibAdapter for two primary reasons.
* To integrate packages (like [net/http](https://golang.org/src/net/http/server.go#L2125) that only work with stdlib.Logger. In this case I think it is best to disable time-stamps from that logger, as we've discussed, to keep all time-stamps in a consistent format. Another choice (albeit probably inferior) is to configure the Go-kit Logger to emit time-stamps in the same format as stdlib.Logger.
* To transition an existing project from non-structured logging to structured logging. I can imagine some people in this situation *may* want to keep the same sort of time-stamps they have had in the past. I think this is the audience that would most likely appreciate that StdlibAdapter extracts the time-stamps for them.

willfaught:
Fair enough. Thanks for putting up with me! ;)

PR #446, opened 2017-01-24
Add TTL for Consul SD

antoniomo:
Referencing issue #343 but for Consul instead of etcd.

All tests pass including integration at least on local Consul, but not sure if the proposed API is good (it's just what I'm using on some services here). Comments and suggestions of all kinds are welcome!


antoniomo:
After using this interface for a bit in our projects, I now think that the special treatment for TTL feels clunky, specially considering it's not the go-to health check in Consul. But I still appreciate the register/deregister check methods which I'm using in all our services. Any ideas on how to change the interface to make it more useful?

antoniomo:
Removing this PR as I'm not maintaining it anymore. Let me know if we want something similar and I can revisit it.

PR #450, opened 2017-02-06
Fix typo in graphite package

jkakar:
Rename `Statsd` to `Graphite` in the docstring for the `graphite.New` constructor.


peterbourgon:
Yep, thanks!

PR #451, opened 2017-02-07
JSON RPC over HTTP Transport

blainsmith:
This is just a first pass at adding a `jsonrpc` package under `http` as a transport. I haven't written tests yet, since I wanted to gather feedback as to whether this was in the spirit of gokit before continuing along. Feel free to rip it apart and let me know how it looks.

This was a awesome learning experience too!

### DIFF
--- a/transport/http/jsonrpc/request_response_types.go
+++ b/transport/http/jsonrpc/request_response_types.go
@@ -1,0 +1,28 @@
+package jsonrpc
+
+import "encoding/json"
+
+// Request defines a JSON RPC request from the spec
+// http://www.jsonrpc.org/specification#request_object
+type Request struct {
+	JSONRPC string          `json:"jsonrpc"`
+	Method  string          `json:"method"`
+	Params  json.RawMessage `json:"params"`
+	ID      interface{}     `json:"id"`
+}
+
+// Response defines a JSON RPC response from the spec
+// http://www.jsonrpc.org/specification#response_object
+type Response struct {
+	JSONRPC string      `json:"jsonrpc"`
+	Result  interface{} `json:"result,omitempty"`
+	Error   Error       `json:"error,omitemty"`
+}
+
+const (
+	// Version defines the version of the JSON RPC implementation
+	Version string = "2.0"
+
+	// ContentType defines the content type to be served.
+	ContentType string = "application/json; charset=utf-8"
+)
--- a/transport/http/jsonrpc/server.go
+++ b/transport/http/jsonrpc/server.go
@@ -1,0 +1,202 @@
+package jsonrpc
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/log"
+	httptransport "github.com/go-kit/kit/transport/http"
+)
+
+// Server wraps an endpoint and implements http.Handler.
+type Server struct {
+	ctx       context.Context
+	ecm       EndpointCodecMap
+	before    []httptransport.RequestFunc
+	after     []httptransport.ServerResponseFunc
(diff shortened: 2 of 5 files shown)


willfaught:
In general, is jsonrpc tied to HTTP or can it be done over raw TCP too? If so, it'd be nice to have that option.

blainsmith:
In general, no it isn't. Technically HTTP is the transport and JSON RPC is the codec. To abstract this even more RPC is is the codec and JSON is the data format, you could have XML :trollface: or anything else if there was an interface defined for it.

I started down that route, but pulled back since I didn't want to get too deep into the rabbit hole before involving maintainers.

willfaught:
Doesn't the std lib RPC library have a jsonrpc codec? If I understand
correctly, that doesn't use HTTP. Is that an outlier then?

peterbourgon:
I've never heard of anyone doing JSON/RPC over anything other than HTTP, I wouldn't put that up as a blocker.

blainsmith:
Yes it does, however it does not provide the convenience structs for doing it over HTTP `POST` methods. It serves over HTTP, but requires the `CONNECT` method to process the request.

The stdlib does't expose the proper request and response object either, especially for decoding/encoding to JSON with struct tags.

transport/http/jsonrpc/request_response_types.go
```diff
@@ -0,0 +1,28 @@
...
+
+import "encoding/json"
+
+// Request defines a JSON RPC request from the spec
+// http://www.jsonrpc.org/specification#request_object
+type Request struct {
+	JSONRPC string          `json:"jsonrpc"`
+	Method  string          `json:"method"`
+	Params  json.RawMessage `json:"params"`
+	ID      interface{}     `json:"id"`
```
groob: Instead of `interface{}`, the ID could be a custom type that implements MarshalJSON. Thoughts?

transport/http/jsonrpc/server.go
```diff
@@ -0,0 +1,202 @@
...
+		w = iw
+	}
+
+	for _, f := range s.before {
+		ctx = f(ctx, r)
+	}
+
+	// Decode the body into an  object
+	var req Request
+	err := json.NewDecoder(r.Body).Decode(&req)
```
groob: It's a bit troubling that there is no way to wrap the r.Body by another io.Reader here. For example someone might want to wrap it with a `io.LimitReader` to ensure that the server is not flooded with a stream of neverending bytes. 

Perhaps a ServerOption can be added here.

transport/http/jsonrpc/server.go
```diff
@@ -0,0 +1,202 @@
...
+	resParams, err := ecm.Encode(ctx, response)
+	if err != nil {
+		s.logger.Log("err", err)
+		rpcErrorEncoder(ctx, err, w)
+		return
+	}
+
+	res.Result = resParams
+
+	json.NewEncoder(w).Encode(res)
```
groob: Similar to the Decode, but here my request would be to be able to set options on the encoder: Example to pretty print the response, I'd want to `SetIndent`

blainsmith:
@groob Thanks for the feedback here. I appreciate it. I will spend more time cleaning up these parts and update the PR.

peterbourgon:
Just wanted to make it clear I'm +1 on this package existing, and the general direction looks good to me. I'll withhold further review until you get it into a more final state. JSON/RPC is new to me, I'd love to see a comprehensive example as part of the PR.

basvanbeek:
I would like to see if we can extend the transport's logic to include RPC batch support which is part of the JSON-RPC 2.0 spec.

The spec is very friendly in that it is explicit about not enforcing any guarantees in parallelism or response order sequence so having this feature in should not be terribly difficult.

peterbourgon:
Closing in favor of #576

PR #453, opened 2017-02-08
log/alt_experimental_level: Introduce general projection and filtering

seh:
Following up from #444, demonstrate an alternate approach to the "level" package offered via the "github.com/go-kit/kit/log/experimental_package" import path (pending promotion via #449), allowing log event level filtering via functions that specify the filtering threshold.

This capability is implemented in terms of a new concept for `log.Context`: _projection functions_. There are two new exported methods on `log.Context` that mention a new exported type, though defining that type is a convenience and isn't strictly necessary:
- **`log.Projection`**  
  A convenience declaration of a _projection function_ that accepts a proposed set of key/value pairs and either drops the record (and any later records that would include those key/value pairs) or proposes a revised set of key/value pairs.
- **`log.Context.WithPreProjection`**  
  Composes a _projection function_ onto a `Context`, applying it early as the "innermost" such function.
- **`log.Context.WithPostProjection`**  
  Composes a _projection function_ onto a `Context`, applying it late as the "outermost" such function.

Including this function in `log.Context` increases the struct's size by four bytes, which accounts for the 16.67% increase in allocations per operation seen in the benchmarks. That is, each time we allocate a new `Context` in `log.NewContext`, we allocate 28 instead of 24 bytes.

There are a few operations that are slower than the alternate operation, such as the JSON and _logfmt_ conversion. I _think_ this is due to my level values being implementations of `fmt.Stringer` as opposed to `string` values; both those formatters visit such values second in line, and call on them with a deferred function to trap panics. As my interest here is more in cutting down on CPU and allocation cost in cases where most log messages are filtered out, I stuck with my `level.levelValue` type here, as it yielded faster benchmarks for discerning whether the value should be dropped by the filter.

With these _projection functions_ in place, it's possible to implement [log15's `ext.EscalateErrHandler`](https://godoc.org/github.com/inconshreveable/log15/ext#EscalateErrHandler), though only within the `level` package, because the `level.levelValue` type is not exported. We would have to make a copy of the supplied key/value slice and replace the level value in it if we discover a non-nil `error` elsewhere in the sequence, so it's not as efficient as `ext.EscalateErrHandler`, but that library's records include the level directly.

Invoking these projection functions involves trading trust for efficiency. I considered making a defensive copy of the key/value sequence supplied to each invocation of the function, which allows the function the opportunity to mutate the sequence in place. However, that induces allocation and copying even for cases where the projection function is just a pure filter; even sequences it would vote to drop would have been copied first. Instead, I decided to document that a well-behaved `log.Projection` does not mutate the supplied slice, expecting that the most common cases would involve only inspecting the proposed key/value pairs and either dropping them or preserving them as they are. I expect that this choice will be controversial.

Here are the benchmark comparisons:

| name  | old time/op | new time/op  | delta |
| --- | --- | --- | --- |
| NopBaseline-8           |  470ns ± 0% |  443ns ± 0% |  -5.74% |
| NopDisallowedLevel-8    |  513ns ± 0% |  187ns ± 0% | -63.55% |
| NopAllowedLevel-8       |  516ns ± 0% |  451ns ± 0% | -12.60% |
| JSONBaseline-8          | 2.44µs ± 0% | 2.40µs ± 0% |  -1.88% |
| JSONDisallowedLevel-8   |  515ns ± 0% |  184ns ± 0% | -64.27% |
| JSONAllowedLevel-8      | 2.48µs ± 0% | 2.34µs ± 0% |  -5.77% |
| LogfmtBaseline-8        | 1.05µs ± 0% | 1.13µs ± 0% |  +7.71% |
| LogfmtDisallowedLevel-8 |  514ns ± 0% |  181ns ± 0% | -64.79% |
| LogfmtAllowedLevel-8    | 1.13µs ± 0% | 1.13µs ± 0% |  -0.18% |

| name | old alloc/op | new alloc/op | delta |
| --- | --- | --- | --- |
| NopBaseline-8           | 288B ± 0% |  336B ± 0% | +16.67% |
| NopDisallowedLevel-8    | 288B ± 0% |  112B ± 0% | -61.11% |
| NopAllowedLevel-8       | 288B ± 0% |  336B ± 0% | +16.67% |
| JSONBaseline-8          | 968B ± 0% | 1032B ± 0% |  +6.61% |
| JSONDisallowedLevel-8   | 288B ± 0% |  112B ± 0% | -61.11% |
| JSONAllowedLevel-8      | 968B ± 0% | 1032B ± 0% |  +6.61% |
| LogfmtBaseline-8        | 288B ± 0% |  336B ± 0% | +16.67% |
| LogfmtDisallowedLevel-8 | 288B ± 0% |  112B ± 0% | -61.11% |
| LogfmtAllowedLevel-8    | 288B ± 0% |  336B ± 0% | +16.67% |

| name | old allocs/op | new allocs/op | delta |
| --- | --- | --- | --- |
| NopBaseline-8           | 9.00 ± 0% | 9.00 ± 0% |  +0.00% |
| NopDisallowedLevel-8    | 9.00 ± 0% | 5.00 ± 0% | -44.44% |
| NopAllowedLevel-8       | 9.00 ± 0% | 9.00 ± 0% |  +0.00% |
| JSONBaseline-8          | 22.0 ± 0% | 23.0 ± 0% |  +4.55% |
| JSONDisallowedLevel-8   | 9.00 ± 0% | 5.00 ± 0% | -44.44% |
| JSONAllowedLevel-8      | 22.0 ± 0% | 23.0 ± 0% |  +4.55% |
| LogfmtBaseline-8        | 9.00 ± 0% | 9.00 ± 0% |  +0.00% |
| LogfmtDisallowedLevel-8 | 9.00 ± 0% | 5.00 ± 0% | -44.44% |
| LogfmtAllowedLevel-8    | 9.00 ± 0% | 9.00 ± 0% |  +0.00% |

They are not as obviously rosy. I think that this approach's benefits would become more clear when we see filtering applied to a `log.Context` with lots of log messages dropped, such as the following:
```go
func BenchmarkManyDroppedRecords(b *testing.B) {
	logger := level.AllowingInfoAndAbove(log.NewJSONLogger(ioutil.Discard))
	b.ResetTimer()
	b.ReportAllocs()
	debug := level.Debug(logger)
	info := level.Info(logger)
	for i := 0; i < b.N; i++ {
		debug.Log("foo", "1")
		// Only this one will be retained.
		info.Log("baz", "quux")
		debug.Log("foo", "2")
		debug.Log("foo", "3")
		debug.Log("foo", "4")
	}
}
```
Run against the equivalent version from the _experimental_level_ directory:

| name  | old time/op | new time/op  | delta |
| --- | --- | --- | --- |
| ManyDroppedRecords-8 | 3.36µs ± 0% | 2.79µs ± 0% | -16.80% |

| name | old alloc/op | new alloc/op | delta |
| --- | --- | --- | --- |
| ManyDroppedRecords-8 | 1.32kB ± 0% | 1.08kB ± 0% | -18.18% |

| name | old allocs/op | new allocs/op | delta |
| --- | --- | --- | --- |
| ManyDroppedRecords-8 | 33.0 ± 0% | 30.0 ± 0% | -9.09% |

Lacking still are unit tests to cover the new `Context.WithPreProjection` and `Context.WithPostProjection` functions. As each has two parameters, the first of which can be nil, there are at least four interesting cases for each one, as well as exercising proper composition with an existing projection function in the `Context`. I didn't write those yet, as I am uncertain whether the maintainers are amenable to the ideas introduced here.

I welcome your thoughts, but if you're tired of considering these alternative proposals, I understand.

### DIFF
--- a/log/alt_experimental_level/benchmark_test.go
+++ b/log/alt_experimental_level/benchmark_test.go
@@ -1,0 +1,59 @@
+package level_test
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/alt_experimental_level"
+)
+
+func BenchmarkNopBaseline(b *testing.B) {
+	benchmarkRunner(b, log.NewNopLogger())
+}
+
+func BenchmarkNopDisallowedLevel(b *testing.B) {
+	benchmarkRunner(b,
+		level.AllowingInfoAndAbove(log.NewNopLogger()))
+}
+
+func BenchmarkNopAllowedLevel(b *testing.B) {
+	benchmarkRunner(b,
+		level.AllowingAll(log.NewNopLogger()))
+}
+
+func BenchmarkJSONBaseline(b *testing.B) {
+	benchmarkRunner(b, log.NewJSONLogger(ioutil.Discard))
+}
+
+func BenchmarkJSONDisallowedLevel(b *testing.B) {
+	benchmarkRunner(b,
+		level.AllowingInfoAndAbove(log.NewJSONLogger(ioutil.Discard)))
+}
+
+func BenchmarkJSONAllowedLevel(b *testing.B) {
+	benchmarkRunner(b,
+		level.AllowingAll(log.NewJSONLogger(ioutil.Discard)))
+}
+
+func BenchmarkLogfmtBaseline(b *testing.B) {
+	benchmarkRunner(b, log.NewLogfmtLogger(ioutil.Discard))
+}
+
+func BenchmarkLogfmtDisallowedLevel(b *testing.B) {
+	benchmarkRunner(b,
+		level.AllowingInfoAndAbove(log.NewLogfmtLogger(ioutil.Discard)))
+}
+
+func BenchmarkLogfmtAllowedLevel(b *testing.B) {
+	benchmarkRunner(b,
+		level.AllowingAll(log.NewLogfmtLogger(ioutil.Discard)))
+}
+
+func benchmarkRunner(b *testing.B, logger log.Logger) {
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		level.Debug(logger).Log("foo", "bar")
+	}
+}
--- a/log/alt_experimental_level/benchmark_test.go
+++ b/log/alt_experimental_level/benchmark_test.go
@@ -1,6 +1,7 @@
 package level_test
 
 import (
+	"fmt"
 	"io/ioutil"
 	"testing"
 
@@ -9,48 +10,48 @@ import (
 )
 
 func BenchmarkNopBaseline(b *testing.B) {
-	benchmarkRunner(b, log.NewNopLogger())
+	singleRecordBenchmarkRunner(b, log.NewNopLogger())
 }
 
 func BenchmarkNopDisallowedLevel(b *testing.B) {
-	benchmarkRunner(b,
+	singleRecordBenchmarkRunner(b,
 		level.AllowingInfoAndAbove(log.NewNopLogger()))
 }
 
 func BenchmarkNopAllowedLevel(b *testing.B) {
-	benchmarkRunner(b,
+	singleRecordBenchmarkRunner(b,
 		level.AllowingAll(log.NewNopLogger()))
 }
 
 func BenchmarkJSONBaseline(b *testing.B) {
-	benchmarkRunner(b, log.NewJSONLogger(ioutil.Discard))
+	singleRecordBenchmarkRunner(b, log.NewJSONLogger(ioutil.Discard))
 }
 
 func BenchmarkJSONDisallowedLevel(b *testing.B) {
-	benchmarkRunner(b,
+	singleRecordBenchmarkRunner(b,
 		level.AllowingInfoAndAbove(log.NewJSONLogger(ioutil.Discard)))
 }
 
 func BenchmarkJSONAllowedLevel(b *testing.B) {
-	benchmarkRunner(b,
+	singleRecordBenchmarkRunner(b,
 		level.AllowingAll(log.NewJSONLogger(ioutil.Discard)))
 }
 
 func BenchmarkLogfmtBaseline(b *testing.B) {
-	benchmarkRunner(b, log.NewLogfmtLogger(ioutil.Discard))
+	singleRecordBenchmarkRunner(b, log.NewLogfmtLogger(ioutil.Discard))
 }
 
 func BenchmarkLogfmtDisallowedLevel(b *testing.B) {
-	benchmarkRunner(b,
+	singleRecordBenchmarkRunner(b,
 		level.AllowingInfoAndAbove(log.NewLogfmtLogger(ioutil.Discard)))
 }
 
 func BenchmarkLogfmtAllowedLevel(b *testing.B) {
-	benchmarkRunner(b,
+	singleRecordBenchmarkRunner(b,
 		level.AllowingAll(log.NewLogfmtLogger(ioutil.Discard)))
 }
 
-func benchmarkRunner(b *testing.B, logger log.Logger) {
+func singleRecordBenchmarkRunner(b *testing.B, logger log.Logger) {
 	b.ResetTimer()
 	b.ReportAllocs()
 	for i := 0; i < b.N; i++ {
@@ -58,18 +59,27 @@ func benchmarkRunner(b *testing.B, logger log.Logger) {
 	}
 }
 
-func BenchmarkManyDroppedRecords(b *testing.B) {
-	logger := level.AllowingInfoAndAbove(log.NewJSONLogger(ioutil.Discard))
-	b.ResetTimer()
-	b.ReportAllocs()
+func BenchmarkDroppedRecords(b *testing.B) {
+	logger := log.NewNopLogger()
+	logger = log.NewContext(logger).With("ts", log.DefaultTimestamp, "caller", log.DefaultCaller)
+	for _, dropped := range []uint{1, 3, 9, 99, 999} {
+		b.Run(fmt.Sprintf("%d-of-%d", dropped, dropped+1), func(b *testing.B) {
+			manyRecordBenchmarkRunner(b, logger, dropped)
+		})
+	}
+}
+
+func manyRecordBenchmarkRunner(b *testing.B, logger log.Logger, droppedRecords uint) {
+	logger = level.AllowingInfoAndAbove(logger)
 	debug := level.Debug(logger)
 	info := level.Info(logger)
+	b.ResetTimer()
+	b.ReportAllocs()
 	for i := 0; i < b.N; i++ {
-		debug.Log("foo", "1")
 		// Only this one will be retained.
-		info.Log("baz", "quux")
-		debug.Log("foo", "2")
-		debug.Log("foo", "3")
-		debug.Log("foo", "4")
+		info.Log("foo", "bar")
+		for ; droppedRecords != 0; droppedRecords-- {
+			debug.Log("baz", "quux")
+		}
 	}
 }
--- a/log/alt_experimental_level/benchmark_test.go
+++ b/log/alt_experimental_level/benchmark_test.go
@@ -57,3 +57,19 @@ func benchmarkRunner(b *testing.B, logger log.Logger) {
 		level.Debug(logger).Log("foo", "bar")
 	}
 }
+
+func BenchmarkManyDroppedRecords(b *testing.B) {
+	logger := level.AllowingInfoAndAbove(log.NewJSONLogger(ioutil.Discard))
+	b.ResetTimer()
+	b.ReportAllocs()
+	debug := level.Debug(logger)
+	info := level.Info(logger)
+	for i := 0; i < b.N; i++ {
+		debug.Log("foo", "1")
+		// Only this one will be retained.
+		info.Log("baz", "quux")
+		debug.Log("foo", "2")
+		debug.Log("foo", "3")
+		debug.Log("foo", "4")
+	}
+}
--- a/log/alt_experimental_level/benchmark_test.go
+++ b/log/alt_experimental_level/benchmark_test.go
@@ -78,7 +78,7 @@ func manyRecordBenchmarkRunner(b *testing.B, logger log.Logger, droppedRecords u
 	for i := 0; i < b.N; i++ {
 		// Only this one will be retained.
 		info.Log("foo", "bar")
-		for ; droppedRecords != 0; droppedRecords-- {
+		for dropped := droppedRecords; dropped != 0; dropped-- {
 			debug.Log("baz", "quux")
 		}
 	}
--- a/log/alt_experimental_level/level_test.go
+++ b/log/alt_experimental_level/level_test.go
@@ -112,6 +112,20 @@ func TestContextLevel(t *testing.T) {
 	}
 }
 
+func TestLevelRestrictionOverLeveledContext(t *testing.T) {
+	var buf bytes.Buffer
+
+	var logger log.Logger
+	logger = log.NewLogfmtLogger(&buf)
+	logger = level.Debug(logger)
+	logger = level.AllowingInfoAndAbove(logger)
+	logger.Log("this is", "debug log")
+
+	if buf.Len() > 0 {
+		t.Errorf("want no output, have %q", buf.String())
+	}
+}
+
 func TestLevelLayerRestrictions(t *testing.T) {
 	factories := []struct {
 		name string
--- a/log/alt_experimental_level/level_test.go
+++ b/log/alt_experimental_level/level_test.go
@@ -1,0 +1,231 @@
+package level_test
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/alt_experimental_level"
+)
+
+func TestInstanceLevels(t *testing.T) {
+	for _, testcase := range []struct {
+		allowed string
+		allow   func(log.Logger) log.Logger
+		want    string
+	}{
+		{
+			"all",
+			level.AllowingAll,
+			strings.Join([]string{
+				`{"level":"debug","this is":"debug log"}`,
+				`{"level":"info","this is":"info log"}`,
+				`{"level":"warn","this is":"warn log"}`,
+				`{"level":"error","this is":"error log"}`,
+			}, "\n"),
+		},
+		{
+			"debug+",
+			level.AllowingDebugAndAbove,
+			strings.Join([]string{
+				`{"level":"debug","this is":"debug log"}`,
+				`{"level":"info","this is":"info log"}`,
+				`{"level":"warn","this is":"warn log"}`,
+				`{"level":"error","this is":"error log"}`,
+			}, "\n"),
+		},
+		{
+			"info+",
+			level.AllowingInfoAndAbove,
+			strings.Join([]string{
+				`{"level":"info","this is":"info log"}`,
+				`{"level":"warn","this is":"warn log"}`,
+				`{"level":"error","this is":"error log"}`,
+			}, "\n"),
+		},
+		{
+			"warn+",
+			level.AllowingWarnAndAbove,
+			strings.Join([]string{
+				`{"level":"warn","this is":"warn log"}`,
+				`{"level":"error","this is":"error log"}`,
+			}, "\n"),
+		},
+		{
+			"error",
+			level.AllowingErrorOnly,
+			strings.Join([]string{
+				`{"level":"error","this is":"error log"}`,
+			}, "\n"),
+		},
+		{
+			"none",
+			level.AllowingNone,
+			``,
+		},
+	} {
+		var buf bytes.Buffer
+		logger := testcase.allow(log.NewJSONLogger(&buf))
+
+		level.Debug(logger).Log("this is", "debug log")
+		level.Info(logger).Log("this is", "info log")
+		level.Warn(logger).Log("this is", "warn log")
+		level.Error(logger).Log("this is", "error log")
+
+		if want, have := testcase.want, strings.TrimSpace(buf.String()); want != have {
+			t.Errorf("given Allowed=%s: want\n%s\nhave\n%s", testcase.allowed, want, have)
+		}
+	}
+}
+
+func TestLevelContext(t *testing.T) {
+	var buf bytes.Buffer
+
+	// Wrapping the level logger with a context allows users to use
+	// log.DefaultCaller as per normal.
+	var logger log.Logger
+	logger = log.NewLogfmtLogger(&buf)
+	logger = level.AllowingAll(logger)
+	logger = level.Info(logger)
+	logger = log.NewContext(logger).With("caller", log.DefaultCaller)
+
+	logger.Log("foo", "bar")
+	if want, have := `level=info caller=level_test.go:93 foo=bar`, strings.TrimSpace(buf.String()); want != have {
+		t.Errorf("want %q, have %q", want, have)
+	}
+}
+
+func TestContextLevel(t *testing.T) {
+	var buf bytes.Buffer
+
+	// Wrapping the level logger with a context allows users to use
+	// log.DefaultCaller as per normal.
+	var logger log.Logger
+	logger = log.NewLogfmtLogger(&buf)
+	logger = log.NewContext(logger).With("caller", log.DefaultCaller)
+
+	logger = level.AllowingAll(logger)
+	level.Info(logger).Log("foo", "bar")
+	if want, have := `level=info caller=level_test.go:109 foo=bar`, strings.TrimSpace(buf.String()); want != have {
+		t.Errorf("want %q, have %q", want, have)
+	}
+}
+
+func TestLevelLayerRestrictions(t *testing.T) {
+	factories := []struct {
+		name string
+		f    func(log.Logger) log.Logger
+	}{
+		{"all", level.AllowingAll},
+		{"debug+", level.AllowingDebugAndAbove},
+		{"info+", level.AllowingInfoAndAbove},
+		{"warn+", level.AllowingWarnAndAbove},
+		{"error", level.AllowingErrorOnly},
+		{"none", level.AllowingNone},
+	}
+	emitters := []struct {
+		name string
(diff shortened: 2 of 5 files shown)


log/log.go
```diff
@@ -133,6 +184,90 @@ func (l *Context) WithPrefix(keyvals ...interface{}) *Context {
 	}
```
seh: - [ ] Add the following:
```go
projection: l.projection,
```

seh:
Here's a hastily written implementation of error escalation, situated within the "level" package:
```go
func EscalatingErrors(logger log.Logger) log.Logger {
	return log.NewContext(logger).WithPreProjection(func(keyvals []interface{}) ([]interface{}, bool) {
		resolve := func(v *levelValue, index int) ([]interface{}, bool) {
			if v != valError {
				kvs := make([]interface{}, len(keyvals))
				base := copy(kvs, keyvals[:index])
				kvs[index] = valError
				copy(kvs[base+1:], keyvals[index+1:])
				return kvs, true
			}
			return keyvals, true
		}
		for i, end := 1, len(keyvals); i < end; i += 2 {
			switch v := keyvals[i].(type) {
			case *levelValue:
				index := i
				i += 2
				for ; i < end; i += 2 {
					if err, ok := keyvals[i].(error); ok && err != nil {
						return resolve(v, index)
					}
				}
				break
			case error:
				if v == nil {
					continue
				}
				i += 2
				for ; i < end; i += 2 {
					if l, ok := keyvals[i].(*levelValue); ok {
						return resolve(l, i)
					}
				}
				break
			}
		}
		return keyvals, true
	}, true)
}
```
Note that it's a "pre" (early) projection, to intervene before the "info"-and-above filter, and it's deferred, to preclude the "info"-and-above filter from rendering the `Context` inert as soon as it sees the "debug"-level value captured early on, before we've seen whether there's a non-nil `error` value in the eventual log record.

A corresponding test case:
```go
func TestEscalatingErrors(t *testing.T) {
	var buf bytes.Buffer
	var logger log.Logger
	logger = level.EscalatingErrors(level.AllowingInfoAndAbove(log.NewLogfmtLogger(&buf)))
	logger = level.Debug(logger)
	logger.Log("msg", "success", "err", nil)
	logger.Log("msg", "failure", "err", errors.New("failed"))
	if want, have := `level=error msg=failure err=failed`, strings.TrimSpace(buf.String()); want != have {
	}
}
```

log/alt_experimental_level/level_test.go
```diff
@@ -0,0 +1,231 @@
...
+	logger.Log("foo", "bar")
+	if want, have := `level=info caller=level_test.go:93 foo=bar`, strings.TrimSpace(buf.String()); want != have {
+		t.Errorf("want %q, have %q", want, have)
+	}
+}
+
+func TestContextLevel(t *testing.T) {
+	var buf bytes.Buffer
+
+	// Wrapping a context with the level logger still works, but requires users
```
seh: - [ ] Fix this comment.  
  It should instead match the one at line 85 above.

ChrisHines:
I found some time to read through this idea today. I haven't completely wrapped my head around the new concepts. I get the idea of a projection, but the implementation and its ramifications are not obvious to me after 1.5 passes. That is not encouraging.

From a user's perspective I am concerned that the interaction between pre/post projections are not easy grasp.

From an API design perspective Projections and Loggers are not orthogonal. I think deferred projections are redundant because the same choices could be made in the Logger called at the end of `Context.Log` without any extra overhead. That creates two ways to do common tasks with this package. That is undesirable.

seh:
Thank you for considering the proposal. I value your feedback.

I started out with just the "pre" projection, and no deferral. I only added the deferral option after you mentioned `ext.EscalateErrorHandler`, and it seemed like it was a variation on what I was in the middle of implementing. It wasn't until early yesterday morning (when all things from the day prior become more clear) that I thought, "You know, that deferred projection is really the same as a `log.Logger`." I didn't go back and remove the deferral option, though, not yet being sure whether it was a worthwhile idea.

As for the pre- and post-composition, I thought of this analogously to languages like Common Lisp that have the notion of "before methods," "after methods," and "around methods." I didn't want repeated calls to what was originally `Context.WithProjection` to snuff out preceding projections already bound to the `Context`, but I wasn't sure that it's always correct to treat the new projection function as either the "outer" one or the "inner" one, if we are going to compose a new function with it. For leveled record filtering, calling the filtering function earliest seems correct.

Do you think it's worth pulling back from the deferral and composition options, and just exposing a single `Context.WithProjection` method? I expect that would take me a few hours to whittle down.

ChrisHines:
Yes, I would be interested to see what the simpler version looks like.

seh:
Good. I'm already most of the way through it.
I will probably have to wait until Monday to post an update here. I'll provide fresh benchmark results too.

seh:
Per your suggestion, I retained only `Context.WithProjection`, which is composed before any previous projection functions.

Here are the latest benchmarks:
```
name                          old time/op    new time/op    delta
NopBaseline-8                    518ns ± 0%     485ns ± 0%   -6.37%
NopDisallowedLevel-8             568ns ± 0%     206ns ± 0%  -63.73%
NopAllowedLevel-8                568ns ± 0%     490ns ± 0%  -13.73%
JSONBaseline-8                  2.52µs ± 0%    2.64µs ± 0%   +4.96%
JSONDisallowedLevel-8            563ns ± 0%     202ns ± 0%  -64.12%
JSONAllowedLevel-8              2.57µs ± 0%    2.72µs ± 0%   +5.88%
LogfmtBaseline-8                1.14µs ± 0%    1.28µs ± 0%  +12.00%
LogfmtDisallowedLevel-8          562ns ± 0%     202ns ± 0%  -64.06%
LogfmtAllowedLevel-8            1.20µs ± 0%    1.26µs ± 0%   +5.51%
DroppedRecords/1-of-2-8         1.79µs ± 0%    1.64µs ± 0%   -8.11%
DroppedRecords/3-of-4-8         1.77µs ± 0%    1.65µs ± 0%   -7.21%
DroppedRecords/9-of-10-8        1.77µs ± 0%    1.65µs ± 0%   -6.93%
DroppedRecords/99-of-100-8      1.78µs ± 0%    1.62µs ± 0%   -9.31%
DroppedRecords/999-of-1000-8    1.77µs ± 0%    1.64µs ± 0%   -7.62%

name                          old alloc/op   new alloc/op   delta
NopBaseline-8                     352B ± 0%      336B ± 0%   -4.55%
NopDisallowedLevel-8              352B ± 0%      112B ± 0%  -68.18%
NopAllowedLevel-8                 352B ± 0%      336B ± 0%   -4.55%
JSONBaseline-8                  1.03kB ± 0%    1.03kB ± 0%   +0.00%
JSONDisallowedLevel-8             352B ± 0%      112B ± 0%  -68.18%
JSONAllowedLevel-8              1.03kB ± 0%    1.03kB ± 0%   +0.00%
LogfmtBaseline-8                  352B ± 0%      336B ± 0%   -4.55%
LogfmtDisallowedLevel-8           352B ± 0%      112B ± 0%  -68.18%
LogfmtAllowedLevel-8              352B ± 0%      336B ± 0%   -4.55%
DroppedRecords/1-of-2-8           335B ± 0%      335B ± 0%   +0.00%
DroppedRecords/3-of-4-8           335B ± 0%      335B ± 0%   +0.00%
DroppedRecords/9-of-10-8          335B ± 0%      335B ± 0%   +0.00%
DroppedRecords/99-of-100-8        335B ± 0%      335B ± 0%   +0.00%
DroppedRecords/999-of-1000-8      335B ± 0%      335B ± 0%   +0.00%

name                          old allocs/op  new allocs/op  delta
NopBaseline-8                     10.0 ± 0%       9.0 ± 0%  -10.00%
NopDisallowedLevel-8              10.0 ± 0%       5.0 ± 0%  -50.00%
NopAllowedLevel-8                 10.0 ± 0%       9.0 ± 0%  -10.00%
JSONBaseline-8                    23.0 ± 0%      23.0 ± 0%   +0.00%
JSONDisallowedLevel-8             10.0 ± 0%       5.0 ± 0%  -50.00%
JSONAllowedLevel-8                23.0 ± 0%      23.0 ± 0%   +0.00%
LogfmtBaseline-8                  10.0 ± 0%       9.0 ± 0%  -10.00%
LogfmtDisallowedLevel-8           10.0 ± 0%       5.0 ± 0%  -50.00%
LogfmtAllowedLevel-8              10.0 ± 0%       9.0 ± 0%  -10.00%
DroppedRecords/1-of-2-8           8.00 ± 0%      7.00 ± 0%  -12.50%
DroppedRecords/3-of-4-8           8.00 ± 0%      7.00 ± 0%  -12.50%
DroppedRecords/9-of-10-8          8.00 ± 0%      7.00 ± 0%  -12.50%
DroppedRecords/99-of-100-8        8.00 ± 0%      7.00 ± 0%  -12.50%
DroppedRecords/999-of-1000-8      8.00 ± 0%      7.00 ± 0%  -12.50%
```
It looks like I'll need to rebase atop #449 and #454.

seh:
> From an API design perspective Projections and Loggers are not orthogonal. I think deferred projections are redundant because the same choices could be made in the Logger called at the end of Context.Log without any extra overhead. 

I revised `level.EscalatingErrors` to yield a decorating `log.Logger` to confirm that we can replace deferring a projection with interposing a wrapping `log.Logger`.
```go
func withReplacement(orig []interface{}, i int, replacement interface{}) []interface{} {
	s := make([]interface{}, len(orig))
	copy(s, orig[:i])
	s[i] = replacement
	copy(s[i+1:], orig[i+1:])
	return s
}

func EscalatingErrors(logger log.Logger) log.Logger {
	return log.LoggerFunc(func(keyvals ...interface{}) error {
	loop:
		for i, end := 1, len(keyvals); i < end; i += 2 {
			switch v := keyvals[i].(type) {
			case *levelValue:
				if v != valError {
					index := i
					i += 2
					for ; i < end; i += 2 {
						if err, ok := keyvals[i].(error); ok && err != nil {
							return logger.Log(withReplacement(keyvals, index, valError)...)
						}
					}
				}
				break loop
			case error:
				if v != nil {
					i += 2
					for ; i < end; i += 2 {
						if l, ok := keyvals[i].(*levelValue); ok {
							if l != valError {
								return logger.Log(withReplacement(keyvals, i, valError)...)
							}
							break
						}
					}
					break loop
				}
			}
		}
		return logger.Log(keyvals)
	})
}
```
It looks mostly [as it did before](https://github.com/go-kit/kit/pull/453#issuecomment-278473364). Now let's look at a unit test.
```go
func TestEscalatingErrors(t *testing.T) {
	var buf bytes.Buffer
	var logger log.Logger
	logger = level.EscalatingErrors(level.AllowingInfoAndAbove(log.NewLogfmtLogger(&buf)))
	logger = level.Debug(logger)
	logger.Log("msg", "success", "err", nil)
	logger.Log("msg", "failure", "err", errors.New("failed"))
	if want, have := `level=error msg=failure err=failed`, strings.TrimSpace(buf.String()); want != have {
	}
}
```
It matters that the call to `level.EscalatingErrors` sits between the calls to `level.Debug` and `level.AllowingInfoAndAbove`. Consider this different test case:
```go
func TestImmediateProjectionYieldsInertLogger(t *testing.T) {
	var buf bytes.Buffer
	var logger log.Logger
	logger = level.AllowingInfoAndAbove(log.NewLogfmtLogger(&buf))
	logger = level.Debug(logger)
	// NB: logger is now nil.
	logger = level.EscalatingErrors(logger)
	logger.Log("msg", "success", "err", nil)
	logger.Log("msg", "failure", "err", errors.New("failed"))
	if buf.Len() > 0 {
		t.Errorf("want no output, have %q", buf.Bytes())
	}
}
```
Note that the call to `level.Debug` returns a nil `log.Logger`; it applies the projection _immediately_ to the new key/value pair pushed into the context, which, for `level.Debug`, is "level" and `level.valDebug`. The projection installed by `level.AllowingInfoAndAbove` catches that too-low level and decides that no log records with these values can ever be emitted, and renders the context produced by `level.Debug` inert.

Even though `level.EscalatingErorrs` will do its job and change the level of error-bearing candidate records, when it pushes them through to the logger returned by `level.Debug`, nothing more happens.

ChrisHines:
After more review and tinkering I find the simplified version of the projection idea much better, but I don't find it compelling enough to warrant the additional complexity it adds to the log package. Projections are not required to implement the log levels functionality with the exported API in this PR. Projections merely improve the marginal performance of the level filtering logic. Furthermore, the gains only represent a small fraction of the overall gains achieved by filtering unwanted levels prior to log record formatting. Which means we don't lose much and can always add projections later after more time to explore the idea.

I do really like this part of the PR though:
```go
type level byte

const (
	levelDebug level = 1 << iota
	levelInfo
	levelWarn
	levelError
)

type levelValue struct {
	level
	name string
}

func (v *levelValue) String() string {
	return v.name
}
```
I'd like to see this code used in the `experimental_level` package with appropriate API changes before it gets promoted. What do you think @peterbourgon?

peterbourgon:
@ChrisHines Nice. I also like that better than pushing plain strings around. I would be happy to take a PR with just that bit.

log/alt_experimental_level/benchmark_test.go
```diff
@@ -0,0 +1,85 @@
...
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		// Only this one will be retained.
+		info.Log("foo", "bar")
+		for ; droppedRecords != 0; droppedRecords-- {
+			debug.Log("baz", "quux")
+		}
+	}
+}
```
seh: I'm sorry again that it took me this long to respond.

You are right that I introduced a defect into the benchmark function when I moved some statements around, mistakenly mutating the function parameter _inside_ the benchmark loop. When I first wrote it, that loop that decays the "droppedRecords" parameter was inside a function called on from the benchmark loop.

In any case, your version does something different from what I had intended. Your drops a record when `testing.B.N` has a certain value. Instead, I intended to drop the "droppedRecords" parameters value and only emit one, so if we say we're benchmarking 99 dropped records, our test involves emitting one record and dropping 99, and doing so on each iteration of the benchmark.

Commit 27c16543e81b6b3481999c76fc9dc6ac2ade8453 restores the behavior I intended. However, the measurements aren't much different from yours:
```
name                          old time/op    new time/op    delta
DroppedRecords/1-of-2-8         2.01µs ± 0%    1.72µs ± 0%  -14.36%
DroppedRecords/3-of-4-8         2.56µs ± 0%    1.95µs ± 0%  -23.97%
DroppedRecords/9-of-10-8        4.14µs ± 0%    2.67µs ± 0%  -35.42%
DroppedRecords/99-of-100-8      27.6µs ± 0%    13.0µs ± 0%  -53.05%
DroppedRecords/999-of-1000-8     268µs ± 0%     116µs ± 0%  -56.65%

name                          old alloc/op   new alloc/op   delta
DroppedRecords/1-of-2-8           463B ± 0%      399B ± 0%  -13.82%
DroppedRecords/3-of-4-8           719B ± 0%      527B ± 0%  -26.70%
DroppedRecords/9-of-10-8        1.49kB ± 0%    0.91kB ± 0%  -38.74%
DroppedRecords/99-of-100-8      13.0kB ± 0%     6.7kB ± 0%  -48.71%
DroppedRecords/999-of-1000-8     128kB ± 0%      64kB ± 0%  -49.87%

name                          old allocs/op  new allocs/op  delta
DroppedRecords/1-of-2-8           12.0 ± 0%      10.0 ± 0%  -16.67%
DroppedRecords/3-of-4-8           20.0 ± 0%      16.0 ± 0%  -20.00%
DroppedRecords/9-of-10-8          44.0 ± 0%      34.0 ± 0%  -22.73%
DroppedRecords/99-of-100-8         404 ± 0%       304 ± 0%  -24.75%
DroppedRecords/999-of-1000-8     4.00k ± 0%     3.00k ± 0%  -24.98%
```

seh:
> I'd like to see this code used in the experimental_level package with appropriate API changes before it gets promoted.

Do you want me to attempt that change, @peterbourgon, or would you prefer to do so?

ChrisHines:
>In any case, your version does something different from what I had intended. Your drops a record when testing.B.N has a certain value. Instead, I intended to drop the "droppedRecords" parameters value and only emit one, so if we say we're benchmarking 99 dropped records, our test involves emitting one record and dropping 99, and doing so on each iteration of the benchmark.

I don't think that is the proper way to make a fair benchmark with Go's benchmarking tools. I guess it depends on what you are trying to measure, but it seems to me that each loop should make the same number of calls to `Log` for a fair measurement.

seh:
> it seems to me that each loop should make the same number of calls to Log for a fair measurement

Ah, but to which loop are you referring there? I don't see how my test doesn't make the same number of calls to `Log` per benchmark loop. Can you clarify your concern?

In my mind, the test case is, say, "emit 1 record, drop 99," and then allow `testing.B.N` to vary as required to do that 100-call procedure as many times as necessary until the measurements stabilize. In my test, `Log` gets called _B.N_ * _("droppedRecords" + 1)_ times.

ChrisHines:
`go test -bench` reports things as duration/op, bytes/op, or allocs/op. Each "op" is one iteration of the `testing.B.N` loop. So each `test.B.N` loop should represent one "op" of similar size to make the results comparable to each other. Comparing 2 calls to `Log` with 1000 calls to `Log` does not make sense to me.

seh:
> Comparing 2 calls to Log with 1000 calls to Log does not make sense to me.

I agree with you. What I still don't understand is where you think I'm trying to do that, or where I'm doing that mistakenly.

ChrisHines:
Your benchmarks call `debug.Log` multiple times per `testing.B.N` loop. So one "op" in your benchmark comprises 1 call to `info.Log` followed by a variable number of calls to `debug.Log`.

For example, here: https://github.com/seh/kit/blob/27c16543e81b6b3481999c76fc9dc6ac2ade8453/log/alt_experimental_level/benchmark_test.go#L81

seh:
> So one "op" in your benchmark comprises one call to `info.Log` followed by a variable number of calls to `debug.Log`.

Thank you for clarifying. Yes, I did consider the unit of operation in this case to be numerous calls to `Log`, because I was interested to see what percentage of the cost could be saved by dropping a few notable fractions of records (e.g. 1 out of every 10, or 1 out of every 100). If I understand your concern properly, it's that that's too much work to be considered as a single operation. Is that criticism aesthetic, or do you think it causes inaccuracy in the measurement?

ChrisHines:
I think we have different goals for what we are trying to measure. I want to know the amortized cost of a `Log` call. For my goal the criticism is fundamental. I don't think you had that goal in mind, so from your perspective I suppose it is more aesthetic. But I'd like to persuade you that it is important because keeping a consistent unit of operation allows deeper analysis. It gives meaning to the `time/op` data, while your approach only has meaning for the `delta` column.

In my benchmark results we can compare the results between both old and new implementations as well as between different ratios within an implementation. In other words, with a consistent "op" size we can compare the times across both rows and columns, not just columns.
```
name                          old time/op    new time/op    delta
DroppedRecords/1-of-2-8         1.93µs ± 1%    1.83µs ± 1%   -5.17%  (p=0.029 n=4+4)
DroppedRecords/3-of-4-8         1.09µs ± 2%    0.97µs ± 1%  -10.71%  (p=0.008 n=5+5)
DroppedRecords/9-of-10-8         591ns ± 3%     467ns ± 2%  -21.07%  (p=0.008 n=5+5)
DroppedRecords/99-of-100-8       297ns ± 6%     155ns ± 4%  -47.84%  (p=0.008 n=5+5)
DroppedRecords/999-of-1000-8     263ns ± 4%     130ns ± 6%  -50.68%  (p=0.008 n=5+5)
```
When I look at these numbers I clearly see that the new implementation filters records faster than the old, but also that the savings gained by filtering a higher ratio of log records dwarf the gains of faster filtering. I can also see that although the percentage difference between old and new is much higher when most log events are discarded, the absolute speed difference is fairly constant across all ratios. That fact tells me that the gains come from a reduction in O(1) overhead rather than a O(n) scaling factor.

seh:
Thank you for explaining. I see your point now.

My concern with [your proposed revision sketched here](https://github.com/go-kit/kit/pull/453#discussion_r101188582) is that it assumes that `B.N` ≫ "droppedRecords." Perhaps this is not a justifiable concern, but if "droppedRecords" is 999 but _go test_ decides that `B.N` of, say, 100 iterations is enough to stabilize, then we really only tested dropping one out of 100 records, even though the benchmark claims (by its name) that it was testing one of every 1,000 records. Until we've gone through at least 101 iterations, those two cases are indistinguishable.

ChrisHines:
That is an excellent point. Your approach guarantees that the desired ratio is measured exactly. To preserve both your goal and my goal all of the `DroppedRecords` sub tests should call `Log` 1000 times inside the `testing.B.N` loop and adjust the ratio of `Info` to `Debug` calls appropriately. Luckily the ratios you have chosen all have denominators that are divisors of 1000. Also the test names should provide a clue that the batch size is 1000.

As a side note: In case you were not aware (by default) `go test -bench` searches for an `N` that will take at least 1 second (and not some idea of "stability"). Thus, doing more work in each loop just decreases the eventual value of `N` that `go test` chooses. It also [likes to pick values of `N` that are multiples of powers of 10](https://github.com/golang/go/blob/master/src/testing/benchmark.go#L187), so I was probably getting lucky that it picked multiples of 1000 already.

ChrisHines:
>>I'd like to see this code used in the experimental_level package with appropriate API changes before it gets promoted.

>Do you want me to attempt that change, @peterbourgon, or would you prefer to do so?

I'd like to work on it.

>edit: just to be sure—it looks like it would also involve changes to the experimental_level.logger.Log method, which would be good to before/after benchmark so that we don't regress.

Agreed. I've already done a quick check on that and I believe it performs slightly better than the current approach. I'll post numbers once I have a for real implementation though.

seh:
One thing to consider when adapting the typed level value (`levelValue`) technique: It's still possible for a caller to sneak in a "leveled" log record by using strings:
```go
level.Debug(logger).Log("msg", "You'll never see this.")
logger.Log("level", "debug", "msg", "But this one sneaks by.")
```
I expected a reviewer might bring this up as a concern. My defense is that we're not trying to prevent certain key/value pairs from making it into the log records; we're trying to help inject those key/value pairs into the log records, and make filtering them possible and efficient.

Still, I just want to make sure that you take this vulnerability into consideration, and decide whether it's tolerable.

peterbourgon:
We've moved forward with an alternate implementation. Thanks to @seh and everyone involved for the incredible discussion, experimentation, and leveling-up — pun intended.

Issue #427, opened 2017-01-10
Go kit 0.4.0 release


peterbourgon:
**Edit**: Due to lots of feature churn in the last few weeks/months, I've been convinced that it makes more sense to hold off on a 1.0.0 until things settle. Notes in comments below.

—

Hello! I want to cut a v1.0.0 release in the near future. This issue is created per [the request](https://github.com/go-kit/kit/pull/368#issuecomment-271614515) of @ChrisHines to serve as a notice to anyone who may be interested, for example to raise questions, concerns, or propose specific additions. 

The only thing outstanding that I'd like to get in is a promotion of log/experimental_level to log/level. I've asked @seh to weigh in on that, as he's previously commented on the package. Otherwise, let me know your thoughts. I'll leave this open until at least the weekend, so another 5 days.

_edit_: outstanding issues I'd want to resolve before a 1.0.0

- [x] #476 — Merge log/experimental_level to level; move existing levels to deprecated_levels
- [x] #420 / #421 — use stdlib context where possible
- [x] #432 / #437 — log: StdlibAdapter "file" key to "caller"
- [x] Optional: #434 — log: normalize StdlibAdapter timestamp format
- [x] Optional: #431 — transport: implement http.EncodeJSONRequest
- [x] Evaluate breaking change: #474 — Replace log.Context with log.With/WithPrefix?

If anyone feels strongly about either of those two Optionals, (cc: @willfaught) please get your PRs in within the next few days...

pruggera:
Is some kind of connector architecture planned?  If so, shouldn't this not
be part version 1?

peterbourgon:
What is a connector architecture?

ChrisHines:
What will be the compatibility policy after v1.0.0 has been released?

pruggera:
Connectors would be packages that implement the endpoint, transport, security, and other policies required by calling or called SaaS services like Salesforce.com,  Workday,  ServiceNow, LinkedIn, or Marketo.  The connector architecture describes the wrapping of a SaaS service to call them or the wrapping of the go-kit service to be called by a SaaS services.

peterbourgon:
@pruggera No, nothing like this is planned.

@ChrisHines I plan to follow typical SemVer rules. Do you think that is sufficient?

ChrisHines:
>I plan to follow typical SemVer rules. Do you think that is sufficient?

That's good. We shall see if it is sufficient. ;) Do we want to document this anywhere?

seh:
[...]

PR #454, opened 2017-02-09
Use variadic function to pass parameters instead of Config struct

groob:
Using function to pass optional params makes the API more consistent
with the rest of go-kit packages.

For #427 

Following up on discussion in https://github.com/go-kit/kit/issues/427#issuecomment-275938942
> > Why not use functional options for level.New instead of Config?

> There's no practical difference, but I can in hindsight see the value of using functional options for consistency with the rest of package log. I'd accept a PR :)

I thought it was important to get this change in before #449 and ultimately 1.0.0 to avoid making breaking changes in a stable API.

### DIFF
--- a/log/experimental_level/level.go
+++ b/log/experimental_level/level.go
@@ -69,6 +69,8 @@ func Debug(logger log.Logger) log.Logger {
 
 // New wraps the logger and implements level checking. See the commentary on the
 // Option functions for a detailed description of how to configure levels.
+// If no options are provided, all leveled log events created with level.Debug,
+// Info, Warn or Error helper methods will be squelched.
 func New(next log.Logger, options ...Option) log.Logger {
 	l := logger{
 		next: next,
--- a/log/experimental_level/level.go
+++ b/log/experimental_level/level.go
@@ -18,31 +18,31 @@ func AllowAll() []string {
 }
 
 // AllowDebugAndAbove allows all of the four default log levels.
-// Its return value may be provided as the Allowed parameter in the Config.
+// Its return value may be provided with the Allowed Option.
 func AllowDebugAndAbove() []string {
 	return []string{errorLevelValue, warnLevelValue, infoLevelValue, debugLevelValue}
 }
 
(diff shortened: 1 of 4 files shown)


log/experimental_level/level.go
```diff
@@ -67,42 +67,50 @@ func Debug(logger log.Logger) log.Logger {
...
 }
 
-// Config parameterizes the leveled logger.
-type Config struct {
-	// Allowed enumerates the accepted log levels. If a log event is encountered
-	// with a level key set to a value that isn't explicitly allowed, the event
-	// will be squelched, and ErrNotAllowed returned.
-	Allowed []string
+// New wraps the logger and implements level checking. See the commentary on the
+// Option functions for a detailed description of how to configure levels.
```
peterbourgon: Would you mind adding an extra sentence of commentary describing the default behavior with no options set?

log/experimental_level/level_test.go
```diff
@@ -75,10 +75,11 @@ func TestVariousLevels(t *testing.T) {
...
 	myError := errors.New("squelched!")
-	logger := level.New(log.NewNopLogger(), level.Config{
-		Allowed:       level.AllowWarnAndAbove(),
-		ErrNotAllowed: myError,
-	})
+	opts := []level.Option{
+		level.Allowed(level.AllowWarnAndAbove()),
+		level.ErrNotAllowed(myError),
+	}
+	logger := level.New(log.NewNopLogger(), opts...)
```
peterbourgon: Any reason not to inline these? I'm not fussed, just curious.

groob: No specific reason. I usually prefer to create a slice when I have multiple options because it looks more organized, especially if there's a chance the list of options will grow in the future. 

I'd be glad to change it if you have the opposite preference.

peterbourgon: Cool! No worries.

peterbourgon:
Brilliant, thanks for the contribution!

PR #465, opened 2017-02-16
Allow calling ServerBefore/After multiple times when creating ServerOptions

djgilcrease:
This allows you to set some default options that multiple httptransport.NewServer's use, but still allows each handler to specify it's own as well

```
router = mux.NewRouter().StrictSlash(false)
options := []httptransport.ServerOption{
		httptransport.ServerErrorLogger(logger),
		httptransport.ServerErrorEncoder(httptransport.DefaultErrorEncoder),
		httptransport.ServerBefore(
			prepareRequestContext,
		),
	}
router = router.Methods("GET").Path("/path-A").Handler(httptransport.NewServer(
		ctx,
		endpointA,
		decodeRequest,
		encodeResponse,
		append(
			options,
			httptransport.ServerBefore(
				opentracing.FromHTTPRequest(tracer, "path-A", logger),
			),
		)...,
	))
router = router.Methods("GET").Path("/path-B").Handler(httptransport.NewServer(
		ctx,
		endpointB,
		decodeRequest,
		encodeResponse,
		append(
			options,
			httptransport.ServerBefore(
				opentracing.FromHTTPRequest(tracer, "path-B", logger),
			),
		)...,
	))
```
Without this change you would have to make sure that any global ServerBefore or ServerAfter methods we added to every handler

### DIFF
--- a/transport/grpc/server.go
+++ b/transport/grpc/server.go
@@ -59,13 +59,13 @@ type ServerOption func(*Server)
 // ServerBefore functions are executed on the HTTP request object before the
 // request is decoded.
 func ServerBefore(before ...RequestFunc) ServerOption {
-	return func(s *Server) { s.before = before }
+	return func(s *Server) { s.before = append(s.before, before...) }
 }
 
 // ServerAfter functions are executed on the HTTP response writer after the
 // endpoint is invoked, but before anything is written to the client.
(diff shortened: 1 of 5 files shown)


peterbourgon:
Ah, neat! Would you mind adding a quick test that would fail with the previous implementation but succeed with this one?

PR #466, opened 2017-02-16
Add Zipkin HTTP Collector to addsvc example

basvanbeek:
To make it easier for people to testdrive Zipkin an HTTP Collector set-up has been added to the `addsvc` example. 

This allows people to testdrive Zipkin with Go-kit using the official Zipkin docker image without the need of running Kafka and Zookeeper.


peterbourgon:
Yes, brilliant.

PR #467, opened 2017-02-16
Add the ability to specify ServerBefore & ServerAfter multiple times in the ServerOption section

djgilcrease:
This makes it so you can set some global ServerBefore or After options without having to make sure you specify them individually on every handler


djgilcrease:
This is just a cleanup and squash of the previous pull request

peterbourgon:
Awesome! Thank you!

PR #468, opened 2017-02-20
Remove drone due to deprecation

17twenty:
Drone.io disbanded their service and so this diff removes the drone.io status badge.


peterbourgon:
Oh no! That's sad :(

PR #469, opened 2017-02-20
Fix code typo in zipkin readme

frozzare:
PeerHostname line has a comma instead of a dot and that's not working :)


basvanbeek:
oops.... thanks...

PR #470, opened 2017-02-23
Reimplement log/experimental_levels

ChrisHines:
The main focus here is to get away from identifying level keyvals with string matching and use type matching with a bit set for level filtering in the manner demonstrated by @seh in #453.

I am very happy that all of the configuration options could be preserved. I have taken the liberty to simplify the `Options` API a bit to make it friendlier. The `Allowed` option was removed in favor of converting all of its helper functions (e.g. `AllowInfoAndAbove`) into functional Options that can be used directly. I also changed `level.New` to `level.NewFilter` because it better expresses its purpose.

The documentation is updated and corrected.

Before updating the implementation I rewrote the benchmarks to be more comprehensive with the use of the new (Go 1.7) sub-benchmark feature. The before and after results of the benchmarks follow.

Updated benchmarks after 6a12b3162160a71abf7458cdfccdc8aa0cca0ba0.
```
name                                              old time/op    new time/op    delta
/NoContext/Baseline/Nop-8                            434ns ±13%     371ns ± 7%  -14.55%  (p=0.032 n=5+5)
/NoContext/Baseline/Logfmt-8                         796ns ± 1%     828ns ± 1%   +3.99%  (p=0.008 n=5+5)
/NoContext/Baseline/JSON-8                          2.06µs ± 3%    2.06µs ± 2%     ~     (p=0.690 n=5+5)
/NoContext/DisallowedLevel/Nop-8                     422ns ± 6%     368ns ±19%     ~     (p=0.095 n=5+5)
/NoContext/DisallowedLevel/Logfmt-8                  435ns ± 9%     377ns ±12%     ~     (p=0.056 n=5+5)
/NoContext/DisallowedLevel/JSON-8                    422ns ±11%     355ns ± 4%  -15.89%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/Nop-8                        447ns ± 8%     343ns ± 9%  -23.20%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/Logfmt-8                     824ns ± 1%     833ns ± 2%     ~     (p=0.111 n=5+5)
/NoContext/AllowedLevel/JSON-8                      2.07µs ± 3%    2.05µs ± 1%     ~     (p=0.690 n=5+5)
/TimeContext/Baseline/Nop-8                          780ns ± 4%     702ns ± 1%   -9.95%  (p=0.008 n=5+5)
/TimeContext/Baseline/Logfmt-8                      1.46µs ± 2%    1.47µs ± 1%     ~     (p=0.222 n=5+5)
/TimeContext/Baseline/JSON-8                        3.00µs ± 1%    2.99µs ± 4%     ~     (p=0.690 n=5+5)
/TimeContext/DisallowedLevel/Nop-8                   799ns ± 3%     711ns ± 2%  -11.06%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/Logfmt-8                796ns ± 3%     698ns ± 0%  -12.23%  (p=0.000 n=5+4)
/TimeContext/DisallowedLevel/JSON-8                  792ns ± 1%     714ns ± 3%   -9.85%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/Nop-8                      806ns ± 5%     717ns ± 1%  -10.97%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/Logfmt-8                  1.47µs ± 1%    1.48µs ± 2%     ~     (p=0.056 n=5+5)
/TimeContext/AllowedLevel/JSON-8                    2.99µs ± 3%    2.94µs ± 0%     ~     (p=0.206 n=5+5)
/CallerContext/Baseline/Nop-8                       1.03µs ± 3%    0.97µs ± 1%   -6.45%  (p=0.008 n=5+5)
/CallerContext/Baseline/Logfmt-8                    2.31µs ± 1%    2.33µs ± 1%     ~     (p=0.246 n=5+5)
/CallerContext/Baseline/JSON-8                      3.86µs ± 1%    3.88µs ± 1%     ~     (p=0.286 n=5+5)
/CallerContext/DisallowedLevel/Nop-8                1.04µs ± 0%    0.99µs ± 3%   -4.61%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/Logfmt-8             1.08µs ± 6%    0.97µs ± 2%  -10.10%  (p=0.016 n=5+4)
/CallerContext/DisallowedLevel/JSON-8               1.04µs ± 0%    1.14µs ±29%     ~     (p=1.000 n=4+5)
/CallerContext/AllowedLevel/Nop-8                   1.04µs ± 0%    1.06µs ± 4%     ~     (p=0.698 n=4+5)
/CallerContext/AllowedLevel/Logfmt-8                2.38µs ± 4%    2.79µs ±13%  +17.09%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/JSON-8                  4.03µs ± 3%    4.44µs ± 5%  +10.29%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/Nop-8              1.57µs ±10%    1.57µs ± 8%     ~     (p=1.000 n=5+5)
/TimeCallerReqIDContext/Baseline/Logfmt-8           3.41µs ± 3%    3.56µs ± 1%   +4.56%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/JSON-8             5.60µs ± 2%    5.76µs ± 4%     ~     (p=0.056 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/Nop-8       1.50µs ± 1%    1.46µs ± 2%   -2.69%  (p=0.032 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/Logfmt-8    1.52µs ± 2%    1.46µs ± 1%   -3.86%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/JSON-8      1.50µs ± 1%    1.46µs ± 2%   -2.67%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/Nop-8          1.51µs ± 1%    1.46µs ± 1%   -3.11%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/Logfmt-8       3.45µs ± 3%    3.49µs ± 1%     ~     (p=0.151 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/JSON-8         5.55µs ± 2%    5.80µs ± 2%   +4.42%  (p=0.008 n=5+5)

name                                              old alloc/op   new alloc/op   delta
/NoContext/Baseline/Nop-8                             288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/Baseline/Logfmt-8                          288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/Baseline/JSON-8                          1.06kB ± 0%    1.05kB ± 0%   -1.50%  (p=0.008 n=5+5)
/NoContext/DisallowedLevel/Nop-8                      288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/DisallowedLevel/Logfmt-8                   288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/DisallowedLevel/JSON-8                     288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/Nop-8                         288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/Logfmt-8                      288B ± 0%      256B ± 0%  -11.11%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/JSON-8                      1.06kB ± 0%    1.05kB ± 0%   -1.50%  (p=0.008 n=5+5)
/TimeContext/Baseline/Nop-8                           384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/Baseline/Logfmt-8                        384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/Baseline/JSON-8                        1.28kB ± 0%    1.26kB ± 0%   -1.25%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/Nop-8                    384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/Logfmt-8                 384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/JSON-8                   384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/Nop-8                       384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/Logfmt-8                    384B ± 0%      352B ± 0%   -8.33%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/JSON-8                    1.28kB ± 0%    1.26kB ± 0%   -1.25%  (p=0.008 n=5+5)
/CallerContext/Baseline/Nop-8                         352B ± 0%      320B ± 0%   -9.09%  (p=0.008 n=5+5)
/CallerContext/Baseline/Logfmt-8                      520B ± 0%      488B ± 0%   -6.15%  (p=0.008 n=5+5)
/CallerContext/Baseline/JSON-8                      1.42kB ± 0%    1.40kB ± 0%   -1.13%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/Nop-8                  352B ± 0%      320B ± 0%   -9.09%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/Logfmt-8               352B ± 0%      320B ± 0%   -9.09%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/JSON-8                 352B ± 0%      320B ± 0%   -9.09%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/Nop-8                     352B ± 0%      320B ± 0%   -9.09%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/Logfmt-8                  520B ± 0%      488B ± 0%   -6.15%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/JSON-8                  1.42kB ± 0%    1.40kB ± 0%   -1.13%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/Nop-8                592B ± 0%      560B ± 0%   -5.41%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/Logfmt-8             768B ± 0%      736B ± 0%   -4.17%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/JSON-8             1.86kB ± 0%    1.85kB ± 0%   -0.86%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/Nop-8         592B ± 0%      560B ± 0%   -5.41%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/Logfmt-8      592B ± 0%      560B ± 0%   -5.41%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/JSON-8        592B ± 0%      560B ± 0%   -5.41%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/Nop-8            592B ± 0%      560B ± 0%   -5.41%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/Logfmt-8         768B ± 0%      736B ± 0%   -4.17%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/JSON-8         1.86kB ± 0%    1.85kB ± 0%   -0.86%  (p=0.008 n=5+5)

name                                              old allocs/op  new allocs/op  delta
/NoContext/Baseline/Nop-8                             9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/Baseline/Logfmt-8                          9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/Baseline/JSON-8                            24.0 ± 0%      23.0 ± 0%   -4.17%  (p=0.008 n=5+5)
/NoContext/DisallowedLevel/Nop-8                      9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/DisallowedLevel/Logfmt-8                   9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/DisallowedLevel/JSON-8                     9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/Nop-8                         9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/Logfmt-8                      9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/NoContext/AllowedLevel/JSON-8                        24.0 ± 0%      23.0 ± 0%   -4.17%  (p=0.008 n=5+5)
/TimeContext/Baseline/Nop-8                           10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/Baseline/Logfmt-8                        10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/Baseline/JSON-8                          28.0 ± 0%      27.0 ± 0%   -3.57%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/Nop-8                    10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/Logfmt-8                 10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/DisallowedLevel/JSON-8                   10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/Nop-8                       10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/Logfmt-8                    10.0 ± 0%       8.0 ± 0%  -20.00%  (p=0.008 n=5+5)
/TimeContext/AllowedLevel/JSON-8                      28.0 ± 0%      27.0 ± 0%   -3.57%  (p=0.008 n=5+5)
/CallerContext/Baseline/Nop-8                         9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/CallerContext/Baseline/Logfmt-8                      13.0 ± 0%      11.0 ± 0%  -15.38%  (p=0.008 n=5+5)
/CallerContext/Baseline/JSON-8                        31.0 ± 0%      30.0 ± 0%   -3.23%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/Nop-8                  9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/Logfmt-8               9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/CallerContext/DisallowedLevel/JSON-8                 9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/Nop-8                     9.00 ± 0%      7.00 ± 0%  -22.22%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/Logfmt-8                  13.0 ± 0%      11.0 ± 0%  -15.38%  (p=0.008 n=5+5)
/CallerContext/AllowedLevel/JSON-8                    31.0 ± 0%      30.0 ± 0%   -3.23%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/Nop-8                11.0 ± 0%       9.0 ± 0%  -18.18%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/Logfmt-8             16.0 ± 0%      14.0 ± 0%  -12.50%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/Baseline/JSON-8               39.0 ± 0%      38.0 ± 0%   -2.56%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/Nop-8         11.0 ± 0%       9.0 ± 0%  -18.18%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/Logfmt-8      11.0 ± 0%       9.0 ± 0%  -18.18%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/DisallowedLevel/JSON-8        11.0 ± 0%       9.0 ± 0%  -18.18%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/Nop-8            11.0 ± 0%       9.0 ± 0%  -18.18%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/Logfmt-8         16.0 ± 0%      14.0 ± 0%  -12.50%  (p=0.008 n=5+5)
/TimeCallerReqIDContext/AllowedLevel/JSON-8           39.0 ± 0%      38.0 ± 0%   -2.56%  (p=0.008 n=5+5)
```

### DIFF
--- a/log/experimental_level/level.go
+++ b/log/experimental_level/level.go
@@ -1,142 +1,58 @@
 package level
 
-import (
-	"github.com/go-kit/kit/log"
-)
+import "github.com/go-kit/kit/log"
 
-var (
-	levelKey        = "level"
-	errorLevelValue = "error"
-	warnLevelValue  = "warn"
-	infoLevelValue  = "info"
-	debugLevelValue = "debug"
-)
-
-// AllowAll is an alias for AllowDebugAndAbove.
-func AllowAll() []string {
-	return AllowDebugAndAbove()
-}
-
-// AllowDebugAndAbove allows all of the four default log levels.
-// Its return value may be provided with the Allowed Option.
-func AllowDebugAndAbove() []string {
-	return []string{errorLevelValue, warnLevelValue, infoLevelValue, debugLevelValue}
-}
-
-// AllowInfoAndAbove allows the default info, warn, and error log levels.
-// Its return value may be provided with the Allowed Option.
-func AllowInfoAndAbove() []string {
-	return []string{errorLevelValue, warnLevelValue, infoLevelValue}
-}
-
-// AllowWarnAndAbove allows the default warn and error log levels.
-// Its return value may be provided with the Allowed Option.
-func AllowWarnAndAbove() []string {
-	return []string{errorLevelValue, warnLevelValue}
-}
-
-// AllowErrorOnly allows only the default error log level.
-// Its return value may be provided with the Allowed Option.
-func AllowErrorOnly() []string {
-	return []string{errorLevelValue}
-}
-
-// AllowNone allows none of the default log levels.
-// Its return value may be provided with the Allowed Option.
-func AllowNone() []string {
-	return []string{}
-}
-
-// Error returns a logger with the level key set to ErrorLevelValue.
+// Error returns a logger that includes an error level keyval.
 func Error(logger log.Logger) log.Logger {
 	return log.NewContext(logger).WithPrefix(levelKey, errorLevelValue)
 }
 
-// Warn returns a logger with the level key set to WarnLevelValue.
+// Warn returns a logger that includes a warn level keyval.
 func Warn(logger log.Logger) log.Logger {
 	return log.NewContext(logger).WithPrefix(levelKey, warnLevelValue)
 }
 
-// Info returns a logger with the level key set to InfoLevelValue.
+// Info returns a logger that includes an info level keyval.
 func Info(logger log.Logger) log.Logger {
 	return log.NewContext(logger).WithPrefix(levelKey, infoLevelValue)
 }
 
-// Debug returns a logger with the level key set to DebugLevelValue.
+// Debug returns a logger that includes a debug level keyval.
 func Debug(logger log.Logger) log.Logger {
 	return log.NewContext(logger).WithPrefix(levelKey, debugLevelValue)
 }
 
-// New wraps the logger and implements level checking. See the commentary on the
-// Option functions for a detailed description of how to configure levels.
-// If no options are provided, all leveled log events created with level.Debug,
-// Info, Warn or Error helper methods will be squelched.
-func New(next log.Logger, options ...Option) log.Logger {
-	l := logger{
+// NewFilter wraps next and implements level filtering. See the commentary on
+// the Option functions for a detailed description of how to configure levels.
+// If no options are provided, all leveled log events created with Debug,
+// Info, Warn or Error helper methods are squelched and non-leveled log
+// events are passed to next unmodified.
+func NewFilter(next log.Logger, options ...Option) log.Logger {
+	l := &logger{
 		next: next,
 	}
 	for _, option := range options {
-		option(&l)
+		option(l)
 	}
-	return &l
+	return l
 }
-
-// Allowed enumerates the accepted log levels. If a log event is encountered
-// with a level key set to a value that isn't explicitly allowed, the event
-// will be squelched, and ErrNotAllowed returned.
-func Allowed(allowed []string) Option {
-	return func(l *logger) { l.allowed = makeSet(allowed) }
-}
-
-// ErrNoLevel is returned to the caller when SquelchNoLevel is true, and Log
-// is invoked without a level key. By default, ErrNoLevel is nil; in this
-// case, the log event is squelched with no error.
-func ErrNotAllowed(err error) Option {
-	return func(l *logger) { l.errNotAllowed = err }
-}
-
-// SquelchNoLevel will squelch log events with no level key, so that they
-// don't proceed through to the wrapped logger. If SquelchNoLevel is set to
-// true and a log event is squelched in this way, ErrNoLevel is returned to
-// the caller.
-func SquelchNoLevel(squelch bool) Option {
-	return func(l *logger) { l.squelchNoLevel = squelch }
-}
-
-// ErrNoLevel is returned to the caller when SquelchNoLevel is true, and Log
-// is invoked without a level key. By default, ErrNoLevel is nil; in this
-// case, the log event is squelched with no error.
-func ErrNoLevel(err error) Option {
-	return func(l *logger) { l.errNoLevel = err }
-}
-
-// Option sets a parameter for the leveled logger.
-type Option func(*logger)
 
 type logger struct {
 	next           log.Logger
-	allowed        map[string]struct{}
-	errNotAllowed  error
+	allowed        level
 	squelchNoLevel bool
+	errNotAllowed  error
 	errNoLevel     error
 }
 
 func (l *logger) Log(keyvals ...interface{}) error {
 	var hasLevel, levelAllowed bool
-	for i := 0; i < len(keyvals); i += 2 {
-		if k, ok := keyvals[i].(string); !ok || k != levelKey {
-			continue
+	for i := 1; i < len(keyvals); i += 2 {
+		if v, ok := keyvals[i].(*levelValue); ok {
+			hasLevel = true
+			levelAllowed = l.allowed&v.level != 0
+			break
 		}
-		hasLevel = true
-		if i >= len(keyvals) {
-			continue
-		}
-		v, ok := keyvals[i+1].(string)
-		if !ok {
-			continue
-		}
-		_, levelAllowed = l.allowed[v]
-		break
 	}
 	if !hasLevel && l.squelchNoLevel {
 		return l.errNoLevel
@@ -147,10 +63,88 @@ func (l *logger) Log(keyvals ...interface{}) error {
 	return l.next.Log(keyvals...)
 }
 
-func makeSet(a []string) map[string]struct{} {
-	m := make(map[string]struct{}, len(a))
-	for _, s := range a {
-		m[s] = struct{}{}
-	}
-	return m
+// Option sets a parameter for the leveled logger.
+type Option func(*logger)
+
+// AllowAll is an alias for AllowDebugAndAbove.
+func AllowAll() Option {
+	return AllowDebugAndAbove()
+}
+
+// AllowDebugAndAbove allows all of the four default log levels.
+func AllowDebugAndAbove() Option {
+	return allowed(levelDebug | levelInfo | levelWarn | levelError)
+}
+
+// AllowInfoAndAbove allows the default info, warn, and error log levels.
+func AllowInfoAndAbove() Option {
+	return allowed(levelInfo | levelWarn | levelError)
+}
+
+// AllowWarnAndAbove allows the default warn and error log levels.
+func AllowWarnAndAbove() Option {
+	return allowed(levelWarn | levelError)
+}
+
+// AllowErrorOnly allows only the default error log level.
+func AllowErrorOnly() Option {
+	return allowed(levelError)
+}
+
+// AllowNone allows none of the default log levels.
+func AllowNone() Option {
+	return allowed(0)
+}
+
+func allowed(allowed level) Option {
+	return func(l *logger) { l.allowed = allowed }
+}
+
+// ErrNotAllowed sets the error to return from Log when it squelches a log
+// event below the configured filtering level. By default, ErrNotAllowed is
+// nil; in this case, the log event is squelched with no error.
+func ErrNotAllowed(err error) Option {
+	return func(l *logger) { l.errNotAllowed = err }
+}
+
+// SquelchNoLevel instructs Log to squelch log events with no level, so that
+// they don't proceed through to the wrapped logger. If SquelchNoLevel is set
+// to true and a log event is squelched in this way, the error value
+// configured with ErrNoLevel is returned to the caller.
+func SquelchNoLevel(squelch bool) Option {
+	return func(l *logger) { l.squelchNoLevel = squelch }
+}
+
+// ErrNoLevel sets the error to return from Log when it squelches a log event
+// with no level. By default, ErrNoLevel is nil; in this case, the log event
+// is squelched with no error.
+func ErrNoLevel(err error) Option {
+	return func(l *logger) { l.errNoLevel = err }
+}
+
+const levelKey = "level"
+
+var (
+	errorLevelValue = &levelValue{level: levelError, name: "error"}
+	warnLevelValue  = &levelValue{level: levelWarn, name: "warn"}
+	infoLevelValue  = &levelValue{level: levelInfo, name: "info"}
+	debugLevelValue = &levelValue{level: levelDebug, name: "debug"}
+)
+
+type level byte
+
+const (
+	levelDebug level = 1 << iota
+	levelInfo
+	levelWarn
+	levelError
+)
+
+type levelValue struct {
+	name string
+	level
+}
+
+func (v *levelValue) String() string {
+	return v.name
 }
--- a/log/experimental_level/level.go
+++ b/log/experimental_level/level.go
@@ -124,7 +124,8 @@ func ErrNoLevel(err error) Option {
 }
 
 // NewInjector wraps next and returns a logger that adds a Key/level pair to
-// the beginning of log events that don't already contain a level.
+// the beginning of log events that don't already contain a level. In effect,
+// this gives a default level to logs without a level.
 func NewInjector(next log.Logger, level Value) log.Logger {
 	return &injector{
 		next:  next,
--- a/log/experimental_level/level.go
+++ b/log/experimental_level/level.go
@@ -2,24 +2,24 @@ package level
 
 import "github.com/go-kit/kit/log"
 
-// Error returns a logger that includes an error level keyval.
+// Error returns a logger that includes a Key/ErrorValue pair.
 func Error(logger log.Logger) log.Logger {
-	return log.NewContext(logger).WithPrefix(levelKey, errorLevelValue)
+	return log.NewContext(logger).WithPrefix(key, errorValue)
 }
 
-// Warn returns a logger that includes a warn level keyval.
+// Warn returns a logger that includes a Key/WarnValue pair.
 func Warn(logger log.Logger) log.Logger {
-	return log.NewContext(logger).WithPrefix(levelKey, warnLevelValue)
+	return log.NewContext(logger).WithPrefix(key, warnValue)
 }
 
-// Info returns a logger that includes an info level keyval.
+// Info returns a logger that includes a Key/InfoValue pair.
 func Info(logger log.Logger) log.Logger {
-	return log.NewContext(logger).WithPrefix(levelKey, infoLevelValue)
+	return log.NewContext(logger).WithPrefix(key, infoValue)
 }
 
-// Debug returns a logger that includes a debug level keyval.
+// Debug returns a logger that includes a Key/DebugValue pair.
 func Debug(logger log.Logger) log.Logger {
-	return log.NewContext(logger).WithPrefix(levelKey, debugLevelValue)
+	return log.NewContext(logger).WithPrefix(key, debugValue)
 }
 
 // NewFilter wraps next and implements level filtering. See the commentary on
@@ -123,12 +123,47 @@ func ErrNoLevel(err error) Option {
 	return func(l *logger) { l.errNoLevel = err }
 }
 
+func NewInjector(next log.Logger, level Value) log.Logger {
+	return &injector{
+		next:  next,
+		level: level,
+	}
+}
+
+type injector struct {
+	next  log.Logger
+	level interface{}
+}
+
+func (l *injector) Log(keyvals ...interface{}) error {
+	for i := 1; i < len(keyvals); i += 2 {
+		if _, ok := keyvals[i].(*levelValue); ok {
+			return l.next.Log(keyvals...)
+		}
+	}
+	kvs := make([]interface{}, len(keyvals)+2)
+	kvs[0], kvs[1] = key, l.level
+	copy(kvs[2:], keyvals)
+	return l.next.Log(kvs...)
+}
+
+type Value interface {
+	String() string
+	levelVal()
+}
+
+func Key() interface{}  { return key }
+func ErrorValue() Value { return errorValue }
+func WarnValue() Value  { return warnValue }
+func InfoValue() Value  { return infoValue }
+func DebugValue() Value { return debugValue }
+
 var (
-	levelKey        interface{} = "level"
-	errorLevelValue             = &levelValue{level: levelError, name: "error"}
-	warnLevelValue              = &levelValue{level: levelWarn, name: "warn"}
-	infoLevelValue              = &levelValue{level: levelInfo, name: "info"}
-	debugLevelValue             = &levelValue{level: levelDebug, name: "debug"}
+	key        interface{} = "level"
+	errorValue             = &levelValue{level: levelError, name: "error"}
+	warnValue              = &levelValue{level: levelWarn, name: "warn"}
+	infoValue              = &levelValue{level: levelInfo, name: "info"}
+	debugValue             = &levelValue{level: levelDebug, name: "debug"}
 )
 
 type level byte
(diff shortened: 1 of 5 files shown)


log/experimental_level/level.go
```diff
@@ -1,142 +1,58 @@
...
-// Info, Warn or Error helper methods will be squelched.
-func New(next log.Logger, options ...Option) log.Logger {
-	l := logger{
+// NewFilter wraps next and implements level filtering. See the commentary on
+// the Option functions for a detailed description of how to configure levels.
+// If no options are provided, all leveled log events created with Debug,
+// Info, Warn or Error helper methods are squelched and non-leveled log
+// events are passed to next unmodified.
+func NewFilter(next log.Logger, options ...Option) log.Logger {
+	l := &logger{
```
groob: Any technical reason for moving the pointer here instead of passing it as `option(&l)`?
After some discussions with @goinggo I've taken to passing with `&` because it implies the semantic meaning of "I'm sharing this to be modified" 

Just interested in the discussion on readability and performance with the declarations here :)

ChrisHines: I don't think there is any technical reason. The reason I moved it is because type `logger` has a pointer receiver on its `Log` methoc so it must be a pointer in the returned `log.Logger` interface value to satisfy the interface. I prefer to create it in the same form that it will live its life, as a `*logger`.

log/experimental_level/level.go
```diff
@@ -147,10 +63,87 @@ func (l *logger) Log(keyvals ...interface{}) error {
...
+
+// AllowAll is an alias for AllowDebugAndAbove.
+func AllowAll() Option {
+	return AllowDebugAndAbove()
+}
+
+// AllowDebugAndAbove allows all of the four default log levels.
+func AllowDebugAndAbove() Option {
+	return allowed(levelDebug | levelInfo | levelWarn | levelError)
+}
```
peterbourgon: Looking at this through the fresh eyes of someone reviewing the Proposal doc, I'm raising my eyebrow a little bit at the AllowFooAndAbove nomenclature, specifically the _AndAbove_ bit, which seems necessarily implicit. Therefore the bitfield implementation here, which allows arbitrary selection of levels, may be needless. What do you think about [this patch](https://gist.github.com/peterbourgon/d8efcb79812e5c03373f1c2a3abd4f56)?

seh: @peterbourgon, that's the inverse of my earlier commit in #453 (["Use a mask to filter log records by level"](https://github.com/go-kit/kit/pull/453/commits/134206ee8fc5b2dec23f2224d206f4b5c4cc6d07)). I started with the inequality that you propose here, but found that the bitmask implementation beat it in my benchmarks. Unfortunately, I didn't include the improvement in the commit message, but I recall it being on the order of 3%. It's worth testing again.

peterbourgon: Hmm. My philosophy tends to be that any code change that deviates from a naive, intuitive implementation needs a degree of justification proportionate to that deviation. There is a lot of subjectivity built in to that criteria, unfortunately. To my subjective eyes, bitmasking levels in this way is enough of a headscratcher that I'd want to see a much clearer win than a 3% performance improvement. With that objection raised, I defer to Chris' final call.

seh: I agree that thinking of the levels as being ordered significantly and comparing via inequality is more intuitive. My goal in #453 was to "build a race car," so for each operation I was trying to find the fastest way to achieve the same functional outcome.

log/experimental_level/level_test.go
```diff
@@ -148,10 +149,41 @@ func TestContextLevel(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+
+			logger := tc.format(&buf)
+			level.Info(logger).Log("foo", "bar")
+			if want, have := tc.output, strings.TrimSpace(buf.String()); want != have {
+				t.Errorf("\nwant: '%s'\nhave '%s'", want, have)
```
ChrisHines: `%q` works the same on all platforms as it is geared towards escaping strings like you would see in Go source code. It is often unsatisfactory when printing Windows file paths because the `\`s get doubled up, but that's not my motivation here. In this case the motivation is that I added a test that outputs JSON and didn't want to see lots of `\"` in the output.

IMO, `%q` should be used when the intent is for escaped strings. If you just want to demarcate the beginning and ending of a string (to spot extra white space or spot empty strings) then explicit markers are preferred.

Of course now I am thinking that perhaps `` ` `` would be better than `'` for Go because of its use within the language for unescaped string literals.

log/experimental_level/level.go
ChrisHines: I would love to see the exported function names get less verbose. In particular I think the current _AndAbove_ terminology is not always intuitive because I often think of level error as _more critical_ than level info and therefore above it in some way. I am not sure that everyone has the same intuition about which levels are the "high" levels and which are the "low" levels. I've seen both philosophies implemented in various logging libraries.

I do think it is generally accepted that level info is a superset of level warn which is a superset of level error from an output perspective. I think that intuition is what our exported API should lean on.

I am not too concerned about the implementation details. I am fine either way. If we can measure a consistent performance difference I would be inclined to go with the faster implementation along with an appropriate level of justification in internal comments.

seh: I'm relieved to see another acknowledgment that the order of these common levels is inconsistent in different libraries. I recall disagreeing with [_log15_'s ordering](https://github.com/inconshreveable/log15/blob/master/logger.go#L18-L22), which puts "critical" at the bottom/lowest and "debug" at the top/highest. "Severity" is an ordering that people will understand, but using function names like "InfoAndMoreSevere" didn't ring well to my ear.

peterbourgon: Great, so shall we leave the bitmasking as-is, and rename the options?

ChrisHines: Yes, let's focus on the exported names and API for now. I think the option names can be improved.

peterbourgon:
While working on the proposal, it struck me that a great feature for NewFilter would be an option to promote unleveled log records to leveled, at a chosen level, on an opt-in basis. What do you think?

ChrisHines:
>a great feature for NewFilter would be an option to promote unleveled log records to leveled

That might be useful to some people, but I think it should be implemented as a new `Logger` rather than an additional option on `NewFilter`. It is a different kind of functionality and orthogonal to what `NewFilter` does. `NewFilter` controls whether a log event is removed from the stream. This new idea is about modifying a log event by injecting a level keyval that doesn't exist.

#### Outline of one approach

My first idea was to export the level values and provide the following function.
```go
func NewInjector(logger log.Logger, level *levelValue) log.Logger
```
Which actually works even though `levelValue` is not exported, but it looks really weird. After futzing around a bit more I arrived at the following structure.

```go
func NewInjector(next log.Logger, level Value) log.Logger

type Value interface {
	String() string
	levelVal()
}

func Key() interface{}  { return key }
func ErrorValue() Value { return errorValue }
func WarnValue() Value  { return warnValue }
func InfoValue() Value  { return infoValue }
func DebugValue() Value { return debugValue }

var (
	key        interface{} = "level"
	errorValue             = &levelValue{level: levelError, name: "error"}
	warnValue              = &levelValue{level: levelWarn, name: "warn"}
	infoValue              = &levelValue{level: levelInfo, name: "info"}
	debugValue             = &levelValue{level: levelDebug, name: "debug"}
)

type levelValue struct {
	name string
	level
}

func (v *levelValue) String() string { return v.name }
func (v *levelValue) levelVal()      {}
```

Providing the `Key` and `[Level]Value` functions gives other packages access to the special values in a read only manner. This is valuable to users that want to write their own level filtering or injection logic. It gives an application that needs to modify the levels emitted by some library according to some specific rule the tools it needs to look for and assign level keyvals.

An application would use `NewInjector` as follows.
```go
var logger log.Logger
logger = log.NewLogfmtLogger(os.Stdout)
logger = level.NewInjector(logger, level.InfoValue())
```

I have this coded up with a unit test but no docs yet. Let me know if this looks worth adding and I'll finish it up and add it to this PR.

peterbourgon:
Brilliant. Looks great to me. The `levelVal` method is just to distinguish the interface from e.g. `fmt.Stringer`, is that correct? And why are `func Key()` and `type key` of type `interface{}` versus e.g. `string`?

seh:
I'm leery of the utility in exposing the level values like this, even behind these functions.

Though it sounds like you did consider an alternative like this, I'll describe it just to solicit your reaction. What if instead of exporting these six new symbols (a type and five functions), you instead exported four functions like this:
- `func WithDebugAsDefault(log.Logger) log.Logger`
- `func WithInfoAsDefault(log.Logger) log.Logger`
- `func WithWarnAsDefault(log.Logger) log.Logger`
- `func WithErrorAsDefault(log.Logger) log.Logger`

Each fixes `NewInjector`'s second parameter into the name of the function, and we avoid exposing the level values to callers.

We use the word "default" all the time in programming, but here it feels especially apt: for lack of a specified alternative, we use this value.

peterbourgon:
Why are you leery of exposing level values? I think it's great: it means users may implement their own identical decorators, or even fall back to a more explicit behavior:

```go
logger.Log(level.Key, level.DebugValue, "msg", "hello")
```

seh:
Doing so allows use of the level value in a key position, which one could emulate well enough with a bare string key, but it seems like the filtering `Logger` might have to contend with more weird situations pertaining to these values when the package doesn't have complete control over their insertion.

It's no worse than it was with strings as the keys and values, but with the private type we have the opportunity to constrain its use. It sounds like you see that more as an opportunity than a liability. I don't have any real examples of potential misuse here. That's why I said "leery," as opposed to "vehemently opposed." [The threat I described](https://github.com/go-kit/kit/pull/453#issuecomment-281223955) over in #453 is more likely to arise.

ChrisHines:
I am not sympathetic to the concerns about misuse. This package provides convenient helpers that ensure the usage is correct. Code that chooses to use the individual values must also assume the responsibility of doing so correctly.

There is little we can do about code using string values that match the results of the `String` method on the levelValues. But it does raise an interesting question about what _defines_ a leveled log record, the typed data or the final `[]byte` representation.

I am also not sure how I feel about the fact that this PR will require packages to use its types in order to be compatible with its filter. That is not the case with the existing string based approach. Having said that, exporting the `Key` and `*Value` functions opens the door for people to write adapters for other logging packages (e.g. logrus, log15) or packages that use our `log.Logger` interface but some other means of adding level keyvals. The adapters can inspect the incoming data and convert it to this package's representation.

ChrisHines:
>And why are func Key() and type key of type interface{} versus e.g. string?

That is where one of the allocations is saved. Preallocating those as interface values avoids an allocation per `Log` call. If they were strings then the compiler would add a heap allocation each time it gets converted to an interface{}. Passing a variable that is already an interface type, or some pointer type, to an interface{} parameter only copies the interface header, and doesn't need to copy the wrapped value and so doesn't need to allocate space for it.

Granted it does make the return type of `Key` look odd.

seh:
> Granted it does make the return type of Key look odd.

You could convert it back and return a string from `Key`, though. That is, use the `interface{}` variables internally, but return a string to callers who want the key. Though that yields a less peculiar interface, it then forces each caller to pay a subtle price until they discover the optimization that we are using here internally.

peterbourgon:
> Preallocating those as interface values avoids an allocation per Log call.

Perfect. Perhaps a comment to this effect?

ChrisHines:
Ready for review.

/cc @peterbourgon @seh

log/experimental_level/level.go
```diff
@@ -1,156 +1,204 @@
...
+// squelched with no error.
 func ErrNoLevel(err error) Option {
 	return func(l *logger) { l.errNoLevel = err }
 }
 
-// Option sets a parameter for the leveled logger.
-type Option func(*logger)
+// NewInjector wraps next and returns a logger that adds a Key/level pair to
+// the beginning of log events that don't already contain a level.
+func NewInjector(next log.Logger, level Value) log.Logger {
```
peterbourgon: "In effect, this gives a default level to logs without a level key."

peterbourgon:
With that small addition to the comment on NewInjector, :shipit:

PR #473, opened 2017-02-28
Fix: rename receiver name

kisPocok:
Receiver name `bs` should be consistent with previous receiver name `s` for Service.


peterbourgon:
Thanks!

kisPocok:
It's my pleasure. Thanks for your great examples!

PR #479, opened 2017-03-03
Support for gRPC Server Response Headers and Trailers, adding ClientAfter handler

### LINKED ISSUE #477: Allow multiple calls of http.ClientBefore() and http.ClientAfter() (for server as well)
Right now we are limited to one last call to set Before and After handlers.
It would be good to have ability to call them multiple times.

Changes are small and won't affect anyone logic.

basvanbeek:
This is a (minor) breaking change to the existing gRPC transport but imho is a better and more Go kit idiomatic approach to handling server response metadata.

Stuff changed:
- no more Server wide `context` as it purely uses the per request context as provided by gRPC
-changed ServerResponseFunc to provide you with empty header and trailer metadata pointers so you can fill them with chained middlewares
- new ClientResponseFunc to read the received header and trailer metadata
- before and after functions in the client transport can now be added multiple times (fixes #477 too)


examples/addsvc/cmd/addsvc/main.go
```diff
@@ -222,7 +222,7 @@ func main() {
 			return
 		}
 
-		srv := addsvc.MakeGRPCServer(ctx, endpoints, tracer, logger)
+		srv := addsvc.MakeGRPCServer(endpoints, tracer, logger)
```
peterbourgon: +1

peterbourgon:
Yep, this all scans perfectly, LGTM. In the merge commit, would you mind adding a small sentence or two about the changes required for consumers?

basvanbeek:
will do

PR #480, opened 2017-03-05
Use per request context as provided by net/http. (deprecates server wide context)

basvanbeek:
Since go 1.7 we have per request context available. This eliminates the need for server wide scoped context. Since this is also our now lowest supported version we can move away from the old implementation.

This PR removes using the server wide context inside the ServeHTTP handler in favor of the one provided by *http.Request.

Examples and tests have been updated to reflect the change too.

### DIFF
--- a/examples/addsvc/cmd/addsvc/main.go
+++ b/examples/addsvc/cmd/addsvc/main.go
@@ -207,7 +207,7 @@ func main() {
 	// HTTP transport.
 	go func() {
 		logger := log.NewContext(logger).With("transport", "HTTP")
-		h := addsvc.MakeHTTPHandler(ctx, endpoints, tracer, logger)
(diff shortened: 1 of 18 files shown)


yurishkuro:
+1

I am not even sure there was a static context in the api, since context is a request-scoped concept.

peterbourgon:
LGTM! Merge at your leisure.

basvanbeek:
See: https://golang.org/pkg/context/

`Package context defines the Context type, which carries deadlines, cancelation signals, and other request-scoped values across API boundaries and between processes.`

With both HTTP and gRPC transports providing a per request generated Context server side we should really use those.

kwiesmueller:
When testing this with the JWT Middleware there are issues as it seams the context is no longer populated with the passed in Tokens. Any updates on the related docs or is there a need to change the jwt part too?

PR #482, opened 2017-03-06
Allow IsTerminal passthrough for SyncLogger

ereOn:
The current implementation of `IsTerminal` does not properly work when stdout or stderr are used through a `SyncLogger`.

This PR adresses that by defining and implementing a new `interface` for the `SyncWriter` that exposes the encapsulated `io.Writer`. This new interface is then used inside `term.go` which now performs the `IsTerminal` check on the deepest `io.Writer` instead.

This ensures transivity of the `IsTerminal` attribute.

There is no API break, tests are passing and a new test was added to cover the new interface.

### DIFF
--- a/log/term/term.go
+++ b/log/term/term.go
@@ -7,11 +7,23 @@ import (
 	"github.com/go-kit/kit/log"
 )
 
+type parentWriter interface {
+	ChildWriter() io.Writer
+}
+
+func resolveWriter(w io.Writer) io.Writer {
+	if w, ok := w.(parentWriter); ok {
+		return resolveWriter(w.ChildWriter())
+	}
+
+	return w
+}
+
 // NewLogger returns a Logger that takes advantage of terminal features if
 // possible. Log events are formatted by the Logger returned by newLogger. If
 // w is a terminal each log event is colored according to the color function.
 func NewLogger(w io.Writer, newLogger func(io.Writer) log.Logger, color func(keyvals ...interface{}) FgBgColor) log.Logger {
-	if !IsTerminal(w) {
+	if !IsTerminal(resolveWriter(w)) {
 		return newLogger(w)
 	}
 	return NewColorLogger(NewColorWriter(w), newLogger, color)
--- a/log/sync.go
+++ b/log/sync.go
@@ -42,6 +42,11 @@ type SyncWriter struct {
 	w  io.Writer
 }
 
+// ChildWriter returns the associated child io.Writer.
+func (w *SyncWriter) ChildWriter() io.Writer {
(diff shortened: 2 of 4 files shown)


log/term/term.go
```diff
@@ -7,11 +7,23 @@ import (
 	"github.com/go-kit/kit/log"
 )
 
+type parentWriter interface {
+	ChildWriter() io.Writer
+}
+
+func resolveWriter(w io.Writer) io.Writer {
```
ereOn: Note: This method is a pass-through for all  `io.Writer` instances that do not implement the `parentWriter` interfacem, thus **not** breaking the existing behavior.

ChrisHines:
Sorry it has taken me a few days to get back to this.

The solution here seems straightforward at first glance. However ... it points out the more general issue that passing any decorated `io.Writer` to `IsTerminal` causes it to return false. For example `bufio.NewWriter(os.Stdout)` also tricks `IsTerminal`. So even if we merge this PR there is still a problem with `IsTerminal(log.NewSyncWriter(bufio.NewWriter(os.Stdout)))`. I'm not sure how I feel about this being only a partial fix.

Aside from the above I'm not crazy about the method name ChildWriter. I like WrappedWriter slightly better, but I'm not sold on that either. Suggestions anyone?

ereOn:
@ChrisHines No problem about the delay !

I agree that the proposed solution does not account for all cases but I would still distinguish external types and types provided by go-kit.

For the first ones, it's harder (impossible ?) to make any guarantee on their exposed interface. For instance, in your example there is just no way to get the underlying `io.Writer` even if we detect the correct instance type: `bufio.Writer` just doesn't offer that possibility.

For the go-kit provided types, I'd say it's another deal because we do control exactly how they behave and what they expose.

I agree that added the support for any `Writer` is a nice ultimate goal, but this scoped change still has some consistency in its own I believe.

As for the name, I'm not too fond of `ChildWriter` either to be honest. What do you think of:

- `WriterWrapper` - It's after all a interface for all `Writer` instances that wrap another one.
- `CompositeWrapper`
- `NestedWrapper`

ereOn:
@ChrisHines ping ?

ChrisHines:
Ping received, server overloaded at the moment.  ;)

I haven't forgotten about this, just too busy this week.

ereOn:
@ChrisHines No worries. I just wanted to make sure my PR had not fallen into a crack. :D

I understand being busy, trust me ;)

ereOn:
Closing in favor or #501 which addresses the problem in a less intrusive manner.

Issue #484, opened 2017-03-06
IsTerminal does not properly detect MSYS2 terminals on Windows.


ereOn:
The current implementation of `IsTerminal` for Windows does only detect when a Go program is running in `cmd.exe` but not when it is running in `MSYS2` (that includes Git bash for Windows).

Adding support for such a terminal detection can easily be done ([see this C implementation](https://fossies.org/linux/vim/src/iscygpty.c)) and I am willing to do a PR for this if desired.

The change would be 100%-backward compatible and self-contained.

What do you think ?

ereOn:
I added a PR (#485) that allows MSYS/Cygwin terminal detection, using the same strategy followed by Vim (the text editor).

Let me know what you think.

ereOn:
@peterbourgon ping ?

peterbourgon:
I have no Windows exposure or experience. @ChrisHines?

ChrisHines:
I do have Windows here. I've been too busy the last couple of weeks to put time into this. I haven't forgotten and will get to it when I have time. Also, I want to sort out #482 first.

ChrisHines:
Fixed by #501.

ereOn:
@ChrisHines Isn't there a mistake ? I don't think this was addresses by #501 which only deals with passing through the file descriptor when a `SyncWriter` is used.

ChrisHines:
Yeah, my mistake.

ChrisHines:
Implemented in #485.

PR #485, opened 2017-03-06
Added msys2/cygwin terminal detection support

ereOn:
This PR adds support for MSYS/Cygwin terminal support detection on Windows (#484).

The strategy for the regular terminal is still to use `GetConsoleMode` to detect the classic `cmd.exe` console.

If that fails, instead of failing directly, we now fallback to the following sequence:
1. If the `io.Writer`'s handle is a *pipe* handle, we move to 2.
2. If the pipe's name matches `\\(cygwin|msys)-\w+-pty\d?-(to|from)-master`, MSYS/Cygwin terminal is detected and we return `true`.

If any of the two points above fails, we return `false` like the previous implementation.

The exact [same check](https://fossies.org/linux/vim/src/iscygpty.c) is done in Vim, the popular UNIX text-editor.

I also added a test that validates (on Windows only) that STDOUT is detected as a terminal.

### DIFF
--- a/log/term/terminal_windows.go
+++ b/log/term/terminal_windows.go
@@ -8,7 +8,9 @@
 package term
 
 import (
+	"encoding/binary"
 	"io"
+	"regexp"
 	"syscall"
 	"unsafe"
 )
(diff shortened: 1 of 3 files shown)


log/term/terminal_windows_test.go
```diff
@@ -0,0 +1,37 @@
...
+func (w *myWriter) Write(p []byte) (int, error) {
+	return 0, fmt.Errorf("not implemented")
+}
+
+func (w *myWriter) Fd() uintptr {
+	return w.fd
+}
+
+func TestIsTerminal(t *testing.T) {
+	handle, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
```
ereOn: The test directly fetches STDOUT using `GetStdHandle` because `go test` redirects its output (and so fails) but **only** if not run with `-v` which makes for a surprising behavior otherwise !

log/term/terminal_windows.go
```diff
@@ -8,15 +8,23 @@
...
 
 import (
+	"bytes"
+	"encoding/binary"
 	"io"
+	"regexp"
 	"syscall"
 	"unsafe"
+
+	"golang.org/x/sys/windows"
```
ChrisHines: I would rather avoid this enormous dependency since we're only using a few of its functions and they are not difficult to implement locally.

log/term/terminal_windows_test.go
ChrisHines: Please add this as a comment in the code.

log/term/terminal_windows.go
ereOn: Doesn't this come with Go itself ?

I mainly used that instead of reinventing the wheel but I guess I can reimplement those in there if you think it is better form.

ChrisHines:
I have verified that this PR fixes `IsTerminal` for git bash on Windows. I discovered, however, that `NewColorWriter` does not work correctly for a git bash terminal anymore. It turns out that git bash handles ANSI color codes natively, and all of the syscalls in `colorwriter_windows.go` that are enabled when `IsTerminal` works properly for git bash break NewColorWriter. So we should address the logic in `NewColorWriter` to make sure it works properly on MSYS2/Cygwin in combination with the `IsTerminal` fix here.

log/term/terminal_windows.go
ChrisHines: No, it requires using `go get` to retrieve and would need to be vendored by those that vendor their dependencies.

ereOn: Done. I removed all the reference to external modules. I discovered that `syscall` actually already has a lot of useful stuff we needed, including for handling UTF16 names (on which Windows heavily relies).

log/term/terminal_windows_test.go
ereOn: Done.

ereOn:
@ChrisHines I changed everything you mentionned.

I noticed a weird behavior of `go test`: whenever it is launched without `-v` or ran indirectly (like `go test ./...`), `os.Stdout` is **not** a terminal. I had to explicitely open `CONOUT$` (which directly points at the console output) to write a realiable test.

All the code now uses only `syscall` which avoid a useless dependency (thanks for the advice on this).

Let me know what you think of this.

ereOn:
Updated the PR to restore the behavior.

I manually tested it with: https://gist.github.com/ereOn/32aafdf41c34b9749a90f464b2c96498.

image
image

log/term/terminal_windows.go
```diff
@@ -8,24 +8,94 @@
...
+	} else {
+		// The writer has no file-descriptor and so can't be a terminal.
 		return false
 	}
+
 	var st uint32
-	r, _, e := syscall.Syscall(procGetConsoleMode.Addr(), 2, fw.Fd(), uintptr(unsafe.Pointer(&st)), 0)
-	return r != 0 && e == 0
+
+	// If the terminal is a Windows console, succeed right away.
```
ereOn: Indeed. I had to push in a hurry, forgot about those comments. Fixed.

log/term/terminal_windows.go
```diff
@@ -8,24 +8,94 @@
...
+	err := syscall.GetConsoleMode(handle, &st)
+
+	return (err == nil && st != 0)
```
ereOn: Done.

log/term/terminal_windows.go
```diff
@@ -8,24 +8,94 @@
...
+	var handle syscall.Handle
+
+	if fw, ok := w.(fder); ok {
+		handle = syscall.Handle(fw.Fd())
+	} else {
+		// The writer has no file-descriptor and so can't be a terminal.
+		return false
+	}
+
+	// The terminal is not a cmd.exe terminal, let's try to detect MSYS2 terminals.
```
ereOn: Yep indeed. That was a leftover too. Removed.

ereOn:
@ChrisHines I addressed all the remaining comment issues. Thanks for the feedback !

ChrisHines:
Thanks for the contribution. Good work!

Issue #333, opened 2016-08-03
log: support syslog formatter


fabxc:
The option of a syslog formatter would add a critical feature for Prometheus to consider switching to the go-kit log package.

We are currently using our own structured logging wrapping logrus and added our syslog formatter here: https://github.com/prometheus/common/blob/master/log/syslog_formatter.go

ChrisHines:
I think it would be fantastic if the Go kit log package could be adopted by Prometheus. Can you help with an implementation? I don't have any experience with syslog or how Prometheus uses it. I can certainly help think through how to integrate it with Go kit log and review or polish the code once we figure out how it should work.

I've taken a look at the code you linked. It seems to me that the biggest challenge is how to surface the log levels to the logging client and preserve them through to the eventual call to one of the syslog.Writer methods.

fabxc:
To be honest, neither have I. This was a third party contribution to Prometheus's log package.
Log levels seem like a challenge indeed.

If I can find time I'd love to dig into it a bit but I cannot give any guarantees at the moment.

ChrisHines:
Looking more closely, I see that Prometheus already defines a [common/log.Logger interface](https://godoc.org/github.com/prometheus/common/log#Logger) that surfaces the log levels to the rest of the code base. That answers the question of how to surface the log levels.

Since that is taken care of, I can imagine an implementation that would work within Prometheus. I am not sure how much of that code would be general purpose enough to publish within Go kit itself. It is probably easier to build the specific Prometheus solution first and see what we learn from that.

I could help with a PR for https://github.com/prometheus/common that replaced the current implementation of common/log.Logger with a Go kit log based implementation (including a new syslog adapter) if that is desired.

fabxc:
My idea (haven't asked other maintainers extensively yet) was more about replacing our usage of `github.com/prometheus/common/log` with the go-kit log package through our entire code base.

In the end we don't really want to maintain our own log package if it's in the end has the same goals as an existing one.

The only thing the gokit logger does not have, which was a requirement back then, is colored output.

ChrisHines:
We have a colored Logger in the kit/log/term package. See if it meets your
needs.

fabxc:
Ah, my oversight, thanks!

ChrisHines:
@fabxc:

>My idea (haven't asked other maintainers extensively yet) was more about replacing our usage of github.com/prometheus/common/log with the go-kit log package through our entire code base.

[...]

PR #486, opened 2017-03-07
Added basic syslog support

ereOn:
This PR is an attempt to add basic syslog support as requested in #333.

It does so by relying on the very recent `log/level` package.

Example usage:

    w, _ := syslog.New(syslog.LOG_WARNING, "foo")
    defer w.Close()
    logger := level.NewSyslogLogger(w, log.NewLogfmtLogger)
    level.Error(logger).Log("name", "jake")


peterbourgon:
Syslog is defined by a particular [format on the wire](https://tools.ietf.org/html/rfc5424#section-6), which this PR doesn't appear to implement. Perhaps you've forgotten to push some files?

peterbourgon:
Thanks, interesting approach. I like the mapping of the 4 levels to the the corresponding syslog priorities. I don't like the fact that it's defined in package level, nor that it uses a factory parameter to manufacture new loggers. 

I guess I'd want to see the constructor to a SyslogLogger take a `*syslog.Writer`, which has a default priority for log messages built-in already. Then, I'd want to see the Log method introspect on keyvals to pick which Alert/Crit/Debug/Info/... method to ultimately call on the syslog.Writer, defaulting to Write if no matching level key is found. I'd want users to be able to declare and define that mapping, and I'd want the mapping you've hard-coded (the 4 levels in package level) to be a sensible default that users can opt-in to.

Does all of that make sense?

ereOn:
@peterbourgon It does make a lot of sense. Thanks for the feedback.

I'd actually like to propose another option first where there no explicit syslog logger, but rather a new kind of `io.Writer` that is able to write to syslog. I'll show you what I mean and you can tell me what you think about it. If it's not good, I'll implement it the way you just suggested.

Is that okay ?

ereOn:
@peterbourgon Here is my updated PR. It probably won't compile because I only have access to Windows at the moment (and `syslog` is not available on Windows) but it should give you a gist of the idea:

- I added the concept of `SpecializedWriter` which is an interface that, provided a `keyvals ...interface{}` returns a specialized `io.Writer` instance.
- I added a `specializedWriter` method that makes a resolution on a `io.Writer` or falls back to returning the `io.Writer` itself if it doesn't implement `SpecializedWriter`, thus ensuring backward-compatibility.
- All the existing loggers (LogFmt and JSON explicitely) use `specializedWriter`.
- I added a new `SyslogWriter` that implements `SpecializedWriter` and by default, reads the `level` field to determine which syslog priority to use.
- The syslog priority resolution can be overriden by users with an option in the `SyslogWriter` constructor.

log/syslog.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (a *syslogWriterAdapter) Write(b []byte) (int, error) {
+	return len(b), a.f(string(b))
+}
+
+func (w syslogWriter) GetSpecializedWriter(keyvals ...interface{}) io.Writer {
+	priority := w.selector(keyvals...)
+
+	switch priority {
```
ereOn: A complete implementation will of course handle all supported `syslog.Priority` values.

peterbourgon:
> All the existing loggers (LogFmt and JSON explicitely) use specializedWriter.

Yeah, this isn't gonna fly, sorry. It's not reasonable to have every other logger implementation wrap its underlying writer at serialization time with a decorator — neither from a design perspective nor (I'm guessing) a performance perspective.

As far as I can see, the design challenges here boil down to two competing requirements. First, the Syslog thing needs to introspect on the keyvals, in order to determine a priority, which means it most naturally needs to implement the Logger interface. But second, the Syslog thing is ultimately going to be a destination for other logger implementations, which means it most naturally needs to implement io.Writer. I'm not sure how to square this circle, except to say the current approach isn't the right one. 

Speculating: what about a type that implements both Logger and io.Writer, and can be wired in as both a decorator and a destination, shuttling state (i.e. priority) from the decorator part to the writer part with each invocation? e.g.

```
sw := log.NewSyslogWriter(...)

var logger log.Logger
logger = log.NewLogfmtLogger(sw)
logger = log.With(logger, "ts", log.DefaultTimestampUTC)
logger = sw.Decorate(logger) // capture level somehow
```

ChrisHines:
_Disclaimer: I haven't looked at the code in this PR yet, so forgive me if this is off base._

>I think it makes sense to do it that way because from my understanding, loggers implement how the logs are formatted and writers define where/how they are written. Syslog does not impose any format in the log messages themselves, so this implementation allows users to use whatever log formatter they want (LogFmt or JSON for instance).

That sounds like the same problem that we had with [term.NewColorLogger](https://godoc.org/github.com/go-kit/kit/log/term#NewColorLogger). Take a look at its approach to composing a `func(io.Writer) log.Logger` and `io.Writer` to support arbitrary log formats with other functionality.

ereOn:
@peterbourgon & @ChrisHines Okay. I'll look at that and I see how I can follow the same design.

Looking back at the derivative `io.Writer` classes, I'm not too fond either. It just isn't "simple" enough.

I'll keep you posted.

Mingan:
Any progress on the code, @ereOn, or should I take crack at it?

ereOn:
@Mingan You can take over if you like. I sadly am swamped in many other things to do these days.

peterbourgon:
Closing in favor of #574

PR #487, opened 2017-03-08
Add helper script for building all example binaries

yurishkuro:
This is especially useful when refactoring APIs, to ensure all examples still compile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,11 @@
 examples/addsvc/addsvc
 examples/addsvc/client/client
+examples/addsvc/cmd/addcli/addcli
+examples/addsvc/cmd/addsvc/addsvc
 examples/apigateway/apigateway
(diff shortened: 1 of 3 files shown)


ChrisHines:
It's not clear to me how this is better than `go build github.com/go-kit/kit/examples/...`, or from the kit directory `go build ./examples/...`

yurishkuro:
For compiling only, agreed. But `go build ./examples/...` doesn't actually create any executables.

ChrisHines:
To build binaries we can use `go install ./examples/...`.

yurishkuro:
That would install them to GOROOT? Seems extreme.

ChrisHines:
No, it installs them in GOPATH/bin, not GOROOT.

Also you can use `go clean -i ./examples/...` to delete them.

PR #493, opened 2017-03-08
Remove log.NewContext mention from log/level docs

narqo:


ChrisHines:
Good catch. I thought I found them all.

Issue #489, opened 2017-03-08
CloudWatch metrics package


cam-stitt:
It would be nice to have a metrics implementation for CloudWatch. Ideally, we could utilise [aws-sdk-go](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/). We should be able to just use `PutMetricData` as it creates the metric if it doesn't yet exist. I need to spend some more time looking into how to map `Counter`, `Gauge` and `Histogram` into CloudWatch metrics data types.

peterbourgon:
Oh hell yeah, I'm all about this. +1000.

cam-stitt:
I made a start to this. Happy to open a WIP pr to get some feedback on the direction that I am taking. You can see the direction I'm taking here: https://github.com/cam-stitt/kit/blob/cloudwatch-metrics/metrics/cloudwatch/cloudwatch.go

basvanbeek:
closed by #495

PR #495, opened 2017-03-08
Cloudwatch metrics

cam-stitt:
Ref #489.

I'm reasonably happy with the current implementation, but I am having some trouble working out the best way to test the writing of metrics. Going to spend some time investigating ways to run integration tests for this.

CloudWatch charges per API request so we batch the requests for all metrics at once.

### DIFF
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -1,15 +1,14 @@
 package cloudwatch
 
 import (
-	"sync"
-
-	"time"
-
 	"fmt"
+	"sync"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
 	"github.com/aws/aws-sdk-go/service/cloudwatch/cloudwatchiface"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/generic"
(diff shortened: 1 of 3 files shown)


peterbourgon:
I imagine the best way to test it would be to make a mock cloudwatchiface.CloudWatchAPI in the test, have it accept the writes, and compare output.

xla:
An approach could be to have a slimmed down interface locally which also makes it explicit what part of the Cloudwatch API the code depends on.  Currently only `PutMetricData` is used, which would reduce the surface to mock significantly.

cam-stitt:
@peterbourgon I did in fact use the interface earlier today. It's all there.

cam-stitt:
I'm happy with where this PR is at. Would love a review.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+	"time"
+
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/aws/aws-sdk-go/service/cloudwatch/cloudwatchiface"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/generic"
```
peterbourgon: Import grouping here.

```go
import (
    // stdlib

    // third-party

    // go-kit/kit/...
)
```

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+	counters   map[string]*Counter
+	gauges     map[string]*Gauge
+	histograms map[string]*Histogram
+	logger     log.Logger
+}
+
+// New returns a CloudWatch object that may be used to create metrics. Namespace is
+// applied to all created metrics and maps to the CloudWatch namespace.
+// Callers must ensure that regular calls to Send are performed, either manually or with one of the helper methods.
+func New(namespace string, logger log.Logger, svc cloudwatchiface.CloudWatchAPI) *CloudWatch {
```
peterbourgon: The logger is typically the last parameter in the constructor.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+	return err
+}
+
+// Counter is a CloudWatch counter metric.
+type Counter struct {
+	c *generic.Counter
+}
+
+// NewCounter returns a new usable counter metric.
+func NewCounter(name string) *Counter {
```
peterbourgon: It seems the only "correct" way to construct a new {Counter, Gauge, Histogram} is via the corresponding method on the CloudWatch struct. Am I reading this correctly? If so, should these free functions be eliminated?

metrics/cloudwatch/cloudwatch_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"sync"
+
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/aws/aws-sdk-go/service/cloudwatch/cloudwatchiface"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics/teststat"
+)
```
peterbourgon: Import grouping, as above.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+// usage, create a time.Ticker and pass its C channel to this method.
+func (cw *CloudWatch) WriteLoop(c <-chan time.Time) {
+	for range c {
+		if err := cw.Send(); err != nil {
+			cw.logger.Log("during", "Send", "err", err)
+		}
+	}
+}
+
+// Send will fire an api request to CloudWatch with the latest stats for
```
peterbourgon: API

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+// all metrics.
```
peterbourgon: Perhaps a note that most users will want to call WriteLoop, instead? Or — better yet — maybe you can unexport this method?

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+	cw.mtx.Lock()
+	cw.histograms[name] = h
+	cw.mtx.Unlock()
+	return h
+}
+
+// WriteLoop is a helper method that invokes Send every
+// time the passed channel fires. This method blocks until the channel is
+// closed, so clients probably want to run it in its own goroutine. For typical
+// usage, create a time.Ticker and pass its C channel to this method.
```
peterbourgon: Comment wrapping.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+
+// NewCounter returns a counter. Observations are aggregated and emitted once
+// per write invocation.
+func (cw *CloudWatch) NewCounter(name string) *Counter {
+	c := NewCounter(name)
+	cw.mtx.Lock()
+	cw.counters[name] = c
+	cw.mtx.Unlock()
+	return c
+}
```
peterbourgon:
```go
cw.mtx.Lock()
defer cw.mtx.Unlock()
c := &counter{...} // assuming you remove the free constructor
cw.counters[name] = c
return c
```

metrics/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,226 @@
...
+
+// Send will fire an api request to CloudWatch with the latest stats for
+// all metrics.
+func (cw *CloudWatch) Send() error {
+	cw.mtx.RLock()
+	defer cw.mtx.RUnlock()
+	now := time.Now()
+
+	datums := []*cloudwatch.MetricDatum{}
```
peterbourgon: I _slightly_ prefer

```go
var datums []*cloudwatch.MetricDatum
```

but this is a tiny nit.

peterbourgon:
Oh! And can you also please add CloudWatch to the appropriate table/s in the documentation in the parent directory?

metrics/cloudwatch/cloudwatch.go
cam-stitt: I thought it might be worth keeping this around as someone might have their own way they want to fire the api call.

metrics/cloudwatch/cloudwatch.go
cam-stitt: Sure, I followed the pattern from one of the other metric packages. I'll clean this up.

metrics/cloudwatch/cloudwatch.go
peterbourgon: Yep, totally fair, just some guidance in the comment would be good.

cam-stitt:
The only thing I haven't done is add it to the table in documentation. I don't quite understand the table and haven't yet had enough time to read into it.

peterbourgon:
Mentioned this in Slack but just for the record:

> DIM = does the system support arbitrary dimensionality (i.e. does it support With)? if yes, `n`; if not, `1`
> COUNTERS = batch push-aggregate
> GAUGES = batch push-aggregate
> HISTOGRAMS = synthetic (because they're modeled as per-quantile gauges), batch, push-aggregate

cam-stitt:
Sorted. Thanks for clearing that up.

peterbourgon:
Amazing, thank you so much!

PR #497, opened 2017-03-12
Request to move from prometheus http (deprecated) to the promhttp

### LINKED ISSUE #491: Request to move from prometheus http (deprecated) to the new one
As it stands on the `http.go` file in the `prometheus/client_golang` repository, that file has been deprecated and moved to `prom/http` directory, with some changes. Quoted from `http.go` comment section:
```
It is a partial mirror of promhttp/http.go (to avoid circular import chains) where 
everything HTTP related should live. The functions here are just for avoiding breakage. 
Everything is deprecated.
``` 
For this reason, would it be possible to update the prometheus metrics in go-kit to use the new one?

kavirajk:
fixes #491 

Changed the metrics code in the examples that uses depricated prometheus handler to promhttp handler.


peterbourgon:
Ace! Thanks!

PR #501, opened 2017-03-20
log: Update NewSyncWriter to work with term.IsTerminal.

### LINKED ISSUE #483: IsTerminal does not account for SyncLogger-encapsulated loggers
The current implementation of IsTerminal does not properly work when stdout or stderr are used through a SyncLogger.

I came up with a really simple PR that addresses the problem in an elegant, 100% backwards-compatible manner. See #482.

ChrisHines:
- Change the return type to io.Writer
- Return an io.Writer that also has the `Fd() uintptr` method of the
  wrapped io.Writer if present.

Fixes #483 as an alternate approach to #482.

What do you think of this approach @ereOn & @peterbourgon?

### DIFF
--- a/log/sync.go
+++ b/log/sync.go
@@ -1,4 +1,4 @@
-package log
+﻿package log
 
 import (
 	"io"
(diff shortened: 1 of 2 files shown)


peterbourgon:
I like the approach.

ereOn:
@ChrisHines I like it too. Good for me !

PR #507, opened 2017-03-27
Fix overriding of metadata set by the user.

kujtimiihoxha:
The metadata set by the user is being overridden by: 
https://github.com/go-kit/kit/blob/master/transport/grpc/client.go#L94-L98
I think we should be allowed to set the metadata without using `before` functions just like the official doc describes it : https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md
This fixes this issue and allows users to set the metadata.

### DIFF
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -91,8 +91,8 @@ func (c Client) Endpoint() endpoint.Endpoint {
 			return nil, err
 		}
 
-		md,ok := metadata.FromContext(ctx)
-		if !ok{
+		md, ok := metadata.FromContext(ctx)
+		if !ok {
 			md = metadata.MD{}
 		}
 		for _, f := range c.before {
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -91,7 +91,10 @@ func (c Client) Endpoint() endpoint.Endpoint {
 			return nil, err
 		}
 
-		md,_ := metadata.FromContext(ctx)
+		md,ok := metadata.FromContext(ctx)
+		if !ok{
+			md = metadata.MD{}
+		}
 		for _, f := range c.before {
 			ctx = f(ctx, &md)
 		}
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -91,11 +91,11 @@ func (c Client) Endpoint() endpoint.Endpoint {
 			return nil, err
 		}
 
-		md := &metadata.MD{}
+		md,_ := metadata.FromContext(ctx)
 		for _, f := range c.before {
-			ctx = f(ctx, md)
+			ctx = f(ctx, &md)
 		}
-		ctx = metadata.NewContext(ctx, *md)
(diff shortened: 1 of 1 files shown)


peterbourgon:
Please gofmt your code.

kujtimiihoxha:
Sorry about that, done.

peterbourgon:
No objections here. @basvanbeek does this look straightforward?

basvanbeek:
It is straightforward but I don't like the implications. It will promote / allow people to handle transport details in their business logic. By enforcing metadata headers to be composed through endpoint middleware and client before functions we keep this clean and separated. I will adjust this comment later as I have to run and won't be back until later this evening.

peterbourgon:
Interesting. I see both sides here; my intuition is to favor @kujtimiihoxha's interpretation, because I guess the whole point of context is to transport information across these layer boundaries. But I'll defer to your future edits.

kujtimiihoxha:
@peterbourgen If we want to keep the current implementation I would suggest to add some docs on how to set the metadata the `gokit` way, because all the docs I could find is with the solution I linked above.

peterbourgon:
@basvanbeek spoke to me offline and cleared a few things up. So, an important property of Go kit is _transport agnosticism_, that is, your business logic shall not be coupled to any specific transport.

One way that property manifests is with Before and After funcs in the transport packages, which are used to move information between the transport and business domains.

For example, if your service layer needs to operate on some information that comes in on an HTTP request header, you need to write a ServerBefore transport/http.RequestFunc to extract it and stick it in the context. If you want to expose the same service on gRPC, you'll write a ServerRequestFunc to do the same work, but (obviously) extract it from the e.g. grpc/metadata.MD instead.

This strict separation of transport metadata (e.g. HTTP headers, gRPC metadata) from the context.Context passed to the endpoints is deliberate. That's why we create an empty metadata.MD in the transport/grpc.Client, we want to ensure it's populated deliberately. This PR would subvert that separation, by allowing business logic or endpoint middlewares to create gRPC metadata and stick it into the context to be consumed by the transport layer.

Put simply: your business logic shouldn't have any knowledge of gRPC, and shouldn't create grpc/metadata.MD objects.

In your use case, the correct thing to do is (probably) place each of the items you want to transport over the grpc/metadata.MD into the context.Context individually, and then use ClientBefore functions to migrate them from thecontext to the MD. But with that said Ladon fits wonderfully with this philosophy, i.e. (paraphrasing @basvanbeek here) the Ladon object _is_ business logic, and is neatly isolated — it can be handled from the business layer by the user, and Go kit can handle transport concerns by providing middleware. 

@kujtimiihoxha, what do you think about writing some Ladon middlewares instead? I think you can use [package opentracing](https://godoc.org/github.com/go-kit/kit/tracing/opentracing) and its helpers as a rough guide...

kujtimiihoxha:
Thanks for the clarifying why we do override the metadata.

Issue #509, opened 2017-03-31
Runtime created middlewares


glerchundi:
First of all thanks for this incredible piece of framework/library/tooling/whatever it is!

I would like to discuss about the idea of having a dynamic middleware/endpoints, one that can do whatever middleware action based on context/request.

For example, rate limit based on the request data, JWT email/identifier, or whatever context information you already know its there:

```
func NewFinishedMiddleware(response interface{}, err error) endpoint.Middleware {
	return func(_ endpoint.Endpoint) endpoint.Endpoint {
		return func(ctx context.Context, request interface{}) (interface{}, error) {
			return response, err
		}		
	}
}

func NewRuntimeMiddleware(mwFunc func (context.Context, interface{}) endpoint.Middleware) endpoint.Middleware {
	return func(next endpoint.Endpoint) endpoint.Endpoint {
		return func(ctx context.Context, request interface{}) (interface{}, error) {
			mw := mwFunc(ctx, request)
			return mw(next)(ctx, request)
		}
	}
}
```

And then rate limit based on token bucket retrieved from redis based on jwt identifier:
```
redis := ...
mw := NewRuntimeMiddleware(func(ctx context.Context, request interface{}) endpoint.Middleware {
	name := ctx.Value(jwt.JWTTokenContextKey).Header["name"]
	tbs, err := redis.Get(fmt.Sprintf("/token-buckets/%s", name)
	if err != nil {
		return NewFinishedMiddleware(nil, err)
	}		
	var tb ...
	if err := proto.Unmarshal(tbs, &tb); err != nil {
		return NewFinishedMiddleware(nil, err)
	}
	return NewTokenBucketLimiter(tb)
})
```

Probably there are a lot of errors in this *pseudocode*, this is just for having an idea how it would work.

WDYT?

peterbourgon:
You can create these yourself, right? What support do you want to see in Go kit proper?

glerchundi:
I would like to see this topic discussed here so that a community thought
solution could benefit to everyone as I think this is a very common problem.

Probably what I have outlined before its not the best solution in terms of
composability, performance, ...

How does people usually solve this problem? For example Signal Server (by
(un)marshalling a leacky bucket in redis uses dynamic rate limiters per
user/action.

In case you think this is something not applicable to this library, be free
to close the issue.

peterbourgon:
[...]

PR #510, opened 2017-04-02
[WIP] ratelimit: allow resolving token bucket using request scoped info

glerchundi:
Not tested yet, just a concrete example for #509. I'll extend this in case it seems interesting enough. Thanks for your time @peterbourgon

### DIFF
--- a/ratelimit/token_bucket.go
+++ b/ratelimit/token_bucket.go
@@ -18,8 +18,22 @@ var ErrLimited = errors.New("rate limit exceeded")
 // limiter based on a token-bucket algorithm. Requests that would exceed the
 // maximum request rate are simply rejected with an error.
 func NewTokenBucketLimiter(tb *ratelimit.Bucket) endpoint.Middleware {
+	return NewPerRequestTokenBucketLimiter(func(_ context.Context, _ interface{}) (*ratelimit.Bucket, error) {
+		return tb, nil
+	})
+}
+
+// NewPerRequestTokenBucketLimiter returns an endpoint.Middleware that acts as a
+// rate limiter based on a token-bucket algorithm retrieved in runtime by a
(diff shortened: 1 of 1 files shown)


glerchundi:
Perhaps token/leaky bucket can be interfaced so that it would be easy to inject your own and customized behaviour, it will decouple from implementation, like:

```
type Bucket interface {
    Take(count int64) time.Duration
    TakeAvailable(count int64) int64
}
```

@peterbourgon I agree that as it's implemented right now it only covers racy use cases but if you abstract the bucket in this way, it would be possible to inject any kind of rate limit implementation, not just the juju's one.

WDYT? Maybe create two different PRs in case you think its worth it?

peterbourgon:
No, there is no need to provide a token bucket abstraction in Go kit. The Middleware abstraction is at the right level: users who want to use something other than the Juju token bucket can/should write the [9 lines of code](https://github.com/go-kit/kit/blob/master/ratelimit/token_bucket.go#L20-L29) required to implement their own Middleware.

glerchundi:
Ok, that being said I will close this PR. Rate limiting is not just Token bucket, there are plenty of implementations using different kind of tools. This is not about the amount of code, I know that it's easy enough to reimplement in client's code but I thought this lib is something, among other things, about best practices. I don't think this lib will be hurt if it abstracts to `ratelimit/middleware.go` `ratelimit/juju/token.go` which would open to any kind of implementation. It seems like there is an inconsistency between how do you decide to use one and **unique** implementation for a pattern like rate limiting and there is another completely different for the service discovery for example!

Sorry if I have bothered you and thanks for your patience.

peterbourgon:
I'd be happy to take a PR to add rate-limiting middlewares that use other, popular rate limiting packages. But I'm definitely not interested in providing abstractions for that, it's too low-level a concern for Go kit to take a stake in it.

Issue #378, opened 2016-10-17
Is it possible to have a factory abstraction for metrics?


yurishkuro:
Currently go-kit only defines interfaces for stats objects like Counter/Gauge/etc. The creation of those objects seems to be delegated to concrete implementations, with the only exception being the `Provider` interface, which only allows a single string name for stats.

Is it possible to define a generic tag-based factory, e.g. like this one:

``` go
// Factory creates new metrics
type Factory interface {
    CreateCounter(name string, tags map[string]string) Counter
    CreateTimer(name string, tags map[string]string) Timer
    CreateGauge(name string, tags map[string]string) Gauge
}
```

?

The reason for this is that our tracing library https://github.com/uber/jaeger-client-go defines a [whole set of metrics](https://github.com/uber/jaeger-client-go/blob/master/metrics.go#L31), and it would be quite cumbersome for a user of this library to construct all these metrics manually when binding to a particular metrics implementation. On the other hand, if the user simply needed to provide the above factory interface, our library would create all the metrics it needs.

I realize that not all backends support tags, but many good ones do, and those that don't (e.g. statsd) can still come up with a unique metric name given a `name+tags` combination.

peterbourgon:
The existing provider implementation is a bit of a one off, and I wouldn't like to see it enshrined as a best practice or duplicated in another form. Indeed, I hope to delete it in the long term, as it's only consumer (Gizmo) will hopefully move to a different model. 

 What do you think about turning the problem on inside, and having your framework except metric references to which a record things? That is, not creating the metrics, but treating them as dependencies, which I think is the right model anyway. Does this work? If not, can you say a bit more about your architecture?

yurishkuro:
Injecting individual metrics into the framework is currently the only way with go-kit. I would not say that it's any more valid IOC approach than injecting an [abstract factory](https://en.wikipedia.org/wiki/Abstract_factory_pattern). There are two problems with injecting individual metrics:
  1. As I mentioned above, our library uses a whole set of metrics, specifically 17 of them. That is a lot of overhead and boilerplate code that the user of the library needs to write just to wire up the library (it's an OpenTracing tracer, the wiring is meant to be just a few lines of code). With a factory approach, wiring the metrics is one line.
    * Furthermore, we're interested in using go-kit not just in the client-facing library, but in our server side components as well, and those have a lot more metrics than 17.
[...]

PR #513, opened 2017-04-14
metrics: Move the Provider interface to metrics

freeformz:
Reasons:

1. The Provider interface is all about returning things from metrics. It makes
more sense to me to have it located there.

2. provider.Provider is a form of stutter. It's more accepted then other forms
though (cf. hash.Hash), but I personally dislike it. For instance, what is a
provider.Provider providing? metrics.Provider says a lot more IMO.

3. Importing the provider package incurrs a lot of additional dependencies, most
of which aren't needed. Our work around to this is to redefine a Provider
interface locally. These are usually defined in a local `metrics` package
requiring either the local or go-kit metrics package to be renamed during import
which is :-(. These local metrics packages proliferate over time across projects.

### DIFF
--- a/metrics/provider.go
+++ b/metrics/provider.go
@@ -1,0 +1,15 @@
+package metrics
+
+// Provider abstracts over constructors and lifecycle management functions for
+// each supported metrics backend. It should only be used by those who need to
+// swap out implementations dynamically.
+//
+// This is primarily useful for intermediating frameworks, and is likely
+// unnecessary for most Go kit services. See the provider package-level doc
+// comment for more typical usage instructions.
+type Provider interface {
+	NewCounter(name string) Counter
+	NewGauge(name string) Gauge
+	NewHistogram(name string, buckets int) Histogram
+	Stop()
+}
--- a/metrics/provider/discard.go
+++ b/metrics/provider/discard.go
@@ -9,7 +9,7 @@ type discardProvider struct{}
 
 // NewDiscardProvider returns a provider that produces no-op metrics via the
 // discarding backend.
-func NewDiscardProvider() Provider { return discardProvider{} }
+func NewDiscardProvider() metrics.Provider { return discardProvider{} }
 
 // NewCounter implements Provider.
 func (discardProvider) NewCounter(string) metrics.Counter { return discard.NewCounter() }
--- a/metrics/provider/doc.go
+++ b/metrics/provider/doc.go
@@ -15,28 +15,9 @@
 //        s := statsd.New(...)
 //        t := time.NewTicker(5*time.Second)
 //        go s.SendLoop(t.C, "tcp", "statsd.local:8125")
-//        latency = s.NewHistogram(...)
 //        requests = s.NewCounter(...)
 //    default:
 //        log.Fatal("unsupported metrics backend %q", *metricsBackend)
 //    }
 //
 package provider
-
-import (
-	"github.com/go-kit/kit/metrics"
-)
-
-// Provider abstracts over constructors and lifecycle management functions for
-// each supported metrics backend. It should only be used by those who need to
-// swap out implementations dynamically.
-//
-// This is primarily useful for intermediating frameworks, and is likely
-// unnecessary for most Go kit services. See the package-level doc comment for
-// more typical usage instructions.
-type Provider interface {
-	NewCounter(name string) metrics.Counter
-	NewGauge(name string) metrics.Gauge
-	NewHistogram(name string, buckets int) metrics.Histogram
-	Stop()
-}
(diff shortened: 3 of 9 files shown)


peterbourgon:
On a first read I'm reasonably convinced. I'm gonna let it bake a bit to see if any unintended consequences shake out.

ChrisHines:
Once you pull the Provider interface up into the metrics package, would it make sense to also break out each concrete implementation of provider into a separate metrics/... package to allow clients to control dependencies better?

yurishkuro:
This is related to #378. I personally would like the provider to be a bit more versatile. 

Cf. with our [metrics.Factory](https://github.com/uber/jaeger-lib/blob/master/metrics/factory.go#L24) (s/Timer/Histogram/)
```go
// Factory creates new metrics
type Factory interface {
	Counter(name string, tags map[string]string) Counter
	Timer(name string, tags map[string]string) Timer
	Gauge(name string, tags map[string]string) Gauge

	// Namespace returns a nested metrics factory.
	Namespace(name string, tags map[string]string) Factory
}
```

metrics/provider.go
```diff
@@ -0,0 +1,15 @@
...
+// swap out implementations dynamically.
+//
+// This is primarily useful for intermediating frameworks, and is likely
+// unnecessary for most Go kit services. See the provider package-level doc
+// comment for more typical usage instructions.
+type Provider interface {
+	NewCounter(name string) Counter
+	NewGauge(name string) Gauge
+	NewHistogram(name string, buckets int) Histogram
+	Stop()
```
yurishkuro: what does stop do? looks a bit like abstraction leak, since a concrete provider is always a wrapper around a concrete backend, which probably has a way of "stopping" itself anyway.

peterbourgon:
I don't know if it's wise to blow up the scope of such a deliberate PR, but I've become convinced there are several flaws with the current metrics package/approach that could be addressed together in another refactor:

- The vividcortex package powering the generic histograms is not good enough
- I now consider a no-op With on unsupported backends an unacceptable compromise
- Provider is more important than I thought, and should be improved and receive first-class support

freeformz:
@ChrisHines Probably, but also orthogonal to this PR.

@yurishkuro : Providers basically work like factories, so I'm not sure what your are suggesting and I still need to read the entirety of #378.

@peterbourgon Does any of that impact this PR now? As a side note, what in your estimation is wrong with the vividcortex histogram?

Related to "provider is more important"...

I've seen 2 ways of using the stuff in metrics/...:
1. Declare a provider in your main(s) and pass that provider down to dependencies so that those deps can construct the metrics they want.
1. Declare a provider in main(s) and create all metrics in main(s), passing required metrics down to their dependencies.

I'm currently leaning to (1) being a poor pattern because: (a) leaves naming of the metrics up to the dependencies, which IMO is a poor choice long term; and (b) it's harder to discover the actual metrics a program (any given main) will emit.

So I am curious to about your statement wrt making provider more important.

yurishkuro:
> Providers basically work like factories

@freeformz the main difference is in our case we want to be able to (a) create metrics with specific tags, not just name, and (b) create hierarchies with sub-scopes. For example, default expvar gives you these:

image

but there's no way to create that with go-kit expvar backend, nor with the current Provider API.

freeformz:
@yurishkuro Hmmm. It's `Publish("memstats", Func(memstats))` that gives you that. You aren't publishing a single metric here and there is structure, but it happens given how the expvar package's http.Handler exports things that aren't actually expvars (it publishes the output of a function that returns a *runtime.MemStats and formats it into JSON).

yurishkuro:
I haven't tried it myself, but I assume if you define a Map, it will create a nested namespace. Don't know if it would support arbitrary nesting, but since it contains `map[string]Var` and `Map` itself is a `Var` it seems feasible.

Anyway, the point of namespace is to create a hierarchy that can even be done simply by concatenating the parent namespaces, as statsd format does.

freeformz:
Yeah, I would probably want to use/abuse the With(...) method to do that. In a system using statsd the only sane thing to do (IMO) would be to map labelValues to either `_` or `.` separators suffixed to the name. You could probably just drop the label parts also.

```go
c := statsd.NewCounter("request_duration")
c.With("method","MyMethod","status",strconv.Itoa(code)).Add(1)
```
Results in `request_duration_MyMethod_200=1`, but it doesn't work that way right now anyway.

WRT: expvars that is because expvars can export anything that matches the Var interface : https://golang.org/pkg/expvar/#Var (i.e. has a String() method that returns a JSON formatted string). We could certainly implement a similar interface and people could write adaptors for them (similar to Func : https://golang.org/pkg/expvar/#Func) but that's orthogonal to this PR (IMO anyway) and probably requires a larger re-think.

yurishkuro:
> Yeah, I would probably want to use/abuse the With(...) method to do that

for the tags part, yes, but not for namespacing. Btw, we have an adapter for expvar that respects the tags and your example would result in `request_duration.method_MyService.status_200` (i.e. not dropping the labels).

PR #518, opened 2017-04-23
Fix typo in addsvc's doc block

annismckenzie:
The server binary is available in `cmd/addsvc`, not in `cmd/addsrv`.


peterbourgon:
Thanks!

PR #522, opened 2017-05-06
fix(shipping): use `log.NewSyncWriter` in example

terinjokes:
Replace the use of `serializedLogger` with `log.NewSyncWriter` which
provides similar functionality.

`SyncWriter` is chosen over `SyncLogger` to reduce the amount of work
contained within a lock.


ChrisHines:
Nice! Thanks for the contribution.

Issue #447, opened 2017-01-25
transport/grpc.NewClient assumes service name has prefix "pb"


willfaught:
It's possible for a proto file to have something like this:

```proto
option go_package = "pb";

package github_com.mycompany.myrepo.myservice.pb;
```

in which case the Go package is pb but the GRPC namespace is github_com.mycompany.myrepo.myservice.pb.

There should at least be an option to override the pb default.

basvanbeek:
The default in Go kit is for the GRPC namespace to be pb. However it is perfectly fine to alter the namespace and do something like:

```
kitgrpc.NewClient(
    conn, 
    "github_com.mycompany.myrepo.myservice.pb.MyService",
    "MyMethod",
     ...,
)
```

willfaught:
You're right, my bad! I missed the period condition. Thanks.

terinjokes:
If you don't provide a "package" in your protobuf file, your endpoints won't pre prefixed at all. Enforcing a default of "pb" in the grpc transport that I can't opt out of means I now can't call this service.

terinjokes:
In my fork, I've added a ClientOption to revert the auto-prefixing. Happy to send a PR, but I think removing, or modifying, the prefixing functionality makes more sense.

terinjokes:
To wrap up the discussion here: with #524 merged, go-kit will no longer autoprefix the "pb" package name. Of course `github_com.mycompany.myrepo.myservice.pb.MyService` is still a valid input.

PR #523, opened 2017-05-08
fix(transport/grpc): add client option to disable prefix

terinjokes:
Add a ClientOption to revert the auto-prefixing of "pb" to service names
when there's no other prefix already specified.

Relates to #447.

---

This implementation touches the fewest lines of code, at the expense at perhaps being a bit more complex. An alternative implementation I have in mind avoids the `fmt.Sprintf` on line 45 to construct the method, stores them separately, until the method is actually needed to `Invoke`. The `ClientOption` could then just set a flag to turn off the prefixing.

What do you think?

---

Additionally, is there something I can do to denote that auto-prefixing is deprecated, so we can re-evaluate this in a future breaking change?

### DIFF
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -79,6 +79,11 @@ func ClientAfter(after ...ClientResponseFunc) ClientOption {
 	return func(c *Client) { c.after = append(c.after, after...) }
 }
 
+// ClientNoPrefix reverts the auto-prefixing of "pb" to serviceName.
+func ClientNoPrefix() ClientOption {
+	return func(c *Client) { c.method = strings.Replace(c.method, "/pb.", "/", 1) }
+}
+
 // Endpoint returns a usable endpoint that will invoke the gRPC specified by the
 // client.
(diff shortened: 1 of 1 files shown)


basvanbeek:
Hi @terinjokes thanks for this. Contrary to what we discussed earlier I'm leaning towards biting the bullet and removing the auto prefixing of `pb.` if no namespace was given. 

We postponed Go kit v1.0.0 and are still in 0.x.x so have no hard API stability guarantees yet. Changing existing gRPC clients which break due to this should be easy.

@peterbourgon what do you think?

peterbourgon:
No objections to breaking the API. I don't have sufficient context to have an opinion on the prefixing or not; I defer to your judgment there.

terinjokes:
Presumably people are vendoring and they can fix the prefixes on their own schedule.

@basvanbeek If you give me a bit to finishing waking up and starting to work, I don't mind sending a PR to remove it.

terinjokes:
@basvanbeek I opened alternative CL #524

basvanbeek:
closed in favor of #524

PR #524, opened 2017-05-09
fix(transport/grpc): remove autoprefixing logic

### LINKED ISSUE #447: transport/grpc.NewClient assumes service name has prefix "pb"

terinjokes:
Removes the autoprefixing of "pb" to service names, to allow for
grpc services created without a package prefix.

Fixes #447.

BREAKING CHANGE: Existing users of the grpc client transport who were
relying on the autoprefixing of the "pb" package name must manually
update their clients to add it. No change is required if a package name
was already being manually prefixed.

### DIFF
--- a/examples/addsvc/client/grpc/client.go
+++ b/examples/addsvc/client/grpc/client.go
@@ -34,7 +34,7 @@ func New(conn *grpc.ClientConn, tracer stdopentracing.Tracer, logger log.Logger)
 	{
 		sumEndpoint = grpctransport.NewClient(
 			conn,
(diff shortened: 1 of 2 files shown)


basvanbeek:
thanks!

PR #533, opened 2017-05-18
added client finalizer

### LINKED ISSUE #472: Logging request responses, including status codes and headers, with go-kit’s Client
Hi, I am using the kit’s Client to make GET requests. For these requests I am looking to do some structured logging around the request (URL, headers, etc) response (Body, response code/status, etc). Unfortunately there is no logger as a part of the Client (like the Server) so I am looking for ways to best achieve my requirement without having logger being passed onto 5 different functions if possible.

One of the big limitations I am running into is if the underlying Client.Do method (https://github.com/go-kit/kit/blob/master/transport/http/client.go#L100) returns an error (like timeout), the decode response and even the clientAfterResponse functions are not called, for a good reason since there is nothing to decode. But there is something missing from the time when the call fails and the error is returned to caller. If you contrast this to the Server implementation, the Server does a good job I feel that if the execute method fails, it calls logger and calls a function to do more processing (such as encoding) the error.

Question is - why does the client not have a way to call a function (as an implementation option) to do post client.Do error processing where we can even hook up logging. It will be great if the framework can provide something, even if that could be optional like a ClientOption.

travissalascox:
based on the server finazlizer. @peterbourgon super open to suggestions to make this better, also excited as it will be my first real contribution to this project if it makes it in. should resolve #472 for the most part i believe.

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -23,6 +23,7 @@ type Client struct {
 	dec            DecodeResponseFunc
 	before         []RequestFunc
 	after          []ClientResponseFunc
+	finalizer      ClientFinalizerFunc
 	bufferedStream bool
 }
 
@@ -72,6 +73,12 @@ func ClientAfter(after ...ClientResponseFunc) ClientOption {
 	return func(c *Client) { c.after = append(c.after, after...) }
 }
 
+// ClientFinalizer is executed at the end of every HTTP request.
+// By default, no finalizer is registered.
+func ClientFinalizer(f ClientFinalizerFunc) ClientOption {
+	return func(s *Client) { s.finalizer = f }
+}
+
 // BufferedStream sets whether the Response.Body is left open, allowing it
 // to be read from later. Useful for transporting a file as a buffered stream.
 func BufferedStream(buffered bool) ClientOption {
@@ -84,7 +91,21 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		ctx, cancel := context.WithCancel(ctx)
 		defer cancel()
 
-		req, err := http.NewRequest(c.method, c.tgt.String(), nil)
+		// Vars used for client finalizer to ensure there are no nil values
+		var (
+			req  *http.Request  = &http.Request{}
+			resp *http.Response = &http.Response{}
+			err  error
+		)
+		if c.finalizer != nil {
+			defer func() {
+				ctx = context.WithValue(ctx, ContextKeyResponseHeaders, resp.Header)
+				ctx = context.WithValue(ctx, ContextKeyResponseSize, resp.ContentLength)
+				c.finalizer(ctx, resp.StatusCode, req)
+			}()
+		}
+
+		req, err = http.NewRequest(c.method, c.tgt.String(), nil)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


transport/http/client.go
```diff
@@ -118,6 +140,13 @@ func (c Client) Endpoint() endpoint.Endpoint {
 	}
 }
 
+// ClientFinalizerFunc can be used to perform work at the end of a client HTTP
+// request, after the response is returned. The principal
+// intended use is for request logging. In addition to the response code
+// provided in the function signature, additional response parameters are
+// provided in the context under keys with the ContextKeyResponse prefix.
+type ClientFinalizerFunc func(ctx context.Context, code int, r *http.Request)
```
peterbourgon: Hmm. This is the same signature as in the ServerFinalizer, but I'm not sure it actually makes sense here. In the server we're guaranteed to have an http.Request, and we're guaranteed to have a code, so they make sense as parameters. But in the client, we may not make it all the way to generating an http.Request, and we may not get a status code back. And, we have an explicit error value that will always be returned via the endpoint (though it may be nil). So I think the better signature here is

```go
type ClientFinalizerFunc func(ctx context.Context, err error) 
```

What do you think?

transport/http/client.go
```diff
@@ -84,7 +91,21 @@ func (c Client) Endpoint() endpoint.Endpoint {
...
+			resp *http.Response = &http.Response{}
+			err  error
+		)
+		if c.finalizer != nil {
+			defer func() {
+				ctx = context.WithValue(ctx, ContextKeyResponseHeaders, resp.Header)
+				ctx = context.WithValue(ctx, ContextKeyResponseSize, resp.ContentLength)
+				c.finalizer(ctx, resp.StatusCode, req)
+			}()
+		}
```
peterbourgon: I'm not sure it's correct to create zero-value http.Requests and Responses; my intuition is that they need to be created by package http in an e.g. NewReqeust constructor. But that's fine, I think we can get by without them. Going with the change in the ClientFinalizerFunc below, perhaps this can look like..

```go
var (
    resp *http.Response
    err  error
)
if c.finalizer != nil {
    defer func() {
        if resp != nil {
            ctx = context.WithValue(ctx, ContextKeyResponseHeaders, resp.Header)
            ctx = context.WithValue(ctx, ContextKeyResponseSize, resp.ContentLength)
        }
        c.finalizer(ctx, err)
    }()
}
```

What do you think?

transport/http/client_test.go
```diff
@@ -140,6 +140,62 @@ func TestHTTPClientBufferedStream(t *testing.T) {
 	}
 }
 
+func TestClientFinalizer(t *testing.T) {
+	var (
+		headerKey    = "X-Henlo-Lizer"
+		headerVal    = "Helllo you stinky lizard"
```
peterbourgon: It's ridiculous how my stream-of-consciousness garbage thoughts when I was writing this test get propagated into the infinite future 🤙

transport/http/client.go
travissalascox: makes sense to me. the main reason i went with the signature of the ServerFinalizer was from a lack of how to tackle the problem. But it makes sense to me to have a signature like that. I'll update the PR later this evening and push up the requested changes. Thanks for the comments 😸 !

travissalascox:
@peterbourgon updated based your comments.

peterbourgon:
Sweet! Thanks a ton!

PR #538, opened 2017-05-26
fix(example/shipping): header should set before `WriteHeader`

### LINKED ISSUE #537: Should not set header after `WriteHeader` in example
In shipping example: https://github.com/go-kit/kit/blame/master/examples/shipping/booking/transport.go#L184

`encodeError` may write header before setting the `content-type` header. From go net/http's [doc][write-header-doc]:

> Changing the header map after a call to WriteHeader (or Write) has no effect unless the modified headers are trailers.

Seems like the `content-type` header should set before the error switch.

[write-header-doc]: https://golang.org/pkg/net/http/#ResponseWriter

bcho:
From `net/http`:

    Changing the header map after a call to WriteHeader (or Write)
    has no effect unless the modified headers are trailers.

To send the `content-type`, need to set before calling `WriteHeader`.

Fixes #537

### DIFF
--- a/examples/shipping/booking/transport.go
+++ b/examples/shipping/booking/transport.go
@@ -179,6 +179,7 @@ type errorer interface {
 
 // encode errors from business-logic
 func encodeError(_ context.Context, err error, w http.ResponseWriter) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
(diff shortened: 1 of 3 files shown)


peterbourgon:
Oh, perfect, thank you!

Issue #529, opened 2017-05-15
Cloudwatch metrics does not limit nr. of MetricDatums to 20 per call


feliksik:
According to the [AWS Docs](http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html), a max of 20 metricdatums can be sent in an API request. 

So this [code should](https://github.com/go-kit/kit/blob/master/metrics/cloudwatch/cloudwatch.go#L136) respect that. 

I might get to fixing this myself, but not today...

cam-stitt:
Good find. Thanks for this.

alpeb:
Would it suffice if I add checks for that in `NewCounter`, `NewGauge` and `NewHistogram` that would panic if that limit is reached?

cam-stitt:
I'm not sure if a panic would be ideal in this situation. We should be able to split it into a max of 20 datums and spread it to multiple api requests.

alpeb:
Sounds good. Can I have the honor of submitting a PR for this, or you prefer to tackle it yourself? :)
I was thinking we could do it in batches of concurrent API requests, say up to 4 concurrent requests per batch. Let me know what you think.

cam-stitt:
Go for it! Happy to share. I would suggest making an option `NumConcurrentRequests` to specify how many requests can be sent per batch.

cam-stitt:
@peterbourgon I believe we can probably close this off now?

peterbourgon:
Jawohl.

PR #540, opened 2017-05-30
Issue #529: Partition API requests to CloudWatch into separate concur…

alpeb:
…rent batches to circumvent the 20 data per request limit that they have

### DIFF
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -14,6 +14,10 @@ import (
 	"github.com/go-kit/kit/metrics/generic"
 )
 
+const (
+	maxConcurrentRequests = 20
+)
+
 // CloudWatch receives metrics observations and forwards them to CloudWatch.
 // Create a CloudWatch object, use it to create metrics, and pass those metrics as
 // dependencies to the components that will use them.
@@ -21,6 +25,7 @@ import (
 // To regularly report metrics to CloudWatch, use the WriteLoop helper method.
 type CloudWatch struct {
 	mtx                   sync.RWMutex
+	sem                   chan struct{}
 	namespace             string
 	numConcurrentRequests int
 	svc                   cloudwatchiface.CloudWatchAPI
@@ -32,27 +37,25 @@ type CloudWatch struct {
 
 // New returns a CloudWatch object that may be used to create metrics.
 // Namespace is applied to all created metrics and maps to the CloudWatch namespace.
+// NumConcurrent sets the number of simultaneous requests to Amazon.
+// A good default value is 10 and the maximum is 20.
 // Callers must ensure that regular calls to Send are performed, either
 // manually or with one of the helper methods.
-func New(namespace string, svc cloudwatchiface.CloudWatchAPI, logger log.Logger) *CloudWatch {
+func New(namespace string, svc cloudwatchiface.CloudWatchAPI, numConcurrent int, logger log.Logger) *CloudWatch {
+	if numConcurrent > maxConcurrentRequests {
+		numConcurrent = maxConcurrentRequests
+	}
+
 	return &CloudWatch{
+		sem:                   make(chan struct{}, numConcurrent),
 		namespace:             namespace,
-		numConcurrentRequests: 10,
+		numConcurrentRequests: numConcurrent,
 		svc:        svc,
 		counters:   map[string]*counter{},
 		gauges:     map[string]*gauge{},
 		histograms: map[string]*histogram{},
 		logger:     logger,
 	}
-}
-
-// SetConcurrency overrides the default number (10) of concurrent requests sent to CloudWatch.
-// CloudWatch allows a maximum of 20 metrics to be sent per request, so when Send is called,
-// we partition the metrics and concurrently call their API. This parameter sets maximum number
-// of concurrent requests.
-func (cw *CloudWatch) SetConcurrency(numConcurrentRequests int) *CloudWatch {
-	cw.numConcurrentRequests = numConcurrentRequests
-	return cw
 }
 
 // NewCounter returns a counter. Observations are aggregated and emitted once
@@ -144,28 +147,30 @@ func (cw *CloudWatch) Send() error {
 		}
 	}
 
-	var tokens = make(chan struct{}, cw.numConcurrentRequests)
-	var errors = make(chan error)
-	var n int
-
+	var batches [][]*cloudwatch.MetricDatum
 	for len(datums) > 0 {
 		var batch []*cloudwatch.MetricDatum
-		lim := min(len(datums), cw.numConcurrentRequests)
+		lim := min(len(datums), maxConcurrentRequests)
 		batch, datums = datums[:lim], datums[lim:]
-		n++
+		batches = append(batches, batch)
+	}
+
+	var errors = make(chan error, len(batches))
+	for _, batch := range batches {
 		go func(batch []*cloudwatch.MetricDatum) {
-			tokens <- struct{}{}
+			cw.sem <- struct{}{}
+			defer func() {
+				<-cw.sem
+			}()
 			_, err := cw.svc.PutMetricData(&cloudwatch.PutMetricDataInput{
 				Namespace:  aws.String(cw.namespace),
 				MetricData: batch,
 			})
-			<-tokens
 			errors <- err
 		}(batch)
 	}
-
 	var firstErr error
-	for ; n > 0; n-- {
+	for i := 0; i < cap(errors); i++ {
 		if err := <-errors; err != nil && firstErr != nil {
 			firstErr = err
 		}
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -20,13 +20,14 @@ import (
 //
 // To regularly report metrics to CloudWatch, use the WriteLoop helper method.
 type CloudWatch struct {
-	mtx        sync.RWMutex
-	namespace  string
-	svc        cloudwatchiface.CloudWatchAPI
-	counters   map[string]*counter
-	gauges     map[string]*gauge
-	histograms map[string]*histogram
-	logger     log.Logger
+	mtx                   sync.RWMutex
+	namespace             string
+	numConcurrentRequests int
+	svc                   cloudwatchiface.CloudWatchAPI
+	counters              map[string]*counter
+	gauges                map[string]*gauge
+	histograms            map[string]*histogram
+	logger                log.Logger
 }
 
 // New returns a CloudWatch object that may be used to create metrics.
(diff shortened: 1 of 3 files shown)


alpeb:
@cam-stitt Looking forward for your comments.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
 		}
 	}
 
-	_, err := cw.svc.PutMetricData(&cloudwatch.PutMetricDataInput{
-		Namespace:  aws.String(cw.namespace),
-		MetricData: datums,
-	})
-	return err
+	var tokens = make(chan struct{}, cw.numConcurrentRequests)
```
cam-stitt: I'm not sure what the value of `tokens` is. I don't really see it being used.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
+	var errors = make(chan error)
+	var n int
+
+	for len(datums) > 0 {
```
cam-stitt: My personal preference would be to loop in jumps of `numConcurrentRequests` and then slicing datums up each time. Rough example: https://play.golang.org/p/slrBko_QdI

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
+			_, err := cw.svc.PutMetricData(&cloudwatch.PutMetricDataInput{
+				Namespace:  aws.String(cw.namespace),
+				MetricData: batch,
+			})
+			<-tokens
+			errors <- err
+		}(batch)
+	}
+
+	var firstErr error
```
cam-stitt: By returning only the first error, there may be some context lost as to what the overall problem was. Maybe we can create an Error type that actually contains multiple errors?

cam-stitt:
Thanks a lot for this. It's a great start. I've put a few of my thoughts on there, but would love to get @peterbourgon's opinion.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
+	var tokens = make(chan struct{}, cw.numConcurrentRequests)
+	var errors = make(chan error)
+	var n int
+
+	for len(datums) > 0 {
+		var batch []*cloudwatch.MetricDatum
+		lim := min(len(datums), cw.numConcurrentRequests)
+		batch, datums = datums[:lim], datums[lim:]
+		n++
+		go func(batch []*cloudwatch.MetricDatum) {
```
cam-stitt: I wonder if it's better to not make each PutMetricData in a goroutine, and just fire them and return the error that comes up. This means that the first request to error will stop any future requests from occurring, which might be nicer behaviour.

metrics/cloudwatch/cloudwatch.go
alpeb: It limits the number of concurrent requests. When the channel gets filled, `tokens <- struct{}{}` will block until a go routine finishes and releases a token through `<-tokens`

metrics/cloudwatch/cloudwatch.go
alpeb: Sure. Would you agree on limiting the number of errors, or shall we return everything?

metrics/cloudwatch/cloudwatch.go
alpeb: If you step through my code, you'll see it's pretty much equivalent to what you're proposing.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -44,6 +46,15 @@ func New(namespace string, svc cloudwatchiface.CloudWatchAPI, logger log.Logger)
...
 }
 
+// SetConcurrency overrides the default number (10) of concurrent requests sent to CloudWatch.
+// CloudWatch allows a maximum of 20 metrics to be sent per request, so when Send is called,
+// we partition the metrics and concurrently call their API. This parameter sets maximum number
+// of concurrent requests.
+func (cw *CloudWatch) SetConcurrency(numConcurrentRequests int) *CloudWatch {
+	cw.numConcurrentRequests = numConcurrentRequests
+	return cw
+}
```
peterbourgon: This should be an option in the constructor, rather than a method on the struct. I don't see any reason to make it modifiable during runtime. (It's also racy as implemented.)

metrics/cloudwatch/cloudwatch.go
peterbourgon: Similarly, this should be global per CloudWatch struct (so a member of the struct), not per-metric (scoped to NewCounter), right?

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
-	})
-	return err
+	var tokens = make(chan struct{}, cw.numConcurrentRequests)
+	var errors = make(chan error)
+	var n int
+
+	for len(datums) > 0 {
+		var batch []*cloudwatch.MetricDatum
+		lim := min(len(datums), cw.numConcurrentRequests)
+		batch, datums = datums[:lim], datums[lim:]
```
peterbourgon: If I'm reading this right, you're sizing the batch as the minimum of (number of datums remaining) and (number of allowable concurrent requests)? That doesn't seem right—batch size and concurrent batches have nothing to do with each other, do they?

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
+		n++
+		go func(batch []*cloudwatch.MetricDatum) {
+			tokens <- struct{}{}
+			_, err := cw.svc.PutMetricData(&cloudwatch.PutMetricDataInput{
+				Namespace:  aws.String(cw.namespace),
+				MetricData: batch,
+			})
+			<-tokens
+			errors <- err
+		}(batch)
```
peterbourgon: Since you construct the batch within the loop scope, there's no reason to pass it as a parameter to this anonymous function.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
+		batch, datums = datums[:lim], datums[lim:]
+		n++
+		go func(batch []*cloudwatch.MetricDatum) {
+			tokens <- struct{}{}
+			_, err := cw.svc.PutMetricData(&cloudwatch.PutMetricDataInput{
+				Namespace:  aws.String(cw.namespace),
+				MetricData: batch,
+			})
+			<-tokens
+			errors <- err
```
peterbourgon: Is there a reason you decrement the semaphore _before_ returning the error along the errors chan? If you didn't need to do that, it would be nicer to express this as

```go
tokens <- struct{}{}
defer func() { <-tokens }()
_, err := cw...
errors <- err
```

metrics/cloudwatch/cloudwatch.go
```diff
@@ -133,11 +144,41 @@ func (cw *CloudWatch) Send() error {
...
+		}(batch)
+	}
+
+	var firstErr error
+	for ; n > 0; n-- {
+		if err := <-errors; err != nil && firstErr != nil {
+			firstErr = err
+		}
+	}
```
peterbourgon: It's a little fragile, for a couple reasons:

- n is only loosely coupled to the number of goroutines you launched, easy to get out of sync w/ maintenance
- If you forget to do precisely as many reads from errors as goroutines you launched, you'll leak goroutines

Suggest precomputing batches, constructing errors as `make(chan error, len(batches))`, and only then launching the goroutines to process the batches. Finally, changing this loop to e.g.

```go
for i := 0; i < cap(errors); i++ {
    if err := <-errors; err != nil ... { ... }
}
```

alpeb:
Thanks for the detailed review! I'll respond to comments inline, and then I'll submit a new patch.

metrics/cloudwatch/cloudwatch.go
alpeb: Right, I wasn't sure how to handle optional parameters in go. Would this do?
```
func New(namespace string, svc cloudwatchiface.CloudWatchAPI, logger log.Logger, numConcurrentRequests ...int) *CloudWatch {
         numConcurrentRequestsValue := 10
         if len(numConcurrentRequests) > 0 {
                 numConcurrentRequestsValue = numConcurrentRequests[0]
         }
          return &CloudWatch{
                  namespace:             namespace,
                 numConcurrentRequests: numConcurrentRequestsValue,
                  svc:        svc,
                  counters:   map[string]*counter{},
                  gauges:     map[string]*gauge{},
                  histograms: map[string]*histogram{},
                  logger:     logger,
          }
  }
```

peterbourgon: No, that's not the way. If I were you I'd just make it a non-optional parameter, and document a reasonable default in the function comment.

```go
// New returns a blah blah.
// The client will emit at most numConcurrent simultaneous requests to Amazon.
// A good default value is 10.
func New(namespace string, svc CloudWatchAPI, numConcurrent int, logger log.Logger) {
    ...
}
```

metrics/cloudwatch/cloudwatch.go
alpeb: Sorry, I don't understand what you mean. This is inside `Send()`, why are you referring to `NewCounter`?

metrics/cloudwatch/cloudwatch.go
alpeb: You're right, I should have instead got the min between `len(datums)` and 20 (the maximum number of datums allowed per request).

metrics/cloudwatch/cloudwatch.go
alpeb: Yep, that's better.

metrics/cloudwatch/cloudwatch.go
alpeb: Indeed 👍

metrics/cloudwatch/cloudwatch.go
alpeb: Understood. I'll follow suit.

metrics/cloudwatch/cloudwatch.go
peterbourgon: Sorry, my mistake—but my point stands: putting the semaphore here means there are numConcurrentRequest simultanous requests per _Send_ invocation, so if the user invokes Send 10 times then you get 10x numConcurrentRequests going out. So in reality I think we want the semaphore to be in the struct.

alpeb: Gotcha 💯

peterbourgon:
```
examples/addsvc/cmd/addcli/main.go:155: multiple-value 
transportFactory.GetTransport() in single-value context
```

I think if you rebase you might get this fixed for you?

alpeb:
Yep, after rebasing checks are passing 🎉

peterbourgon:
@cam-stitt I just want to get a +1 from you before merging. All good from your POV?

cam-stitt:
Yeah, looks good. Thanks @alpeb for jumping on this.

peterbourgon:
Thanks all!

Issue #545, opened 2017-06-06
sd/eureka: tests are incredibly flaky on CI


peterbourgon:
Cmd-F for `FAIL:` on each of these:

- https://circleci.com/gh/go-kit/kit/1222
- https://circleci.com/gh/go-kit/kit/1223
- https://circleci.com/gh/go-kit/kit/1224

I can't reproduce the failures locally. I suspect it is a nondeterminsm/timing issue related to the underlying Fargo library. It started happening after we merged the new SD PR, but I'm not sure that was the cause so much as the trigger. Could someone with more domain knowledge take a look?

@yurishkuro @martinbaillie @seh

yurishkuro:
I'll take a look. Interestingly, it's the unit test that fails, not integration tests, but the latter could use some improvement as they use unconditional sleep().

seh:
I finally got CircleCI to display the failure, so I'm looking into it now too.

seh:
The most mystifying part I've found so far is why [this unit test expects only one instance to be present](https://github.com/go-kit/kit/blob/master/sd/eureka/instancer_test.go#L29). It looks to me like expecting two is more reasonable, given that [the application has two instances](https://github.com/go-kit/kit/blob/master/sd/eureka/util_test.go#L26-L29).

seh:
@yurishkuro, it looks like the one is a holdover in commit 432e29216b6440d7fc2a819478ee200ab357e550. The test could pass mistakenly if it retrieves the cache's state before the goroutine feeding the updates to the cache gets a chance to run for the first time. That's why [there's an "await" parameter for `fargo.ScheduleAppUpdates`](https://github.com/hudl/fargo/blob/a34ea4b0415d575082bd01ec7565441a56c0d012/connection.go#L120-L128): to allow for coordination like this.

yurishkuro:
@seh unless there is an error retrieving the instances in NewInstancer(), the state should be initialized before NewInstancer returns

yurishkuro:
But the following update could add 2 instances instead of one given to connection at construction, and that's certainly a race condition, the instancer could get either of those two states

seh:
`eureka.NewInstancer` [starts a goroutine here](https://github.com/go-kit/kit/blob/master/sd/eureka/instancer.go#L44). That goroutine may not get a chance to run before `NewInstancer` completes and the test goes on to probe the cache.

seh:
[...]

PR #547, opened 2017-06-07
sd/eureka: remove data races in unit tests

seh:
Addressing #545, revise the fake _fargo_ components used in the Eureka-related tests to behave more like a real Eureka server, and reduce the tests' vulnerability to delays in asynchronously-arriving responses from the server by polling, with a limit in place to make the tests fail after polling for one second.

This proposal changes `eureka.Instancer`'s cache priming implementation, though the behavior observable by callers shouldn't change.


peterbourgon:
Closes #545.

Nice!

PR #548, opened 2017-06-07
Replace sleep with periodic checks

yurishkuro:


peterbourgon:
LGTM. Is this pending anything else or can I merge?

yurishkuro:
It can merge. I originally thought it was the cause of flakiness in CircleCI, so I made this change, but turns out CI was failing on regular unit tests, not this integration test. But it still seems a better way than standalone sleep.

Issue #551, opened 2017-06-08
Catch panics in http Server


esenac:
Hi, 
it could be useful to recover from panic in `transport/http.Server.ServeHTTP`. Panic could be caused by errors in encoding or decoding functions or in general in the code of the underlying service. Recover would allow to return a proper HTTP status code (i.e. a 500) instead of dropping the connection.
This is mostly useful for debugging purposes, in the current situation you won't get evidence of the occurred panic in the logs.

esenac:
A possible example could be the following:
```
// ServeHTTP implements http.Handler.
func (s Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	ctx := r.Context()

	if s.finalizer != nil {
		iw := &interceptingWriter{w, http.StatusOK, 0}
		defer func() {
			ctx = context.WithValue(ctx, ContextKeyResponseHeaders, iw.Header())
			ctx = context.WithValue(ctx, ContextKeyResponseSize, iw.written)
			s.finalizer(ctx, iw.code, r)
		}()
		w = iw
	}

	defer func() {
		if r := recover(); r != nil {
			ctx = context.WithValue(ctx, ContextKeyRecoveredFromPanic, r)
		}
	}()
....
```
In this way the error put in the context could be accessed and handled in the _finalizer_ without changing the current behaviour.

peterbourgon:
Why should we give special treatment to ServeHTTP? Why shouldn't we do the same recovery dance before descending into any call stack?

ChrisHines:
What to do with panics is a _policy_ choice that the main application should be allowed to make, which means that at the very least is would have to be configurable. I wonder if a standalone panic handler isn't a better approach. See the top answer to this SO question for an example: https://stackoverflow.com/questions/28745648/global-recover-handler-for-golang-http-panic

esenac:
@peterbourgon we are trying to introduce gokit as a standard for microservices in our company and we are fully based on REST over HTTP, so we have knowledge on that part and we focused just on `ServeHTTP`. Sure it would be a better option to have a common approach to panic handling for all transports.

@ChrisHines we actually developed a solution like the one you propose in our microservices, but we think that it could be useful that the framework would provide a common solution for this problem. The one we proposed would let the choice of what to do with panics to the main application by checking the key in the context.
We came to this proposal because the current behaviour of the gokit framework is just to silence all the panics, making it hard to discover the misbehaviour of the applications.

ChrisHines:
[...]

PR #554, opened 2017-06-13
Catch panic in http and grpc servers

esenac:
As discussed in #551, this PR adds recovering from panic capability to both `kit/transport/http.Server` and `kit/transport/grpc.Server`. In the first case, the value given to panic is also put in context in order to be handled by the `finalizer`. In both cases the panic is logged.

### DIFF
--- a/transport/grpc/server.go
+++ b/transport/grpc/server.go
@@ -18,12 +18,13 @@ type Handler interface {
 
 // Server wraps an endpoint and implements grpc.Handler.
 type Server struct {
-	e      endpoint.Endpoint
-	dec    DecodeRequestFunc
-	enc    EncodeResponseFunc
-	before []ServerRequestFunc
-	after  []ServerResponseFunc
-	logger log.Logger
+	e          endpoint.Endpoint
+	dec        DecodeRequestFunc
+	enc        EncodeResponseFunc
+	before     []ServerRequestFunc
+	after      []ServerResponseFunc
+	catchPanic bool
+	logger     log.Logger
 }
 
 // NewServer constructs a new server, which implements wraps the provided
@@ -70,14 +71,23 @@ func ServerErrorLogger(logger log.Logger) ServerOption {
 	return func(s *Server) { s.logger = logger }
 }
 
+// ServerCatchPanic is used to enable the panic catching and logging functionality
+// in the server. Panic value will be set in ContextKeyPanicValue context's key.
+// By default, panic handling is left to the standard library.
+func ServerCatchPanic(enabled bool) ServerOption {
+	return func(s *Server) { s.catchPanic = enabled }
+}
+
 // ServeGRPC implements the Handler interface.
 func (s Server) ServeGRPC(ctx oldcontext.Context, req interface{}) (oldcontext.Context, interface{}, error) {
 
-	defer func() {
-		if r := recover(); r != nil {
-			s.logger.Log("panic", r)
-		}
-	}()
+	if s.catchPanic {
+		defer func() {
+			if r := recover(); r != nil {
+				s.logger.Log("panic", r)
+			}
+		}()
+	}
 
 	// Retrieve gRPC metadata.
 	md, ok := metadata.FromContext(ctx)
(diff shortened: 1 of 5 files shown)


transport/grpc/server.go
```diff
@@ -72,6 +72,13 @@ func ServerErrorLogger(logger log.Logger) ServerOption {
 
 // ServeGRPC implements the Handler interface.
 func (s Server) ServeGRPC(ctx oldcontext.Context, req interface{}) (oldcontext.Context, interface{}, error) {
+
+	defer func() {
+		if r := recover(); r != nil {
+			s.logger.Log("panic", r)
+		}
+	}()
```
peterbourgon: It's important that this kind of functionality is opt-in. By default, I don't want my ServeGRPC (or ServeHTTP) to be catching, logging, and squelching panics.

transport/http/server.go
```diff
@@ -89,6 +89,10 @@ func (s Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			ctx = context.WithValue(ctx, ContextKeyResponseHeaders, iw.Header())
 			ctx = context.WithValue(ctx, ContextKeyResponseSize, iw.written)
+			if r := recover(); r != nil {
+				ctx = context.WithValue(ctx, ContextKeyRecoveredFromPanic, r)
+				s.logger.Log("panic", r)
+			}
```
peterbourgon: Likewise above.

transport/http/request_response_funcs.go
```diff
@@ -130,4 +130,8 @@ const (
 	// ContextKeyResponseSize is populated in the context whenever a
 	// ServerFinalizerFunc is specified. Its value is of type int64.
 	ContextKeyResponseSize
+
+	// ContextKeyRecoveredFromPanic is populated in the context whenever a
+	// panic is recovered in ServeHTTP. Its value is the one given to panic.
+	ContextKeyRecoveredFromPanic
```
peterbourgon: I think a better name here is something like ContextKeyPanicValue.

esenac:
@peterbourgon I implemented your suggestions. Please have a further review, thanks!

peterbourgon:
@ChrisHines @basvanbeek Any thoughts on this one?

basvanbeek:
I'm not sold on adding this within the Go kit endpoint service handlers. It is incredibly trivial to make a recover wrapper around the transport NewServer functions.

ChrisHines:
For the HTTP case I think it is nice that you can handle panic values, response codes, and similar concerns all in one finalizer rather than do some of it in a finalizer and handle panics with a decorator. I do wonder why gRPC doesn't have the finalizer concept also, or why we chose it for HTTP rather use a decorator approach.

In other words, I think it is important to have a way to do something post-response that has access to all the response meta-data. It would be nice if that feature worked similarly across transports. I am less particular about the particular design of that feature. Most of Go kit has utilized decorator patterns, but the transport packages seem to favor template method designs.

basvanbeek:
The finalizer concept was added later by someone who needed it for HTTP. gRPC was omitted at that time. I do think regardless of what we end up doing for panics we should add finalizer logic to gRPC so the transports are more in line again with their feature sets.

peterbourgon:
> I do wonder why ... we chose [finalizers] for HTTP rather use a decorator approach.

We had sort of an imperative precedent established in transport/http.Server.ServeHTTP, with the serial execution of before funcs, decoding, endpoint execution, after funcs, and encoding. It was most straightforward to me at the time to implement the as part of that imperative flow, so to speak. It definitely could have been layered in as a ServeHTTP decorator, and I reckon we could even switch to that style in the future, if there are advantages to it: the user-facing API of the ServerFinalizerFunc wouldn't have to change, I don't think.

spaghetty:
any news on this PR?

peterbourgon:
The implementation is fine, but I'm still not convinced it makes sense to include. I'll think on it a bit more seriously and try to have a final answer by the weekend. Is anyone blocked and waiting on this? If so, can you write down the details of your use-case here, for posterity?

spaghetty:
No actually we solve using a standard HandlerFunc middleware, it just feels non standard and it would be better to have a clear decision on that.

thanks

esenac:
Any news? Thanks in advance

peterbourgon:
Sorry for the delay—GopherCon and all that. After a fair bit of thought I think the best course of action here is no action: I think if you want to do panic recovery on a per-transport basis, you can use per-transport middlewares pretty effectively. I've [added an entry to the gokit.io FAQ](http://gokit.io/faq/#panics-mdash-how-should-my-service-handle-panics) to that end. I hope this suits. Thanks for your patience and hard work on the PR!

PR #564, opened 2017-07-01
auth/jwt: prevent concurrent reads and writes on MapClaims

### LINKED ISSUE #562: auth/jwt causes "fatal error: concurrent map writes" on concurrent requests
Hello,

When having multiple concurrent requests to the same endpoint that contains the parser, it will cause `fatal error: concurrent map writes` since the same claims will be accessed by all requests at the same time. Here is some code to reproduce it:

```
package main

import (
	"context"

	"time"

	jwt "github.com/dgrijalva/jwt-go"
	jwtkit "github.com/go-kit/kit/auth/jwt"
)

func main() {
	testEndpoint := MakeTestEndpoint()
	testEndpoint = JWTParserMiddleware("secret")(testEndpoint)

	for i := 0; i < 10; i++ {
		go func() {
			ctx := context.WithValue(context.Background(), jwtkit.JWTTokenContextKey, "eyJhbGciOiJIUzI1NiIsImtpZCI6ImtpZCIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjoiZ28ta2l0In0.14M2VmYyApdSlV_LZ88ajjwuaLeIFplB8JpyNy0A19E")
			testEndpoint(ctx, nil)
		}()
	}

	select {}
}

func JWTParserMiddleware(secret string) endpoint.Middleware {
	kf := func(token *jwt.Token) (interface{}, error) {
		return []byte(secret), nil
	}

	return jwtkit.NewParser(kf, jwt.SigningMethodHS256, jwt.MapClaims{})
}

func MakeTestEndpoint() endpoint.Endpoint {
		time.Sleep(1 * time.Second)
		return nil, nil
	}
}
```

Maybe use a factory instead of directly passing the claims?

Thanks!

peterbourgon:
Closes #562.

### DIFF
--- a/auth/jwt/middleware.go
+++ b/auth/jwt/middleware.go
@@ -66,11 +66,27 @@ func NewSigner(kid string, key []byte, method jwt.SigningMethod, claims jwt.Clai
 	}
 }
 
+// ClaimsFactory is a factory for jwt.Claims.
+// Useful in NewParser middleware.
+type ClaimsFactory func() jwt.Claims
+
+// MapClaimsFactory is a ClaimsFactory that returns 
+// an empty jwt.MapClaims.
+func MapClaimsFactory() jwt.Claims {
+    return jwt.MapClaims{}
+}
+
+// StandardClaimsFactory is a ClaimsFactory that returns 
+// an empty jwt.StandardClaims.
+func StandardClaimsFactory() jwt.Claims {
+    return &jwt.StandardClaims{}
+}
+
 // NewParser creates a new JWT token parsing middleware, specifying a
 // jwt.Keyfunc interface, the signing method and the claims type to be used. NewParser
 // adds the resulting  claims to endpoint context or returns error on invalid token.
 // Particularly useful for servers.
-func NewParser(keyFunc jwt.Keyfunc, method jwt.SigningMethod, claims jwt.Claims) endpoint.Middleware {
+func NewParser(keyFunc jwt.Keyfunc, method jwt.SigningMethod, newClaims ClaimsFactory) endpoint.Middleware {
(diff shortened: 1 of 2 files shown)


litriv:
@peterbourgon looks like the test is in place and working (failing).  What do you think of passing a claim factory to NewParser, that can be be used to create a new jwt.Claims, as in 
```
token, err := jwt.ParseWithClaims(tokenString, newClaims(), func(token *jwt.Token) (interface{}, error) {
   ...
})
```

If agreed, then I will make the change, test my side and commit to this pull request if you are too busy.

peterbourgon:
Yes, sorry, I got the failing test but then failed to find the time to fix it ;)

That's exactly what's needed. Probably best to define a new type ClaimFactory func(...) jwt.Claim,  take one in the constructor, and probably provide a default EmptyClaimFactory implementation — or whatever would be most appropriate, I'm not a JWT expert.

litriv:
@peterbourgon  sounds about right.  I'm on it  :+1:

litriv:
@peterbourgon this is now fixed.  how shall i go about getting the changes to the repo?  probably easiest if you can give me temporary write access so i can just push to this branch?

basvanbeek:
why not just fork it to your account and submit a new PR referencing this one?

litriv:
@basvanbeek i created a pull request to merge with this branch, but if it's better to create one for mergin with master, then let me know and i'll do it.  thanks

B3rs:
Hi @peterbourgon, I just encountered the same problem, any clue on when this PR will be merged into master? thanks

peterbourgon:
How about now?

PR #569, opened 2017-07-11
[WIP] A potential re-refactor of package metrics

peterbourgon:
Goals:

- A more baked-in Provider abstraction (see #513)
- All implementations should fully support With

### DIFF
--- a/metrics2/expvar/expvar.go
+++ b/metrics2/expvar/expvar.go
@@ -1,0 +1,278 @@
+package expvar
+
+import (
+	"expvar"
+	"sync"
+
+	"github.com/go-kit/kit/metrics2"
+	"github.com/go-kit/kit/metrics2/internal/histogram"
+	"github.com/go-kit/kit/metrics2/internal/template"
+)
+
+// Provider acts as a store of expvar metrics.
+//
+type Provider struct {
+	mtx        sync.Mutex
+	ints       map[string]*expvar.Int
+	floats     map[string]*expvar.Float
+	histograms map[string]*histogram.Histogram //
+}
+
+// NewProvider returns a new, empty provider.
+func NewProvider() *Provider {
+	return &Provider{
+		ints:       map[string]*expvar.Int{},
+		floats:     map[string]*expvar.Float{},
+		histograms: map[string]*histogram.Histogram{},
+	}
+}
+
+// NewIntCounter returns a Counter whose values are truncated and exposed as
+// expvar.Int. The name can include template interpolation to support With; see
+// package documentation for details.
+func (p *Provider) NewIntCounter(name string) *IntCounter {
+	return &IntCounter{
+		parent: p,
+		name:   name,
+		fields: map[string]string{},
+	}
+}
+
+// NewFloatCounter returns a Counter whose values are exposed as an
+// expvar.Float. The name can include template interpolation to support With;
+// see package documentation for details.
+func (p *Provider) NewFloatCounter(name string) *FloatCounter {
+	return &FloatCounter{
+		parent: p,
+		name:   name,
+		fields: map[string]string{},
+	}
+}
+
+// NewIntGauge returns a Gauge whose values are truncated and exposed as
+// expvar.Int. The name can include template interpolation to support With; see
+// package documentation for details.
+func (p *Provider) NewIntGauge(name string) *IntGauge {
+	return &IntGauge{
+		parent: p,
+		name:   name,
+		fields: map[string]string{},
+	}
+}
+
+// NewFloatGauge returns a Gauge whose values are exposed as an expvar.Float.
+// The name can include template interpolation to support With; see package
+// documentation for details.
+func (p *Provider) NewFloatGauge(name string) *FloatGauge {
+	return &FloatGauge{
+		parent: p,
+		name:   name,
+		fields: map[string]string{},
+	}
+}
+
+// NewHistogram returns a Histogram whose observations are collected and exposed
+// as 4 per-quantile expvar.Floats. The exposed quantiles are 50th, 90th, 95th,
+// and 99th percentile, with names suffixed by _p50, _p90, _p95, and _p99,
+// respectively. The name can include template interpolation to support With;
+// see package documentation for details.
+func (p *Provider) NewHistogram(name string) *Histogram {
+	return &Histogram{
+		parent: p,
+		name:   name,
+		fields: map[string]string{},
+	}
+}
+
+func (p *Provider) int(name string) *expvar.Int {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+	if _, ok := p.ints[name]; !ok {
+		p.ints[name] = expvar.NewInt(name)
+	}
+	return p.ints[name]
+}
+
+func (p *Provider) float(name string) *expvar.Float {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+	if _, ok := p.floats[name]; !ok {
+		p.floats[name] = expvar.NewFloat(name)
+	}
+	return p.floats[name]
+}
+
+func (p *Provider) observe(name string, value float64) {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+
+	h, ok := p.histograms[name]
+	if !ok {
+		h = histogram.New()
+		p.histograms[name] = h
+	}
+	h.Observe(value)
+
+	for _, pair := range []struct {
+		suffix   string
+		quantile float64
+	}{
+		{"_p50", 0.50},
+		{"_p90", 0.90},
+		{"_p95", 0.95},
+		{"_p99", 0.99},
+	} {
+		fullname := name + pair.suffix
+		f, ok := p.floats[fullname]
+		if !ok {
+			f = expvar.NewFloat(fullname)
+			p.floats[fullname] = f
+		}
+		f.Set(h.Quantile(pair.quantile))
+	}
+}
+
+// IntCounter is a Counter whose values are truncated to integers and exposed as
+// expvar.Int. The zero value of an IntCounter is not useful; you must construct
+// it via the Provider.
+type IntCounter struct {
+	parent *Provider
+	name   string
+	fields map[string]string
+}
+
+// With implements Counter.
+func (c *IntCounter) With(keyvals ...string) metrics.Counter {
+	return &IntCounter{
+		parent: c.parent,
+		name:   c.name,
+		fields: appendFields(c.fields, keyvals...),
+	}
+}
+
+// Add inplements Counter.
+func (c *IntCounter) Add(delta float64) {
+	name := template.Render(c.name, c.fields)
+	c.parent.int(name).Add(int64(delta))
+}
+
+// FloatCounter is a Counter whose values are exposed as expvar.Float. The zero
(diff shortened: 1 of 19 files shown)


yurishkuro:
@peterbourgon is the intention to have a common Provider interface that all backends implement?

If so, I ran into an issue today with doing it for Prometheus (https://github.com/uber/jaeger-lib/pull/19) because Prometheus insist on pre-declaring all the labels (tag keys). What are your thoughts on that?

peterbourgon:
@yurishkuro My intent is definitely to have each backend implement a Provider interface, but I haven't thought it through deeply enough with Prometheus to know exactly what shape it will take there. Or even if it will be feasible. My intuition at the moment is that the Prometheus Provider will require a full enumeration of labels at construction time, and at observation time drop those label keys that weren't previously declared. (And substitute e.g. unknown for declared label keys that weren't provided.)

seh:
I've only skimmed it so far (mostly the Prometheus-related parts), but I like where this is going.

metrics2/statsd/statsd.go
```diff
@@ -0,0 +1,359 @@
...
+// StatsD format. WriteTo is best-effort and fails fast; observations are lost
+// if there's a problem with the write. Clients should be sure to call WriteLoop
+// regularly, ideally through the SendLoop or WriteLoop helper methods.
+//
+// This is a low-level function, primarily intended for testing. Most callers
+// should prefer SendLoop.
+func (p *Provider) WriteTo(w io.Writer) (int64, error) {
+	// Copy the maps and reset them to empty.
+	// Do this in a closure to minimize lock time.
+	c, g, t, h := func() (c, g map[string]float64, t, h map[string][]float64) {
```
darrenmcc: I don't see how using a closure minimizes lock time. If anything it increases lock time due to the use of `defer`.

peterbourgon: The point is that you only hold the lock to copy the data out, not when you're writing the data to the remote.

darrenmcc: By remote you mean the Provider fields? I still don't see why you can't just remove the closure and wrap the 4 copy operations with the mutex.

peterbourgon: That's exactly what I'm doing. The closure gives me nice semantics with defer. Believing that defer is a performance penalty is ridiculous in most cases, this one included.

darrenmcc: The code is more complex and less performant than necessary, even if mildly so. That's all I'm trying to point out.

feliksik: I think there are is a tradeoff here: 
* performance; the code is more performant by not doing aggressive locking. This is much more significant than the overhead of defer. Even more so because the provider lock is used far more often than the defer. 
* code complexity: using the defer close to lock acquisition makes sure the lock is released, which seems elegant to me :-) I would argue the inline function doesn't need to return anything: it can just modify the c,g,t,h in a closure. But even that's a matter of taste.

metrics2/expvar/expvar.go
```diff
@@ -0,0 +1,280 @@
...
+// With implements Counter.
+func (c *IntCounter) With(keyvals ...string) metrics.Counter {
+	return &IntCounter{
+		parent:  c.parent,
+		name:    c.name,
+		keyvals: keyval.Merge(c.keyvals, keyvals...),
+	}
+}
+
+// Add inplements Counter.
```
nicolaiskogheim: s/inplements/implements

metrics2/internal/histogram/histogram.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"gopkg.in/caio/go-tdigest.v1"
+)
+
+// Histogram accumulates observations and provides quantile summaries of
+// observations. Note that it's implemented with a TDigest, which
```
nicolaiskogheim: Unfinished sentence.

peterbourgon:
The current interface in master absolutely allows Prometheus labels to work, via `With`.

cyx:
@peterbourgon is there anything we're missing here? https://gist.github.com/cyx/854e5fa0fdf963a99d95df87322a2d91

Maybe you mean not via the provider -- which I agree definitely works.

peterbourgon:
Ah, yes, not via the provider interface. It's not intended for general use:

> Package provider . . . is provided specifically for the needs of the NY Times framework Gizmo. Most normal Go kit users shouldn't need to use it.

This PR is in stasis. I would like to complete it, but I don't currently have time to carry it forward.

cyx:
If you have directions / ideas etc, we might be able to carry it forward -- up to you!

peterbourgon:
I think the patterns are pretty well established with the code as-is. Beyond the guidelines laid out in the Goals section up top...

- Each metrics2/foo package has a NewProvider constructor
- Each metrics2/foo.Provider type has NewCounter, Gauge, and Histogram methods
- Additional methods specific to the metrics backend may also be given (see statsd SendLoop/WriteTo)

For this much, it's just a grind, to work through all of the already-supported metrics backends. I expect the hard parts will be two things. First, making universal support for our concept of labels (essentially Prometheus idea of labels). What I've written for metrics2/statsd is a good example of doing it "from scratch" with a bit of novel engineering, but that's really just getting started. Look at what we had to do to support influxdata and influxstatsd in the existing package metrics. That will need to be re-implemented.

Second, properly testing each implementation. I've chosen to do full in-memory end-to-end tests, pushing random data through each implementation and checking the "output" for statistical accuracy within some tolerance. I think this is important, but it's definitely time-consuming.

Does this sound like something you'd be capable of contributing to, in whole or part?

cyx:
Yes definitely sounds like something! I think what we'll do for now, is copy/paste build what you have here in our own repos making librato and prometheus work as best as we can using the interfaces that have been defined so we're not build it in a vacuum to begin with.

Once we're happy with that we can come back and make a small-ish PR showing a small iteration towards that direction.

One question I have though is the provider interface you have here, and the differences in the signatures for what you have currently (e.g. prometheus does `NewCounter(namespace, subsystem, ...)` vs the expvar one only doing `NewCounter(name)`. Did you have more thoughts on that?

peterbourgon:
SGTM.

> One question I have though is the provider interface you have here, and the differences in the signatures for what you have currently (e.g. prometheus does NewCounter(namespace, subsystem, ...) vs the expvar one only doing NewCounter(name). Did you have more thoughts on that?

So, Go kit's package metrics makes a strong claim that _metric instrumentation in code_ can be reduced to the three types (Counter, Gauge, Histogram) and corresponding methods (Add, Set, Observe) and a common method for establishing cardinality (With). We can provide adapter code to massage these abstractions into each of the backends once the underlying types are constructed, but I'm not sure how to create abstractions for construction of the backend types themselves. If you have ideas, I'm all ears.

peterbourgon:
Well, there's one attempt. I'll need to implement a few more backends to see if it holds up...

cyx:
I think for starters, taking a set of label values will work for both librato + prometheus (e.g. making prometheus the common denominator).

Librato itself doesn't require that, but maybe that restriction is a good thing anyway.

So basically:

```
NewCounter(name string, labels ...string)
NewGauge(name string, labels ...string)
NewHistogram(name string, buckets int, labels ...string)
```

For starters at least.

The only part where we're left feeling like we need more really is the histogram implementations. A strawman idea might re-use the same idea with `With`:

```
h := NewHistogram("request-duration.ms").WithOption(SomeFuncOptionForBuckets(...))`
h.Observe(...)
```

That's just an off the cuff idea though! The declaration of labels in prom seem good. As for `NewCounter`, `NewGauge`, etc returning an error -- it seems like a lot of noise for most implementations. Maybe the original behavior of throwing panics was ok-ish for the prom case there?

cyx:
As for the templated solution, while it's very explicit and the approach is something I like, it might end up with divergence, e.g.

```
NewCounter("my-counter.{region}.something.{az}")
```

Would probably not make sense in a prometheus context? We recently implemented this in librato's v1 API (the one which doesn't support tags) and went with a convention approach (e.g. it gets flattened into `k1.v1:k2.v2` -- not saying that's ideal -- just that with the constraint of a provider interface it's something that would be forced.

peterbourgon:
@cyx Take a look at the [Provider abstraction](https://github.com/go-kit/kit/blob/metrics-rerefactor/metrics2/provider.go) I just introduced, tell me what you think.

peterbourgon:
@cyx Take a look at [the Provider abstraction](https://github.com/go-kit/kit/blob/metrics-rerefactor/metrics2/provider.go) I just introduced. I think it addresses your points...

peterbourgon:
@cyx Concretely: I'd love to see your attempt at a package metrics2/librato, using the Provider abstraction. I'd love to see if/how it breaks down.

cyx:
Cool I'll branch off this in my fork and draft a librato implementation. At first glance it seems to cover everything though.

One idea re: errors and the `NewXYZ` constructor: is that mostly because of Prometheus, and if so can we bypass the errors by checking for the existence of metrics ?

peterbourgon:
Constructing a new metric can fail if e.g. the identifier is a duplicate. This is explicit in Prometheus, yes, but it's also implicit in e.g. expvar. Since that's true, I think it's correct that e.g. NewCounter returns an error. What do you suggest as a replacement?

cyx:
I think an alternative is to simply have a map of all the names and prevent the call to the underlying registry, e.g.

```golang
if _, ok := f.cache[name]; !ok {
  f.cache[name] = expvar.NewFloat(name)
}

return f.cache[name]
```

A similar approach can be taken for prometheus. While it does add some surface area to the implementation, if the only error we expect for these providers are registry related ones, it seems like an ok tradeoff to have a func signature which doesn't have an error.

peterbourgon:
Eugh, I see. Let me think about that for a bit. It's not an obviously good decision, from my perspective.

cyx:
Nice thanks for trying it! One last bike shed moment is the metrics2 name. I assume we'll try to go with `v2/metrics` or something blessed by the new versioning scheme? (Mostly just curious what your plan is with this)

peterbourgon:
Oh yeah don't worry about that, we'll have some migration strategy TBD.

cyx:
I just finished an MVP sketch on an app optics provider (aka librato): https://github.com/go-kit/kit/compare/metrics-rerefactor...cyx:add-appoptics

All in all, I think the new provider structure will definitely allow us to do double prometheus + app optics posting (and potentially others -- e.g. l2met esque STDOUT for development).

peterbourgon:
Great, glad to hear the model works for you. I'll continue pushing forward. Would it be OK if I brought those changes in and made some adaptations?

cyx:
@peterbourgon sure whatever you think. Would be curious to see those adaptations as well 👍

cyx:
@peterbourgon do you have anything baking in your machine?

peterbourgon:
Yeah, but full disclosure: it's gonna be a slow trickle over the next weeks and months before we get anything usable in upstream. I've got too many commitments right now and don't have much time :(

cyx:
Ok if ever we need something you don't mind us using this branch and making a fork I suppose correct?

peterbourgon:
Please do.

PR #570, opened 2017-07-12
update to generic counter

travissalascox:
worked on taking a stab at reworking the implementation of the generic Metrics using [Space](https://github.com/go-kit/kit/blob/master/metrics/internal/lv/space.go#L13) from the lv internal package. before getting too far into that, decided to implement changes only in the generic counter and get @peterbourgon thoughts on it.

If this solution works, will update the other Metrics in a similar way. 

Tests have been added that address the exact problem raised in the issue and looks to be passing.

### DIFF
--- a/metrics/generic/generic.go
+++ b/metrics/generic/generic.go
@@ -18,60 +18,56 @@ import (
 
 // Counter is an in-memory implementation of a Counter.
 type Counter struct {
-	Name string
-	lvs  lv.LabelValues
-	bits uint64
+	Name  string
+	lvs   lv.LabelValues
+	space *lv.Space
 }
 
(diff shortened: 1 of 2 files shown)


travissalascox:
@peterbourgon just noticed you had a PR up for refactoring Metrics. If you want, can easily close one.

peterbourgon:
Yeah—thank you for the effort!! but I think this will be subsumed by the work I'm doing over there, which should be the major change for the 0.6.0 release.

travissalascox:
@peterbourgon NP, let me know if there is anything that i can help with on that front, would be more than happy to 😸 . If not, looking forward to seeing the new feature in `0.6.0`

PR #574, opened 2017-07-14
Add syslog support

mingan:
This is another crack at #333. It incorporates the discussion from #486 but it shares very little code with it. In structure, it's very similar to [term.NewColorLogger](https://godoc.org/github.com/go-kit/kit/log/term#NewColorLogger) - it's so similar that `loggerBuf` could be extracted and be used b both (but in which package should it live? it would have to be exported which might be confusing?).

However, while working on this I realized that it breaks on one the [12 factor](https://12factor.net/logs) app principles. Including it in Go kit might be a promotion of a suboptimal approach, so I'm not sure it should be included at all.


Gouthamve:
Hi, any updates on this?

log/syslog/syslog_logger.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	gosyslog "log/syslog"
+	"io"
+	"github.com/go-kit/kit/log/level"
+	"github.com/go-kit/kit/log"
+	"bytes"
+	"sync"
+)
+
+// PrioritySelector inspects list of keyvals and select a syslog priority
```
peterbourgon: > PrioritySelector inspects _the_ list of keyvals, and _selects_ a syslog priority.

log/syslog/example_test.go
```diff
@@ -0,0 +1,24 @@
+package syslog_test
+
+import (
+	gosyslog "log/syslog"
+	"github.com/go-kit/kit/log/syslog"
+	"github.com/go-kit/kit/log"
+	"fmt"
+	"github.com/go-kit/kit/log/level"
+)
```
peterbourgon: Can we do proper import grouping, i.e.

```go
import (
    // stdlib

    // go-kit
)
```

peterbourgon:
@ChrisHines Any blockers that you see?

log/syslog/syslog_logger.go
```diff
@@ -0,0 +1,124 @@
...
+		lb.logger = l.newLogger(lb.buf)
+	} else {
+		lb.buf.Reset()
+	}
+	return lb
+}
+
+func (l *syslogLogger) putLoggerBuf(cb *loggerBuf) {
+	l.bufPool.Put(cb)
+}
```
ChrisHines: - s/cb/lb to be consistent with variable names in `getLoggerBuf`.
- I feel like it would help readability to move the `type loggerBuf`, `func (l *syslogLogger) getLoggerBuf`, and `func (l *syslogLogger) putLoggerBuf` declarations to after the `func (l *syslogLogger) Log` declaration because they are couple together more tightly than the `Options` related code. I find myself having to skip over the `Options` stuff a lot while reading the code.

log/syslog/syslog_writer.go
```diff
@@ -0,0 +1,23 @@
...
+	Emerg(string) error
+	Alert(string) error
+	Crit(string) error
+	Err(string) error
+	Warning(string) error
+	Notice(string) error
+	Info(string) error
+	Debug(string) error
+}
+
```
ChrisHines: I would prefer all of the code in this file be combined with `syslog_logger.go`  and the combined file be renamed to `syslog.go`. I would put the `type SyslogWriter` declaration directly above the `func NewSyslogLogger` declaration.

peterbourgon:
Thanks everyone, and sorry again for that inexcusably long delay.

Issue #575, opened 2017-07-17
Add JSON RPC Transport


rossmcf:
I found a need for [JSON RPC](http://www.jsonrpc.org) on an upcoming project, and noticed that go-kit doesn't currently support this. I'd like to add this to the available transports.

I notice that an attempt was made at this earlier this year (PR #451). There's a PR on its way that continues @blainsmith's work.

PR #576, opened 2017-07-17
JSON RPC over HTTP

rossmcf:
This PR continues @blainsmith's work from #451 to address issue #575.

This is still a work in progress, but I wanted to open the PR early, to invite discussion.

I've taken the liberty of staring a new fork. @blainsmith's commits have been imported, so that he's duly credited.

I've added some tests, and hopefully addressed some of the feedback from the previous PR, namely creating an `RequestID` object with a custom `JSONUnmarshal`, that can handle IDs as strings, ints or floats. @groob, I'd appreciate your thoughts on that.

I'll try to tackle some more of the feedback, regarding server and encoding options in due course.

Thanks in advance for any feedback you can offer.

Ross


blainsmith:
Thanks for the props and also moving it forward! I got pulled away to do other things and never got a chance to move forward with it.

peterbourgon:
With the caveats that I don't know a whole lot about JSON RPC, and I've only just skimmed the code: the shape of this looks quite good to me. I'll have a few nits to pick in a final review, but otherwise 👍  from me so far.

basvanbeek:
I'll try to take a look too. I have used JSON-RPC extensively in the past.

peterbourgon:
@rossmcf Just ping me when you're ready for a final review! 👍

rossmcf:
Thanks @peterbourgon. I should hopefully have some time to tidy it up towards the end of the week.

rossmcf:
@peterbourgon I'm getting a little blind to this now, so would appreciate your input.

With regard to examples, would I be best to add a JSON RPC transport to the addsvc?

rossmcf:
Also, any pointers on why the travis build is failing would be useful. I can't make sense of the output.

rjeczalik:
> I can't make sense of the output.

```
--- FAIL: TestDefaultEndpointer (0.03s)
	endpointer_test.go:33: want 2, have 0
	endpointer_test.go:39: endpoint a closed, good
	endpointer_test.go:45: endpoint b closed, good
```

This test is racy.

[It expects](https://github.com/go-kit/kit/blob/62f8228976ae9c7a7e3c6e0eb33d1c3e854d5b00/sd/endpointer_test.go#L33) the cache to have certain elements, however [it's updated concurrently](https://github.com/go-kit/kit/blob/62f8228976ae9c7a7e3c6e0eb33d1c3e854d5b00/sd/endpointer.go#L38).

It's pretty easy to reproduce:

 ```
$ go get golang.org/x/tools/cmd/stress
$ go test -race -c .
$ stress ./sd.test 

/var/folders/dl/s7jq4vyn7358p1mtsmn3bk5w0000gn/T/go-stress749329897
--- FAIL: TestDefaultEndpointer (0.00s)
	endpointer_test.go:33: want 2, have 0
	endpointer_test.go:39: endpoint a closed, good
	endpointer_test.go:45: endpoint b closed, good
FAIL


ERROR: exit status 1

33 runs so far, 1 failures
```

**Comment by fredipevcin** [comment]
@rossmcf @peterbourgon What do you think about https://github.com/rossmcf/kit/compare/json-rpc-over-http...fredipevcin:json-rpc-over-http?expand=1

Idea is that **method** correlates to **service**.

For every **service** it is possible to specify **logger**, **before** and **after** instead of only for whole RPC handler.

```go
handler := jsonrpc.NewServer(
    jsonrpc.ServiceMap{
        "add": jsonrpc.NewService(
            func(ctx context.Context, request interface{}) (response interface{}, err error) {
                return nil, nil
            },
            func(ctx context.Context, msg json.RawMessage) (interface{}, error) {
                return nil, nil
            },
            func(ctx context.Context, result interface{}) (json.RawMessage, error) {
                return []byte(`"result"`), nil
            },
            jsonrpc.ServiceBefore(func(ctx context.Context, _ http.Header) context.Context {
                return ctx
            }),
        ),
    },
)
http.Handle("/v1/rpc", handler)
```

Service is not aware about `http.Request` nor `http.ResponseWritter` because `Server` takes care of this.

Code is just WIP as I wanted your option on it if it makes sense.

fredipevcin:
I'm currently working on JSON-RPC implementation https://github.com/fredipevcin/gokit-jsonrpc which we will use in-house. It is not yet completed.

peterbourgon:
@rossmcf Thanks and please forgive my delay! At a glance this seems to be alright. Can you help me review this more thoroughly (and help future users!) by adding a runnable example? I think the best thing would be to add JSON RPC support to examples/addsvc, but I'd also look at something standalone if there were good reasons that weren't feasible.

rossmcf:
@peterbourgon Thanks. I don't see any issues adding to examples/addsvc. I'm not sure exactly when I'll have time to do that, but hopefully in the next week or so.

rossmcf:
@peterbourgon One mythical man week later… here we are. I'm a bit short of satisfying coveralls, mostly down to the client, but the examples are in place now. I'd appreciate any feedback you can offer. Thanks.

transport/http/jsonrpc/server.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return s
+}
+
+// ServerOption sets an optional parameter for servers.
+type ServerOption func(*Server)
+
+// ServerBefore functions are executed on the HTTP request object before the
+func ServerBefore(before ...httptransport.RequestFunc) ServerOption {
+	// request is decoded.
```
peterbourgon: Whoops, wrong line! :)

transport/http/jsonrpc/server.go
```diff
@@ -0,0 +1,205 @@
...
+		s.logger.Log("err", rpcerr)
+		s.errorEncoder(ctx, rpcerr, w)
+		return
+	}
+
+	// Get the endpoint and codecs from the map using the method
+	// defined in the JSON  object
+	ecm, ok := s.ecm[req.Method]
+	if !ok {
+		err := methodNotFoundError(fmt.Sprintf("Method %s was not found.", req.Method))
```
peterbourgon: Oh, I see now how you're using the error types. Hmm, interesting. Disregard my previous comment.

transport/http/jsonrpc/server.go
```diff
@@ -0,0 +1,205 @@
...
+
+	e := Error{
+		Code:    InternalError,
+		Message: err.Error(),
+	}
+	if sc, ok := err.(ErrorCoder); ok {
+		e.Code = sc.ErrorCode()
+	}
+
+	w.WriteHeader(http.StatusOK)
```
peterbourgon: That's surprising, is this what the JSONRPC spec says?

transport/http/jsonrpc/server.go
```diff
@@ -0,0 +1,205 @@
...
+	json.NewEncoder(w).Encode(Response{
+		JSONRPC: Version,
+		Error:   &e,
+	})
+}
+
+// ErrorCoder is checked by DefaultErrorEncoder. If an error value implements
+// ErrorCoder, the Error will be used when encoding the error. By default,
+// InternalError (-32603) is used.
```
peterbourgon: I understand the difference between this and StatusCoder, but it took me a second. Maybe it's worth being a bit more explicit, e.g. "...the integer value will be used as the JSONRPC error code"?

transport/http/jsonrpc/client.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil, err
+		}
+		rpcReq := Request{
+			JSONRPC: "",
+			Method:  c.method,
+			Params:  params,
+			ID:      c.requestID(),
+		}
+
+		req, err := http.NewRequest("POST", c.tgt.String(), nil)
```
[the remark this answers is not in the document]
basvanbeek: JSON-RPC as spec is transport agnostic. Since the method type of the HTTP request does not signal anything in JSON-RPC's world, it makes sense to pick the most obvious and best generally suited method.

transport/http/jsonrpc/server.go
basvanbeek: A JSON-RPC request holds the statuscode internally and has no mapping for HTTP status codes. If we can return a proper JSON-RPC payload even if the payload is an error response the HTTP status OK is a good idea as it signals the request was properly handled even though the response is an error.

yurishkuro: This does make monitoring harder though, especially if it's done with infra like Envoy which might know about the HTTP status code but won't look into the JSON payload.

examples/addsvc/cmd/addsvc/addsvc.go
```diff
@@ -135,11 +136,12 @@ func main() {
...
-		endpoints    = addendpoint.New(service, logger, duration, tracer)
-		httpHandler  = addtransport.NewHTTPHandler(endpoints, tracer, logger)
-		grpcServer   = addtransport.NewGRPCServer(endpoints, tracer, logger)
-		thriftServer = addtransport.NewThriftServer(endpoints)
+		service        = addservice.New(logger, ints, chars)
+		endpoints      = addendpoint.New(service, logger, duration, tracer)
+		httpHandler    = addtransport.NewHTTPHandler(endpoints, tracer, logger)
+		grpcServer     = addtransport.NewGRPCServer(endpoints, tracer, logger)
+		thriftServer   = addtransport.NewThriftServer(endpoints)
+		jsonrpcHandler = addtransport.NewJSONRPCHandler(endpoints, logger)
```
basvanbeek: Would like to see tracing added to this transport too in the addsvc example just like HTTP and gRPC.

transport/http/jsonrpc/server.go
basvanbeek: yes but unfortunately it can break some clients which expect JSON-RPC responses to be transported over HTTP with a 2xx http status code. 

so we might need this to be configurable...

rossmcf: JSONRPC is, dare I say it, a bit SOAP-like in this regard, in that the HTTP status code only relates to the success of the HTTP request. Any app-level errors are encoded in the response body.

I'm not terribly keen on how this works either. It does make the HTTP response opaque to monitoring, but I feel it's what clients will expect.

transport/http/jsonrpc/server.go
rossmcf: I confused myself with this at times. Glad it makes sense to someone else!

examples/addsvc/cmd/addsvc/addsvc.go
rossmcf: I'll need to ponder how best to do this. As it stands, ServerBefore functions are called prior to decoding the request body. This will mean that we can only trace a request to the JSON RPC endpoint; we won't know which method was called, which would hide some useful detail.

I could move the ServerBefore calls to after decoding the JSON, but I'm not sure if/how this would affect other ServerBefore uses, as I'm not really familiar with the use cases for those.

I suppose the other option would be to introduce a tracing-specific option, but that feels a bit icky. Suggestions welcome!

basvanbeek: Not needed, I know how to plug into existing framework.. I'll add tracing independent from this PR.

rossmcf: Great. Thank you.

rossmcf:
Thanks for your feedback, everyone.

Test coverage in the package is looking more respectable now, and I think the issues raised have been addressed. Over to you.

transport/http/jsonrpc/client.go
```diff
@@ -0,0 +1,221 @@
...
+type AutoIncrementRequestID int
+
+// Generate satisfies RequestIDGenerator
+func (i *AutoIncrementRequestID) Generate() *RequestID {
+	id := *i
+	*i++
+	return &RequestID{
+		intValue: int(id),
+	}
+}
```
peterbourgon: This isn't goroutine-safe, and the type that includes it (the client) doesn't perform any synchronization. You've got a data race! You need to either put synchronization in the client, and un-export this type or explicitly mark it as not goroutine-safe; or put synchronization in this type directly.

transport/http/jsonrpc/server.go
peterbourgon: Still got a problem here.

peterbourgon:
Oh, and the test failure reports:

```
# github.com/go-kit/kit/transport/http/jsonrpc_test
transport/http/jsonrpc/client_test.go:119: Fatalf format %d has arg requestAtServer.ID of wrong type *github.com/go-kit/kit/transport/http/jsonrpc.RequestID
transport/http/jsonrpc/client_test.go:171: Fatalf format %d has arg rs of wrong type string
```

which I reckon is fixable, too.

transport/http/jsonrpc/client.go
rossmcf: OK, I think that's addressed now, with an atomic.AddUint64 in Generate().

rossmcf:
OK, coveralls is finally satisfied. Hopefully that's everything taken care of.

transport/http/jsonrpc/client_test.go
```diff
@@ -0,0 +1,226 @@
...
+	result, err := sut.Endpoint()(context.Background(), in)
+	if err != nil {
+		t.Fatal(err)
+	}
+	rs, ok := result.(string)
+	if !ok {
+		t.Fatalf("result is not string: (%T)%+v", result, result)
+	}
+	if rs != "boogaloo" {
+		t.Fatalf("want=boogaloo, got=%d", rs)
```
peterbourgon:
```
# github.com/go-kit/kit/transport/http/jsonrpc_test
transport/http/jsonrpc/client_test.go:190: Fatalf format %d has arg rs of wrong type string
FAIL	github.com/go-kit/kit/transport/http/jsonrpc [build failed]
```

peterbourgon:
Build failure due to an aggressive lint.

rossmcf:
@peterbourgon Thanks for spotting that. Oddly, it didn't come up when I ran `lint`. I've made a few more tweaks to quieten (although not completely silence) the linter.

peterbourgon:
Yes! Thank you so much! This is great. I'll cut a release this week, probably.

rossmcf:
Thank you for all of your support with this.

In the release notes, might you be able to mention the original contribution of @blainsmith, and my employer, @senseyeio, on whose time I did most of the work for this?

sas1024:
Any plans for new release?

PR #579, opened 2017-07-19
Fix confusing names

### LINKED ISSUE #526: auth/jwt: Function names not intuitive.
I've been using the auth/jwt package on a project for a while now and I constantly trip over the difference between `FromHTTPContext` and `ToHTTPContext`. The structure of the names don't help me remember which way the data is flowing.

From godoc:

----------

```go
func FromHTTPContext() http.RequestFunc
```
FromHTTPContext moves JWT token from context to request header. Particularly useful for clients.

```go
func ToHTTPContext() http.RequestFunc
```
ToHTTPContext moves JWT token from request header to context. Particularly useful for servers.

------------

I'd prefer the following renaming of these functions (and the corresponding gRPC pair):

`FromHTTPContext` -> `ContextToHTTP`
`ToHTTPContext` -> `HTTPToContext`

Do others find the current names confusing also?
Does anyone have better suggestions that mine?

peterbourgon:
Closes #526.

@ChrisHines @basvanbeek — interested in a review?

@willfaught @terinjokes — likewise, and FYI.

### DIFF
--- a/auth/jwt/transport.go
+++ b/auth/jwt/transport.go
@@ -17,7 +17,7 @@ const (
 	bearerFormat string = "Bearer %s"
 )
 
-// HTTPToContext moves JWT token from request header to context. Particularly
+// HTTPToContext moves a JWT from request header to context. Particularly
 // useful for servers.
 func HTTPToContext() http.RequestFunc {
 	return func(ctx context.Context, r *stdhttp.Request) context.Context {
@@ -30,7 +30,7 @@ func HTTPToContext() http.RequestFunc {
 	}
 }
 
-// ContextToHTTP moves JWT token from context to request header. Particularly
+// ContextToHTTP moves a JWT from context to request header. Particularly
 // useful for clients.
 func ContextToHTTP() http.RequestFunc {
 	return func(ctx context.Context, r *stdhttp.Request) context.Context {
@@ -42,7 +42,7 @@ func ContextToHTTP() http.RequestFunc {
 	}
 }
 
-// GRPCToContext moves JWT token from grpc metadata to context. Particularly
+// GRPCToContext moves a JWT from grpc metadata to context. Particularly
 // userful for servers.
 func GRPCToContext() grpc.ServerRequestFunc {
 	return func(ctx context.Context, md metadata.MD) context.Context {
@@ -61,7 +61,7 @@ func GRPCToContext() grpc.ServerRequestFunc {
 	}
 }
 
-// ContextToGRPC moves JWT token from context to grpc metadata. Particularly
+// ContextToGRPC moves a JWT from context to grpc metadata. Particularly
 // useful for clients.
 func ContextToGRPC() grpc.ClientRequestFunc {
 	return func(ctx context.Context, md *metadata.MD) context.Context {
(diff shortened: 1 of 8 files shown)


peterbourgon:
For the record,

```fish
#!/usr/bin/env fish

for pair in \
    "FromHTTPContext ContextToHTTP" \
    "ToHTTPContext   HTTPToContext" \
    "FromGRPCContext ContextToGRPC" \
    "ToGRPCContext   GRPCToContext" \
    "FromHTTPRequest HTTPToContext" \
    "ToHTTPRequest   ContextToHTTP" \
    "FromGRPCRequest GRPCToContext" \
    "ToGRPCRequest   ContextToGRPC"
  set fr (echo $pair | awk '{print $1}')
  set to (echo $pair | awk '{print $2}')
  echo change from $fr to $to
  find . -type f -name '*.go' -exec sed -i '' s/$fr/$to/ \{\} +
end
```

auth/jwt/transport.go
```diff
@@ -17,9 +17,9 @@ const (
 	bearerFormat string = "Bearer %s"
 )
 
-// ToHTTPContext moves JWT token from request header to context. Particularly
+// HTTPToContext moves JWT token from request header to context. Particularly
```
ChrisHines: I feel obligated to point out that JWT is an acronym for JSON Web Token, which makes the phrase "JWT token" repetitive. I believe a more grammatically correct phrasing is "... moves a JWT from request header to context."

peterbourgon: Fixed. Thank you.

auth/jwt/transport.go
```diff
@@ -17,9 +17,9 @@ const (
 	bearerFormat string = "Bearer %s"
 )
 
-// ToHTTPContext moves JWT token from request header to context. Particularly
+// HTTPToContext moves a JWT from request header to context. Particularly
```
willfaught: Seems like you'd want to keep the "a" article use consistent—either use them everywhere or nowhere ("a request header", "a context").

auth/jwt/transport.go
```diff
@@ -17,9 +17,9 @@ const (
...
 // useful for servers.
-func ToHTTPContext() http.RequestFunc {
+func HTTPToContext() http.RequestFunc {
```
willfaught: Just curious, and this isn't part of the review for this PR, why have ToHTTPContext/HTTPToContext return the RequestFunc instead of be the RequestFunc? It always returns the same function value with no side effects.

auth/jwt/transport.go
willfaught: I see, it's for consistency with the others. Learning as I go. :)

willfaught: The "a" article use is now inconsistent (present with "JWT", absent with "request header" and "context"), not sure if that's intentional, not a big deal.

auth/jwt/transport.go
willfaught: "Context" seems to be in all of these names, so is it redundant? Would To/FromHTTP and To/FromGRPC be equally clear? I haven't taken a close look at the whole file, and I've never used the package, so I'm genuinely asking. :) If not, the names look good!

Issue #580, opened 2017-07-20
JSONLogger let's error interface take priority over json.Marshaler


cam-stitt:
From what I can see, if you have an object that implements both `json.Marshaler` and `error`, it will always use the `error` interface. This is despite the fact that code states that `json.Marshaler` should always win [here](https://github.com/go-kit/kit/blob/master/log/json_logger.go#L51). Consider this (I know it's not perfect, but we have a stack trace in our error that we want logged):

```go
type Error struct {
  Message string
  Status int
}

func (e Error) Error() string {
  return e.Message
}

func (e Error) MarshalJSON() ([]byte, error) {
  aux := struct{Message string, Status Int}{Message: e.Message, Status: e.Status}
  return json.Marshal(aux)
}
```
And you were to log it:

```go
err := Error{msg: "Error impl", Status: 200}
logger.Log("msg", err)
```

The result will always be `{"err": "Error impl"}`

cam-stitt:
From what I can see, it's because of this: https://github.com/go-kit/kit/blob/master/log/json_logger.go#L48

ChrisHines:
Yes. I reproduced this bug and agree with your analysis. Would you like to submit a PR with a reproducing test case and a fix? I believe the fix is just removing the if statement you linked to above.

ChrisHines:
Fixed by #581.

PR #581, opened 2017-07-20
Remove unnecessary error check.

cam-stitt:
#580

If you want to see the test fail, just add the `error` interface in the test without removing the error type check.

gokit-logger


ChrisHines:
Perfect. Thanks!

Fixes #580.

PR #582, opened 2017-07-21
More generic API to work with x/time/rate

nelz9999:
As per the discussion in Slack (https://gophers.slack.com/archives/C04S3T99A/p1500582764076547), I added the ability to use the `x/time/rate` package for the rate limiting implementation, while also keeping backwards-compatible with the `juju/ratelimit` usage.

I'm pretty pleased with the structure, using subset interfaces off the `x/time/rate.Limiter` to express most of what we needed.

I was even able to port the older usage of `NewTokenBucketLimiter` to use the new implementation under the covers, but was unable to do so with `NewTokenBucketThrottler` (because IMO the API leaked testing configuration).

For any future API-breaking versions, it wouldn't be too hard to make the `juju`-related adapters a sub-package, but I'm not pushing that in this PR.

Would love to hear solid opinions on improvements I could make to the names.

### DIFF
--- a/ratelimit/token_bucket.go
+++ b/ratelimit/token_bucket.go
@@ -23,16 +23,11 @@ func NewTokenBucketLimiter(tb *ratelimit.Bucket) endpoint.Middleware {
 
 // NewTokenBucketThrottler returns an endpoint.Middleware that acts as a
 // request throttler based on a token-bucket algorithm. Requests that would
-// exceed the maximum request rate are delayed via the parameterized sleep
-// function. By default you may pass time.Sleep.
-func NewTokenBucketThrottler(tb *ratelimit.Bucket, sleep func(time.Duration)) endpoint.Middleware {
-	// return NewDelayingLimiter(NewWaiter(tb))
-	return func(next endpoint.Endpoint) endpoint.Endpoint {
-		return func(ctx context.Context, request interface{}) (interface{}, error) {
-			sleep(tb.Take(1))
-			return next(ctx, request)
-		}
-	}
+// exceed the maximum request rate are delayed.
+// The parameterized function "_" is kept for backwards-compatiblity of
+// the API, but it is no longer used for anything. You may pass it nil.
+func NewTokenBucketThrottler(tb *ratelimit.Bucket, _ func(time.Duration)) endpoint.Middleware {
+	return NewDelayingLimiter(NewWaiter(tb))
 }
 
 // Allower dictates whether or not a request is acceptable to run.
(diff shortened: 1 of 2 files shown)


dsudia:
@nelz9999 this looks awesome to me and should fulfill our license limitations while also fulfilling Peter's API request. Thank you!

blainsmith:
Very nice man.

ratelimit/token_bucket_test.go
```diff
@@ -3,13 +3,15 @@ package ratelimit_test
...
 	"math"
+	"strings"
 	"testing"
 	"time"
 
 	jujuratelimit "github.com/juju/ratelimit"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/ratelimit"
+	"golang.org/x/time/rate"
```
peterbourgon: Can you move this import decl up with juju, above?

ratelimit/token_bucket.go
```diff
@@ -18,25 +18,100 @@ var ErrLimited = errors.New("rate limit exceeded")
...
 func NewTokenBucketLimiter(tb *ratelimit.Bucket) endpoint.Middleware {
+	return NewErroringLimiter(NewAllower(tb))
+}
+
+// NewTokenBucketThrottler returns an endpoint.Middleware that acts as a
+// request throttler based on a token-bucket algorithm. Requests that would
+// exceed the maximum request rate are delayed via the parameterized sleep
+// function. By default you may pass time.Sleep.
+func NewTokenBucketThrottler(tb *ratelimit.Bucket, sleep func(time.Duration)) endpoint.Middleware {
+	// return NewDelayingLimiter(NewWaiter(tb))
```
peterbourgon: Is there a reason this didn't work? Would be cool if we could express all the old functions in terms of the new ones!

ratelimit/token_bucket_test.go
nelz9999: Will do

ratelimit/token_bucket.go
nelz9999: I can do it, but then the `sleep` parameter gets ignored. I'll check in that change, so you can evaluate if you like it.

nelz9999:
@peterbourgon This is ready for you to see if you like what I've done to marry the old functions to the new ones.

peterbourgon:
Yep, I'm happy with this for the current point release. For the next one I think we can totally drop Juju. Thanks so much for the work! 👍

peterbourgon:
Urgh, build failures are unrelated: see https://github.com/openzipkin/zipkin-go-opentracing/issues/68

basvanbeek:
Build failures due to Thrift have been fixed in zipkin-go-opentracing

Go kit also has issue with thrift update, PR #585 resolves this.

nelz9999:
@peterbourgon Now that 0.6.0 has been cut, would you like the follow-on PR to remove `juju`?

peterbourgon:
@nelz9999 Sure.

PR #583, opened 2017-07-21
sd: add Stop method to Instancer interface

### LINKED ISSUE #566: Add Stop() to Instancer interface
Stop() func is not part of Instancer interface which makes it impossible to stop instancers in a generic way.

peterbourgon:
Every implementation (modulo mock/test impls) already provided this method, and so it makes sense to lift it to the interface definition.

Closes #566.

### DIFF
--- a/sd/consul/instancer_test.go
+++ b/sd/consul/instancer_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/go-kit/kit/sd"
 )
 
-var _ sd.Instancer = &Instancer{} // API check
+var _ sd.Instancer = (*Instancer)(nil) // API check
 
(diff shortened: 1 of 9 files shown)


yurishkuro:
as mentioned in the issue, it's a code smell, but not worth fighting over. LGTM.

I would recommend a clear spec of the expected behavior of Stop() in the comments, e.g. "instancer guarantees no more updates will be sent to subscribers once the Stop() function returns".

side note - I personally prefer using io.Closer signature for such methods.

basvanbeek:
I'm ok with it. I have no preference for io.Closer over Stop.

peterbourgon:
I think I prefer `func Stop()` to `func Close() error` as the semantics are rather different.

PR #589, opened 2017-07-31
kitgen implementation

nyarly:
So far what I've got is simple generation based on a basic interface definition.

Not complete: different layout engines.

Specific design decision here was to construct the generated AST and then
format it out, as opposed to using a text/template. There's some things that
are made much easier this way (i.e. the output is more likely to be correct
code), but the generation code is a bit more opaque. In working with this
process I've found a few things that make the process clearer, and I'm looking
to implement those before this is merged. Specifically, I think many of the
repeated structures could be snagged from an example codefile and then
processed into target structures, rather than built out of whole cloth or
partial inline parses.


nyarly:
I'll be proceeding with this, but I wanted to get a little oversight to see if it's in a useful vien, @peterbourgon

peterbourgon:
Ah! I missed the notification! This is great. Let me dig in later today and I'll give some more specific comments...

peterbourgon:
Yeah, this is badass. Pretty much exactly what I had in mind, modulo the parameterization that you've already acknowledged. I also noticed it panics when I don't name the parameters in my service definition, i.e.

```go
type Service interface {
    Foo(int, int) (int, error) // panic
    Bar(i, j int) (int, error)   // OK
}
```

If you have enough to move forward, great! If not, please let me know how I can help: clarifying, requirements-gathering, etc.

peterbourgon:
Hehe, fun!

```
$ cat test.go
package main

import "context"

type Service interface {
	Foo(_ context.Context, _ int) (int, error)
}

$ kitgen test.go
 ...
func (s stubService) Foo(ctx context.Context, _ int) (int, error) {
        return "", errors.New("not implemented")
}

type FooRequest struct {
        _ int
}
type FooResponse struct {
        I   int
        Err error
}

func makeFooEndpoint(s stubService) endpoint.Endpoint {
        return func(ctx context.Context, request interface{}) (interface{}, error) {
                req := request.(FooRequest)
                i, err := s.Foo(ctx, req._)
                return FooResponse{I: i, Err: err}, nil
        }
}
 ...
```

nyarly:
I'm good. Just wanted to make sure I wasn't running down a rathole. :)

The edge cases are great. Thanks!

nyarly:
Updates: covers error cases pointed out already, also produces code that `go build`s. Of note: all the import statements are separate, because it turns out to be tricky to get generated ASTs to have recognized parenthesized imports. If that's a dealbreaker, I can dig into figuring out how to fix it.

peterbourgon:
No worries on the imports. We can file an issue later. Any blockers on the templated output? (No pressure at all!)

nyarly:
No blockers. Going to work on it tonight.

peterbourgon:
I just want to say I'm following along with each commit with great anticipation :)

nyarly:
This is nearly ready for review. What's the feeling here about rebasing?

cmd/kitgen/interface.go
```diff
@@ -0,0 +1,73 @@
...
+
+		handleCalls = append(handleCalls, &ast.ExprStmt{X: handleCall})
+	}
+
+	pasteStmts(handlerFn.Body, 1, handleCalls)
+
+	return handlerFn
+}
+
+func (i iface) reciever() *ast.Field {
```
annismckenzie: Typo in the function name? Should probably be `receiver()`?

peterbourgon:
> This is nearly ready for review. What's the feeling here about rebasing?

Nice! I'm not fussed, to be honest. It looks like your commits were more of the work-in-progress type, as opposed to grouping commits by module or functionality, so I'd slightly prefer they be squashed. But if you're more comfortable with everything as-is, please leave it, no worries.

nyarly:
That's everything I *couldn't not* do.

I'm reasonably happy with this, as is, but intently seeking feedback now. I *really* like how the tests can do a test `go build` of the results and the input templates - IMO that's the best bit, and the core argument for doing this as AST manipulation instead of basic string templating.

In the larger scope, I think there's maybe room for something like `replaceStructLit` (the smell here is anywhere that there's a type assertion chain to extract/replace an AST node.) I come to think that there's definitely room for a successor to the stdlib ast.Walk() with a more complex interface for Visitor.

peterbourgon:
Looks great! My weekend is totally slammed, but I'll find some time next week to give this a proper review. Thanks so much!! 🎉

peterbourgon:
```diff
diff --git a/cmd/kitgen/ast_helpers.go b/cmd/kitgen/ast_helpers.go
index c588213..f42daf2 100644
--- a/cmd/kitgen/ast_helpers.go
+++ b/cmd/kitgen/ast_helpers.go
@@ -39,6 +39,8 @@ func baseName(t ast.Expr) string {
 	switch tt := t.(type) {
 	default:
 		panic(fmt.Sprintf("don't know how to choose a base name for %T (%[1]v)", tt))
+	case *ast.ArrayType:
+		return "slice"
 	case *ast.Ident:
 		return tt.Name
 	case *ast.SelectorExpr:
diff --git a/cmd/kitgen/flatlayout.go b/cmd/kitgen/flatlayout.go
index 079dc74..d385b61 100644
--- a/cmd/kitgen/flatlayout.go
+++ b/cmd/kitgen/flatlayout.go
@@ -1,5 +1,7 @@
 package main
 
+import "go/ast"
+
 type flat struct{}
 
 func (f flat) transformAST(ctx *sourceContext) (files, error) {
diff --git a/cmd/kitgen/main.go b/cmd/kitgen/main.go
index c05819e..fdfd1fb 100644
--- a/cmd/kitgen/main.go
+++ b/cmd/kitgen/main.go
@@ -71,7 +71,7 @@ func main() {
 	if len(os.Args) < 2 {
 		log.Fatal(usage())
 	}
-	filename := os.Args[1]
+	filename := flag.Arg(0)
 	file, err := os.Open(filename)
 	if err != nil {
 		log.Fatalf("error while opening %q: %v", filename, err)
```

```
package main

type Service interface {
        PostProfile(ctx context.Context, p Profile) error
        GetProfile(ctx context.Context, id string) (Profile, error)
        PutProfile(ctx context.Context, id string, p Profile) error
        PatchProfile(ctx context.Context, id string, p Profile) error
        DeleteProfile(ctx context.Context, id string) error
        GetAddresses(ctx context.Context, profileID string) ([]Address, error)
        GetAddress(ctx context.Context, profileID string, addressID string) (Address, error)
        PostAddress(ctx context.Context, profileID string, a Address) error
        DeleteAddress(ctx context.Context, profileID string, addressID string) error
}

type Profile struct {
        ID        string    `json:"id"`
        Name      string    `json:"name,omitempty"`
        Addresses []Address `json:"addresses,omitempty"`
}

type Address struct {
        ID       string `json:"id"`
        Location string `json:"location,omitempty"`
}
```

```sh
$ kitgen -repo-layout default -target-dir default in.go
$ go build ./default/...
# github.com/go-kit/kit/cmd/kitgen/testdata/profilesvc/default/service
default/service/service.go:10:53: undefined: Profile
default/service/service.go:13:62: undefined: Profile
default/service/service.go:16:63: undefined: Profile
default/service/service.go:19:65: undefined: Profile
default/service/service.go:25:73: undefined: Address
default/service/service.go:28:87: undefined: Address
default/service/service.go:31:71: undefined: Address
# _/Users/peter/src/github.com/go-kit/kit/cmd/kitgen/testdata/profilesvc/default/service
default/service/service.go:10:53: undefined: Profile
default/service/service.go:13:62: undefined: Profile
default/service/service.go:16:63: undefined: Profile
default/service/service.go:19:65: undefined: Profile
default/service/service.go:25:73: undefined: Address
default/service/service.go:28:87: undefined: Address
default/service/service.go:31:71: undefined: Address
```

nyarly:
Confirming the design here: the Profile and Address structs should basically be copied into the output somewhere, because the "in.go" file doesn't itself make it through - the interface there is a guide for the generator, not an actual target artifact.

nyarly:
Reflecting on this, it seems like, as a user of kitgen, you'd like to write your interface and have it at least syntax check - ideally you should be able to `go build .` it to check for errors. To do that, all of the types in the interface need to be defined.

Therefore: you should be able to define types in the input file and have them copied verbatim into the output, somewhere. Any type declaration should be allowed for this purpose: new structs or types based on  basic types.

But, the contract of kitgen is (currently) that interface type definitions are used to describe the service itself. In general, it probably doesn't make sense for a service definition interface to receive or return interfaces, I think. Maybe there's a use case I don't see, within a larger application perhaps?

So, I think one of the following three things must be true (regardless of whether kitgen actually enforces it...)

1. There may only be one interface defined in the input file.
2. None of the methods of any of the interfaces may receive or return an interface type.
3. Interfaces received or returned by interface methods are "consumed" and are ignored for purposes of service generation - it is an error for all defined interfaces to be so consumed.

Those options increase in flexibility but also in complexity, natch. I suppose there's another option along the lines of "just copy everything that's not an interface type declaration and let the chips fall as they may." It strikes me as unergonomic, but it may not be as bad in practice as I imagine.

What do you think?

peterbourgon:
I like having the interface definition hanging around in the service somewhere, if for no other reason than to act as a point of documentation. So I think the ideal solution here would be to take the input service interface, somehow walk it for all referenced types (recursively), and copy that AST into the generated code. An equally-acceptable and probably-easier solution would be to require that the in.go file parse and build correctly, and just copy its content wholesale into the generated code.

nyarly:
I think we'll need to do the AST thing for the default layout to work properly - of necessity the types will need to go *somewhere* and offhand I think they'll need to be referenced in at least two packages of the default layout, so at a bare minimum we'll need to know the types exist and how to package-ify their names.

"Include all the types" does simplify things though. Gimme a bit...

annismckenzie:
Sooooooo, I've followed this for quite some time. I'm determined to help if you need any. I have a few services (in production) that would greatly benefit from having a stable and extensible foundation – that is, go-kit. 😬  Do you want me to take it for another spin tomorrow and report back? I could help with documentation (if and when I figure out the exact scope you're pushing towards in an initial release)?

nyarly:
@peterbourgon Most recent commit works with the profilesvc - it doesn't copy the interfaces across, since it complicates the naming game to do that (the output structs have the same name - I suppose we could do a Java-style -IFace suffix or something?) Otherwise, as we discussed.

Interesting wrinkle: profilesvc exposed a weakness in modifying the ASTs in place, because there were nodes that got reused, so replacewalk.go winds up producing a new AST in an almost FP kind of a way. I'm happier with this approach as it goes along - extracting it into some kind of toolkit seems like a better and better idea.

peterbourgon:
I'm very happy with this now. I could pick some nits about stuff like import grouping (as you noted) or stuff that prevents a clean `golint` run, like TODO comment blocks on exported identifiers, but I think it's a wash for a v1. @nyarly, thank you _so much_! Is there anything you'd like to do before this gets merged?

nyarly:
I'm happy with it as is. Thrilled to see it merged, and you're very
welcome. Thanks for the interesting problem!

peterbourgon:
Could you rebase on master, to fix the build (I hope)?

nyarly:
As soon as I get back to civilization - next Monday

peterbourgon:
Friendly ping :)

nyarly:
Okay - there's the rebase :crossed_fingers:

peterbourgon:
Woohoo!

zkanda:
I've been trying it out and it works great, thank you @nyarly

Issue #590, opened 2017-07-31
Cloudwatch metrics are accumulating on every Send() call, but should be Reset


feliksik:
On a call to cloudwatch.Send(), the state of the counter is not reset. 
That means the value is always increasing. 

This doesn't make sense to me, as Cloudwatch accumulates the values for you. It seems like we are basically implementing a counter with a Gauge-like behavior now. 

Can we agree this is a bug?

feliksik:
See PR #591 for a failing test case.

feliksik:
I'm also not sure how the Histogram values are supposed to be reset. 

Furthermore, note that if we move the test to teststat.TestCounter(), more (but not all) services fail. The still-succesful tests show which implementations use Reset().Walk(). 

```
$ go test ./metrics/...
ok  	github.com/go-kit/kit/metrics	0.253s
--- FAIL: TestCounter (0.00s)
	cloudwatch_test.go:81: Using counter second time: want 4641.000000, have 5978.000000
FAIL
FAIL	github.com/go-kit/kit/metrics/cloudwatch	0.086s
?   	github.com/go-kit/kit/metrics/discard	[no test files]
ok  	github.com/go-kit/kit/metrics/dogstatsd	0.761s
--- FAIL: TestCounter (0.00s)
	expvar_test.go:14: Using counter second time: want 4641.000000, have 5978.000000
FAIL
FAIL	github.com/go-kit/kit/metrics/expvar	0.077s
--- FAIL: TestCounter (0.00s)
	generic_test.go:25: Using counter second time: want 4641.000000, have 5978.000000
FAIL
FAIL	github.com/go-kit/kit/metrics/generic	0.680s
ok  	github.com/go-kit/kit/metrics/graphite	0.051s
ok  	github.com/go-kit/kit/metrics/influx	0.122s
ok  	github.com/go-kit/kit/metrics/internal/lv	0.002s
?   	github.com/go-kit/kit/metrics/internal/ratemap	[no test files]
ok  	github.com/go-kit/kit/metrics/multi	0.012s
--- FAIL: TestCounter (0.00s)
	pcp_test.go:26: Using counter second time: want 4641.000000, have 5978.000000
FAIL
FAIL	github.com/go-kit/kit/metrics/pcp	1.606s
--- FAIL: TestCounter (0.00s)
	prometheus_test.go:47: Using counter second time: want 4641.000000, have 5978.000000
--- FAIL: TestHistogram (0.02s)
[...]

Issue #578, opened 2017-07-18
Dogstatsd and statsd `func (g *Gauge) Add()` produces incorrect values across flushes


jfhamlin:
Both `Add()` implementations update observations using [`lv.node`'s `add`](https://github.com/go-kit/kit/blob/master/metrics/internal/lv/space.go#L101), but the space is reset on every flush:
- https://github.com/go-kit/kit/blob/master/metrics/statsd/statsd.go#L131
- https://github.com/go-kit/kit/blob/master/metrics/dogstatsd/dogstatsd.go#L139

Following the documentation quoted below, a negative aggregate delta over a flush period is interpreted by statsd as a delta when sent, whereas a positive aggregate delta, as sent, is interpreted as an absolute value.

For example, assume we call `gauge.Add(1)` once per flush period, with no other calls to `Add` or `Set` for that gauge. The resulting metric will have a constant value of 1; the additions will not accumulate.

These gauges either need to track and send the absolute value (noting the caveat for negative values quoted below) or include the sign character on all flushed values to denote deltas.

https://github.com/etsy/statsd/blob/master/docs/metric_types.md#gauges:
> Adding a sign to the gauge value will change the value, rather than setting it.
>
>`gaugor:-10|g`
>`gaugor:+4|g`
So if gaugor was 333, those commands would set it to 333 - 10 + 4, or 327.
>
>Note:
>
>This implies you can't explicitly set a gauge to a negative number without first setting it to zero.

peterbourgon:
The easier path is to take the sign option.

jfhamlin:
From DataDog support:

> So I have confirmed that we do not support gauge deltas.

So the dogstatsd implementation of `Gauge` will have to track the gauge value. I'm happy to work on a fix for this, but I don't have a firm ETA.

pbitty:
Looks like this was fixed in #588.

peterbourgon:
🙇

PR #591, opened 2017-07-31
Make sure Cloudwatch-metrics only send the *new* counter values. 

feliksik:
Testcase and fix for issue #590 . 

The commits are separated and documented. 

I only implemented the failing test in cloudwatch_test.go, instead of in teststat.TestCounter(), as it's not entirely clear to me if we can generalize the expected behavior for all methods. In particular, as the Send() method for e.g. Cloudwatch/Influx/Statsdogd is not implementing a generic interface, the semantics for each such method, and whether it should reset counters, may be defined differently (see #578 for an example of funkyness). 

todos: 

[ ] determine if this is actually a bug (I believe so)
[ ] determine whether the failing test case should be defined in the cloudwatch_test.go or in the teststat.TestCounter() method. 
[ ] fix the functionality in cloudwatch.go and other affected pieces.

### DIFF
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -2,6 +2,7 @@ package cloudwatch
 
 import (
 	"fmt"
+	"os"
 	"sync"
 	"time"
 
@@ -12,11 +13,17 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/generic"
+	"github.com/go-kit/kit/metrics/internal/lv"
 )
 
 const (
 	maxConcurrentRequests = 20
 )
+
+type Percentiles []struct {
+	s string
+	f float64
+}
 
 // CloudWatch receives metrics observations and forwards them to CloudWatch.
 // Create a CloudWatch object, use it to create metrics, and pass those metrics as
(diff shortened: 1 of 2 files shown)


metrics/cloudwatch/cloudwatch.go
```diff
@@ -12,81 +13,131 @@ import (
...
+	gauges     *lv.Space
+	histograms *lv.Space
+	*cwoptions
+}
+
+type cwoptions struct {
+	percentiles           Percentiles
 	logger                log.Logger
+	numConcurrentRequests int
+}
```
peterbourgon: Why is this a separate struct embedded into the type? I'd prefer to see these things flattened into the parent.

metrics/cloudwatch/cloudwatch.go
```diff
@@ -12,81 +13,131 @@ import (
...
+		validated := Percentiles{}
+		for _, entry := range p {
+			if entry.f < 0 || entry.f > 1 {
+				continue // illegal entry
+			}
+			validated = append(validated, entry)
+		}
+		o.percentiles = validated
+	}
+}
```
peterbourgon: No other metrics package has user-modifiable percentiles. Why is it important here?

metrics/cloudwatch/cloudwatch.go
feliksik: I see that; I will change it accordingly.

metrics/cloudwatch/cloudwatch.go
feliksik: The rationale is that in the way it's used here, you have an extra metric for each percentile. That costs $50 cents/month. If you want only the 50th and 95th percentile, you'd pay twice as much when the 90th and 99th are logged, too. 

For many metrics, this makes a difference. I hope that fact that it's configurable, makes this unobtrusive?

feliksik:
@cam-stitt do you have any comments on this PR? 
Would you care if the postfix for the percentiles are renamed from "50" to "p50"  (I think it's customary to use a p for percentile values, but if anybody objects, I won't care that much).

cam-stitt:
I'm not sure what other metric packages in go-kit use the "p" approach, but if @peterbourgon is okay with it, it doesn't bother me. It would probably cause anyone that has been using this package to adjust their dashboards/alerts accordingly though.

feliksik:
Graphite seems to use a p prefix; influx to, but created a separate tag, without the metric name. 

The others I don't understand the semantics of the prepared messages :-)

Issue #503, opened 2017-03-21
Feature Request: Add function to set key value for level logger


jpudney:
Yo! Recently I needed to change the value of the level key to ensure that our current log storage could slurp up the logs of a new service that's been developed with go-kit/kit/log as it's logger.

I've committed the [changes](https://github.com/jpudney/kit/commit/bff888cf052b86a5a7849a90a6738c3825634c7c) to a fork and wondered if you wanted them?

ChrisHines:
Short answer: No, I don't think we want to provide a global `level.SetKey` function.

The commit you linked says:

>The current key value "level" is hard coded to ensure that no unecessary
allocations are made following the instantiation of the logger. However,
there are times when this value needs to be somethin else. This changes
provides a method for doing so.

That is not the whole truth. Avoiding an allocation is a nice benefit of having a global key value, but it was not what drove the design.

The primary reason there is no `level.SetKey` function is because it opens the door for a data race if it is called concurrently with `level.Error` or the like. We could protect the key value with a mutex, but that could then become an unnecessary bottleneck imposed on all leveled log calls for all package clients.

The inability to change the key used by `level.Error` and its siblings was a tradeoff we accepted in order to allow leveled logging to build cleanly on top of the core log package without introducing a new `level.Levels` type to hold configuration information as the deprecated_levels package does. We found over time that the approach taken in deprecated_levels forced client code to make a choice between accepting a `log.Logger` or a `levels.Levels`. This dichotomy resulted in impedance mismatches between code on either side of this divide that we did not like.

Given the above, I believe the best way to support a custom level key without maintaining a fork of the level package, is to add use a `log.Logger` decorator that changes the level key prior to the `keyvals` being formatted for output. For example:

```go

func main() {
	l := log.NewLogfmtLogger(os.Stdout)
	l = setLevelKey(l, "myLevelKey")
	l = level.NewFilter(l, level.AllowInfo())

	level.Info(l).Log("foo", "bar")
}


func setLevelKey(logger log.Logger, key interface{}) log.Logger {
	return log.LoggerFunc(func(keyvals ...interface{}) error {
		for i := 1; i < len(keyvals); i += 2 {
			if _, ok := keyvals[i].(level.Value); ok {
				// overwriting the key without copying keyvals
				// techically violates the log.Logger contract
				// but is safe in this context because none
				// of the loggers in this program retain a reference
				// to keyvals
				keyvals[i-1] = key
				break
			}
		}
[...]

PR #592, opened 2017-07-31
Add SetKey() to log/level package to overwrite default key

metalmatze:
The default `level` key is perfect for most cases. Sadly GCP's logging doesn't parse the `level` key. Instead they use `severity`. I've seen a few workarounds but would really like to see this in go-kit itself, as I need to incorporate the workaround in a dozen projects.

I've added a simple `func SetKey(k interface{}) { key = k }` to the level package. Tests are included too.

Example `main.go`:
```go
func main() {
	logger := log.NewJSONLogger(log.NewSyncWriter(os.Stdout))
	logger = log.With(logger,
		"ts", log.DefaultTimestampUTC,
		"caller", log.DefaultCaller,
	)

	level.SetKey("severity")
	logger = level.NewFilter(logger, level.AllowDebug())
}
```
Before using the level package **set the key once** and then **don't change it afterwards**.

I'm not a 100% sure about the implications of this change. One could possibly change the key from multiple goroutines at the same time.

Without this change I feel like using the level package doesn't have an actual impact, because log levels aren't parsed on GCP.

### DIFF
--- a/log/level/level.go
+++ b/log/level/level.go
@@ -163,6 +163,11 @@ type Value interface {
 // package.
 func Key() interface{} { return key }
 
+// SetKey sets the unique key added to log events by the loggers in this
+// package. This is useful for situations where log aggregation
+// uses e.g. severity and not level.
+func SetKey(k interface{}) { key = k }
+
 // ErrorValue returns the unique value added to log events by Error.
 func ErrorValue() Value { return errorValue }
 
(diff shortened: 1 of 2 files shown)


ChrisHines:
Thanks for the PR, but we're not going add this feature. See the discussion in #503 for a full explanation.

PR #594, opened 2017-08-01
Use separate dimension/label for percentile

feliksik:
Currently for a Cloudwatch histogram, the percentile-values (50,90,95,99) are appended after the metrics name. 

This looks like: https://files.slack.com/files-pri/T029RQSE6-F6GHUGMNW/pasted_image_at_2017_08_01_06_09_pm.png

This makes is harder to filter for all the 50th percentiles of your metrics; In the current situation, if you plot all the 50th percentiles of metrics A, B and C, you will see the lines annotated with A_50, B_50, C_50. 
If you show all percentiles for A, you will see A_50, A_90, A_95, A_99. 

If you would have the percentiles as a separate tag, you: 
* will be able to filter more easily 
* just see A, B, C if you chose to plot 50th percentile for each metric
* just see 50, 90, 95, 99 if you plot all percentiles for A.

### DIFF
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -138,9 +138,15 @@ func (cw *CloudWatch) Send() error {
 			{"95", 0.95},
 			{"99", 0.99},
 		} {
+			dims := makeDimensions(h.h.LabelValues()...)
+			dims = append(dims, &cloudwatch.Dimension{
+				Name:  aws.String("percentile"),
+				Value: aws.String(p.s),
+			})
+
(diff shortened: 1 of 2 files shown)


feliksik:
Note that in the screenshot, only "request-duration" is part of the name; the coins.Pricing is part of another label, as two labels are plotted. The request-duration part is not unique, if it weren't for the percentiles. 

You could also argue that I don't know how to use "metric name" correctly, in which case I'd be happy to take your advice :-)

feliksik:
I think I disagree with myself :-) I'm closing it!

PR #598, opened 2017-08-04
removed deprecated functions

### LINKED ISSUE #597: Use of deprecated grpc Metadata FromContext and NewContext
Was just looking through the grpc [metadata](https://godoc.org/google.golang.org/grpc/metadata) docs the other day and noticed `metadata.NewContext` and `metadata.FromContext` have been deprecated. From [grpc-metadata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) looks seems the new prefer way is to use `FromIncomingContext` and `NewOutgoingContext` respectively (granted there does look to be a bug in the doc where it still states `NewContext` in one place, the examples had it [change](https://github.com/grpc/grpc-go/commit/0c1d39df28cc4e0369b185df2b33b1094cd543fb)).

I know that `FromContext` is used in [grpc/service.go](https://github.com/go-kit/kit/blob/master/transport/grpc/server.go#L76) and `NewContext` in [grpc/client.go](https://github.com/go-kit/kit/blob/master/transport/grpc/client.go#L94). 

This is definitely not a big thing at all, but wasn't sure if it had be brought to your attention yet or if there were any plans to replace those with `FromIncomingContext` and `NewOutgoingContext`. If there aren't and it's alright, can push up a quick change for that.

travissalascox:
changed `NewContext -> NewOutgoingContext` and `FromContext ->
FromIncomingContext` as described in
[metadata](https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go)

resolves #597

### DIFF
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -91,7 +91,7 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		for _, f := range c.before {
 			ctx = f(ctx, md)
 		}
-		ctx = metadata.NewContext(ctx, *md)
+		ctx = metadata.NewOutgoingContext(ctx, *md)
 
 		var header, trailer metadata.MD
 		grpcReply := reflect.New(c.grpcReply).Interface()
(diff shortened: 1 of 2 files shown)


peterbourgon:
Thanks!

PR #599, opened 2017-08-08
Cleaner/easier way for user to specify Cloudwatch metric percentiles

feliksik:
Two changes: 

* As proposed in #591 , prepend names with 'p'; that is: metricname_p50 instead of metricname_50. 
* Make the WithPercentiles() options much nicer: WithPercentiles(0.50, 0.99) instead of WithPercentiles(cloudwatch.Percentiles{"50": 0.50, "0.99": 0.99}) 

sorry for not including these previously, I should have added [WIP] tag, as I intended to still agree on requirements, before merge.

### DIFF
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -196,7 +196,7 @@ func (cw *CloudWatch) Send() error {
 		for _, perc := range cw.percentiles {
 			value := histogram.Quantile(perc)
 			datums = append(datums, &cloudwatch.MetricDatum{
-				MetricName: aws.String(fmt.Sprintf("%s_p%s", name, formatPerc(perc))),
+				MetricName: aws.String(fmt.Sprintf("%s_%s", name, formatPerc(perc))),
 				Dimensions: makeDimensions(lvs...),
 				Value:      aws.Float64(value),
 				Timestamp:  aws.Time(now),
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -14,6 +14,7 @@ import (
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/generic"
 	"github.com/go-kit/kit/metrics/internal/lv"
+	"strconv"
 )
 
 const (
@@ -38,7 +39,7 @@ type CloudWatch struct {
 	counters              *lv.Space
 	gauges                *lv.Space
 	histograms            *lv.Space
-	percentiles           Percentiles
+	percentiles           []float64 // percentiles to track
 	logger                log.Logger
 	numConcurrentRequests int
 }
@@ -57,16 +58,17 @@ func WithLogger(logger log.Logger) option {
 	}
 }
 
-func WithPercentiles(p Percentiles) option {
+// WithPercentiles registers the percentiles to track, overriding the
+// existing/default values.
+func WithPercentiles(percentiles ...float64) option {
 	return func(c *CloudWatch) {
-		validated := Percentiles{}
-		for _, entry := range p {
-			if entry.f < 0 || entry.f > 1 {
-				continue // illegal entry
+		c.percentiles = make([]float64, 0)
+		for _, p := range percentiles {
+			if p < 0 || p > 1 {
+				continue // illegal entry; ignore
 			}
-			validated = append(validated, entry)
+			c.percentiles = append(c.percentiles, p)
 		}
-		c.percentiles = validated
 	}
 }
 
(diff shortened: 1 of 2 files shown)


peterbourgon:
Is there a good reason to make the breaking `p` change besides consistency with other metrics?

feliksik:
@peterbourgon : the reasons are 1) consistency and 2) clarity (that it is a percentile). 

Whether that's good enough is arguable. I can also pull it out and just include the WithPercentiles() change, if you like.

feliksik:
I believe there is no tag released since it was included, so I guessed it didn't matter much; that's why I  [asked](https://github.com/go-kit/kit/pull/591#issuecomment-320534931)  @cam-stitt .

The extent to which you want to keep master backwards compatible is up to you.

feliksik:
I see  that I messed up the test:  

```
--- FAIL: TestHistogram (0.23s)
	cloudwatch_test.go:182: p50: want 500.000000, have 0.000000; p90: want 532.038789, have 0.000000; p95: want 541.121341, have 0.000000; p99: want 558.158697, have 0.000000
FAIL
```

So I'll happily fix it, as soon as you can tell me which of the 2 changes you would accept :-) goodnight!

metrics/cloudwatch/cloudwatch.go
```diff
@@ -57,16 +58,17 @@ func WithLogger(logger log.Logger) option {
...
-func WithPercentiles(p Percentiles) option {
+// WithPercentiles registers the percentiles to track, overriding the
+// existing/default values.
+func WithPercentiles(percentiles ...float64) option {
 	return func(c *CloudWatch) {
-		validated := Percentiles{}
-		for _, entry := range p {
-			if entry.f < 0 || entry.f > 1 {
-				continue // illegal entry
+		c.percentiles = make([]float64, 0)
```
seh: - [ ] Why zero here, as opposed to `len(percentiles)`?  
  It looks like you're betting in favor of them all being illegal.

feliksik: You are correct, `make([]float64, 0, len(percentiles)` would be more efficient. 

I don't care to optimize this stuff for slices that are so short; in this slice is likely to be max 4 entries. 

But I can fix it if you like.

seh: That you wrote a number at all made me take note.

I usually try to avoid the _intermediate_ reallocations (and consequent copying) in loops like these by allocating beforehand the maximum that I might need. I'd rather hang on to that unused memory (assuming it's relatively small like this) than reallocate several times accumulating the values.

I don't know how often this code executes, so it may not make any difference. For your approach, though, I recommend just removing the length entirely, deleting line 65. You can pass a nil slice argument to `append`.

feliksik: Thanks for your comment. You are right, I never needed to set it to 0. It's better to delete the line, or initialize with len(). But the current line makes no sense. 

I'll update this later :-) still waiting for a verdict on the 'p' prefix anyway.

feliksik: Hmm, thinking about it, the percentiles was already initialized by the beginning of the New() function. So not re-initializing it will add percentiles, not override the defaults. That's not good... 

For context: The method is only used once for every cloudwatch instance you make, which is generally 1 per application runtime.

seh: You _could_ write the following:
```go
c.percentiles = nil
```
Whether that's better—assuming your desire is to start here with no room—is a matter of taste.

feliksik: Seems all fine with me; I will see if I can add some tests for the custom percentiles, and will change it to either nil or Len(), whichever you like.

seh: _I_ prefer to use the length of the candidate sequence supplied to the filter—here, `len(percentiles)`—but my contribution to _go-kit_ thus far has been negligible, so I defer to any others more familiar with conventions or preferences applied more consistently throughout the project.

feliksik:
Great, will do! As long as nobody else comments, let's do as you suggest :-)

metrics/cloudwatch/cloudwatch.go
feliksik: done!

metrics/cloudwatch/cloudwatch_test.go
```diff
@@ -193,4 +193,54 @@ func TestHistogram(t *testing.T) {
...
+		svc.mtx.RLock()
+		defer svc.mtx.RUnlock()
+		p50 = svc.valuesReceived[n50]
+		p90 = svc.valuesReceived[n90]
+
+		// our teststat.TestHistogram wants us to give p95 and p99,
+		// but with custom percentiles we don't have those.
+		// So fake them. Maybe we should make teststat.nvq() public and use that?
+		p95 = 541.121341
+		p99 = 558.158697
```
feliksik: @peterbourgon this may not be the nicest to hardcode; we can also make teststat.nvq() public. Do you have a preference?

peterbourgon: Maybe it's fine to just leave this [part of the] test out, for this particular metric?

feliksik: Thanks for you comment @peterbourgon ; Do you mean leave out the whole added part about testing WithPercentile() ? 
I think that's the only way. 

We can also leave it hardcoded for now; it's all fine with me. I actually do think it's useful to test the WithPercentile(), but I'm not really attached to it. 

If you have a preference, let me know; otherwise I'll leave it as-is.

feliksik:
@peterbourgon you mentioned on slack "I rather leave it out, if you don't mind. Keep the old syntax."
Fine with me! The names are referted in the most recent commit.

peterbourgon:
Thanks!

PR #601, opened 2017-08-11
Make level color example work

fabxc:
The value is typically a `*level.levelValue` so the switch over string constants will always evaluate to false.
Since the type is not exported, casting to `fmt.Stringer` seems to be the best approach. I could also change it to `fmt.Sprintf` to handle more arbitrary cases. But for the example working with the `level` package seems fine.

### DIFF
--- a/log/term/example_test.go
+++ b/log/term/example_test.go
@@ -32,7 +32,11 @@ func ExampleNewLogger_levelColors() {
 			if keyvals[i] != "level" {
 				continue
 			}
-			switch keyvals[i+1] {
+			s, ok := keyvals[i+1].(fmt.Stringer)
+			if !ok {
+				return term.FgBgColor{}
+			}
+			switch s.String() {
 			case "debug":
 				return term.FgBgColor{Fg: term.DarkGray}
 			case "info":
--- a/log/term/example_test.go
+++ b/log/term/example_test.go
@@ -2,6 +2,7 @@ package term_test
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


peterbourgon:
@ChrisHines Make sense?

ChrisHines:
This doesn't make sense to me. The example is logging strings, which don't implement `fmt.Stringer`. The change in the PR doesn't work as intended and doesn't add clarity.

@fabxc What is the goal here? Clearly the example, as currently written, caused you confusion. It seems that perhaps you are trying to make this example work with package `github.com/go-kit/kit/log/level`, which was not the original intent. The original intent was to present a self contained example for a coloring function that switched on logged values. How can we make this example more helpful without necessarily addding a dependency on `log/level` just for example code?

fabxc:
@ChrisHines you are totally correct of course. I somehow thought it was meant to show something to work with the `level` package and repeatedly ignored the last lines of it.

In my defence, it was pretty late and I might have had a beer or two :)
Sorry for the noise.

But I could indeed imagine other people running into this because it totally looks like it would just work.  
Maybe just adding a comment somewhere to be cautious is an easy and effective solution?

oryband:
@ChrisHines any chance to get one of the contributors make `log/term` play nice with `log/level`? Currently no colors are outputted.

ChrisHines:
@oryband Rather than discuss on this closed PR, we should create a new issue with the specifics of what doesn't work when using `log/term` with `log/level`. Would you be kind enough to create the issue and describe how `log/term` doesn't play nice with `log/level`?

PR #602, opened 2017-08-15
Fix for dogstatsd metrics with default tags and no labelValues

jfhamlin:
Only omit dogstatsd tag values if there are no labelValues AND there are no default tags.


jfhamlin:
@peterbourgon, fixing a bug in the handling of the new default labelValues on dogstatsd.

Issue #604, opened 2017-08-19
Add Basic Auth support


DimaSalakhov:
Have faced a need for a Basic auth support for one of http servers I've wrote. As there wasn't such middleware handy, I've wrote my own one and would be happy to share it with others.

peterbourgon:
Sure, I'd be happy to review. Got a link to some code, or a PR?

DimaSalakhov:
Great! Have opened PR #605

PR #605, opened 2017-08-19
Add basic auth middleware

DimaSalakhov:
Issue #604

### DIFF
--- a/auth/basic/README.md
+++ b/auth/basic/README.md
@@ -1,4 +1,8 @@
-`package auth/basic` provides a Basic Authentication middleware [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication).
+This package provides a Basic Authentication middleware.
+
+It'll try to compare credentials from Authentication request header to a username/password pair in middleware constructor.
+
+More details about this type of authentication can be found in [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication).
 
 ## Usage
 
@@ -6,7 +10,7 @@
 import httptransport "github.com/go-kit/kit/transport/http"
 
 httptransport.NewServer(
-		endpoint.Chain(AuthMiddleware(cfg.auth.user, cfg.auth.password, "Example Realm"))(makeUppercaseEndpoint()),
+		AuthMiddleware(cfg.auth.user, cfg.auth.password, "Example Realm")(makeUppercaseEndpoint()),
 		decodeMappingsRequest,
 		httptransport.EncodeJSONResponse,
 		httptransport.ServerBefore(httptransport.PopulateRequestContext),
--- a/auth/basic/README.md
+++ b/auth/basic/README.md
@@ -1,4 +1,4 @@
-`package auth/basic` provides a basic auth middleware [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication)
+`package auth/basic` provides a Basic Authentication middleware [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication).
 
 ## Usage
 
@@ -13,4 +13,4 @@ httptransport.NewServer(
 	)
 ```
 
-For AuthMiddleware to be able to pick up Authentication header from a http request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```
+For AuthMiddleware to be able to pick up the Authentication header from a HTTP request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```.
--- a/auth/basic/README.md
+++ b/auth/basic/README.md
@@ -13,4 +13,4 @@ httptransport.NewServer(
 	)
 ```
 
-For AuthMiddleware to be able to pick up the Authentication header from a HTTP request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```.
+For AuthMiddleware to be able to pick up the Authentication header from an HTTP request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```.
--- a/auth/basic/README.md
+++ b/auth/basic/README.md
@@ -1,0 +1,16 @@
+`package auth/basic` provides a basic auth middleware [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication)
+
+## Usage
+
+```go
+import httptransport "github.com/go-kit/kit/transport/http"
+
+httptransport.NewServer(
+		endpoint.Chain(AuthMiddleware(cfg.auth.user, cfg.auth.password, "Example Realm"))(makeUppercaseEndpoint()),
+		decodeMappingsRequest,
+		httptransport.EncodeJSONResponse,
+		httptransport.ServerBefore(httptransport.PopulateRequestContext),
+	)
+```
+
+For AuthMiddleware to be able to pick up Authentication header from a http request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```
--- a/auth/basic/middleware.go
+++ b/auth/basic/middleware.go
@@ -14,12 +14,12 @@ import (
 	httptransport "github.com/go-kit/kit/transport/http"
 )
 
-// AuthError represents an authoriation error.
+// AuthError represents an authorization error.
 type AuthError struct {
 	Realm string
 }
 
-// StatusCode is an iimplementation of the StatusCoder interface in go-kit/http.
+// StatusCode is an implementation of the StatusCoder interface in go-kit/http.
 func (AuthError) StatusCode() int {
 	return http.StatusUnauthorized
 }
@@ -29,7 +29,7 @@ func (AuthError) Error() string {
 	return http.StatusText(http.StatusUnauthorized)
 }
 
-// Headers is an implemntation of the Headerer interface in go-kit/http.
+// Headers is an implementation of the Headerer interface in go-kit/http.
 func (e AuthError) Headers() http.Header {
 	return http.Header{
 		"Content-Type":           []string{"text/plain; charset=utf-8"},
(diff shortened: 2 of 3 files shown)


auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+// Headers is an implemntation of Headerer interface in go-kit/http
+func (e AuthError) Headers() http.Header {
+	return http.Header{
+		"Content-Type":           []string{"text/plain; charset=utf-8"},
+		"X-Content-Type-Options": []string{"nosniff"},
+		"WWW-Authenticate":       []string{fmt.Sprintf(`Basic realm=%q`, e.Realm)}}
+}
+
+func credsAreValid(givenUser, givenPass, requiredUser, requiredPass string) bool {
+	// Equalize lengths of supplied and required credentials
```
seh: Why is it important to make the lengths equal? Are you trying to preclude `subtle.ConstantTimeCompare` from divulging the difference in length by way of it short-circuiting the rest of the comparison?

Though highly unlikely, it looks like you're risking accepting a potential hash collision in trade for defending against a timing attack.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+	// by hashing them
+	givenUserBytes := sha256.Sum256([]byte(givenUser))
+	givenPassBytes := sha256.Sum256([]byte(givenPass))
+	requiredUserBytes := sha256.Sum256([]byte(requiredUser))
```
seh: Since the required user name and password values are fixed at line 85 below, if you decide to retain this hashing technique, consider computing the hash on these two values once just before line 77, and reusing the hash instead of the raw values here.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+type AuthError struct {
+	Realm string
+}
+
+// StatusCode is an implemntation of StatusCoder interface in go-kit/http
+func (AuthError) StatusCode() int {
+	return http.StatusUnauthorized
+}
+
+// Error is an implemntation of Error interface
```
seh: - [ ] s/implemntation of Error/implementation of the Error/
- [ ] Punctuate this sentence.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/go-kit/kit/endpoint"
+	httptransport "github.com/go-kit/kit/transport/http"
+)
+
+// AuthError represents generic Authorization error
```
seh: - [ ] s/represents generic Authorization/represents an authoriation/
- [ ] Punctuate this sentence.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+type AuthError struct {
+	Realm string
+}
+
+// StatusCode is an implemntation of StatusCoder interface in go-kit/http
```
seh: - [ ] s/implemntation of StatusCoder/implementation of the StatusCoder/
- [ ] Punctuate this sentence.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+func (AuthError) StatusCode() int {
+	return http.StatusUnauthorized
+}
+
+// Error is an implemntation of Error interface
+func (AuthError) Error() string {
+	return http.StatusText(http.StatusUnauthorized)
+}
+
+// Headers is an implemntation of Headerer interface in go-kit/http
```
seh: - [ ] s/an implemntation of Headerer/an implementation of the Headerer/
- [ ] Punctuate this sentence.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+func (e AuthError) Headers() http.Header {
+	return http.Header{
+		"Content-Type":           []string{"text/plain; charset=utf-8"},
+		"X-Content-Type-Options": []string{"nosniff"},
+		"WWW-Authenticate":       []string{fmt.Sprintf(`Basic realm=%q`, e.Realm)}}
+}
+
+func credsAreValid(givenUser, givenPass, requiredUser, requiredPass string) bool {
```
seh: - [ ] Given that the currency of your implementation is byte vectors, consider demanding byte vectors as parameters.  
  You could then return byte vectors from your `parseBasicAuth` function below, rather than converting the byte vectors resulting from decoding the base64-encoded values to strings at line 67, only to convert them back—with allocation and copying cost—here.

auth/basic/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	cs := string(c)
+	s := strings.IndexByte(cs, ':')
+	if s < 0 {
+		return
+	}
+	return cs[:s], cs[s+1:], true
+}
+
+// AuthMiddleware returns a Basic Authentication middleware for a particular user and password
```
seh: - [ ] Punctuate this sentence.

auth/basic/README.md
```diff
@@ -0,0 +1,16 @@
+`package auth/basic` provides a basic auth middleware [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication)
```
seh: - [ ] s/basic auth/Basic Authorization/
- [ ] Punctuate this sentence.

auth/basic/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+httptransport.NewServer(
+		endpoint.Chain(AuthMiddleware(cfg.auth.user, cfg.auth.password, "Example Realm"))(makeUppercaseEndpoint()),
+		decodeMappingsRequest,
+		httptransport.EncodeJSONResponse,
+		httptransport.ServerBefore(httptransport.PopulateRequestContext),
+	)
+```
+
+For AuthMiddleware to be able to pick up Authentication header from a http request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```
```
seh: - [ ] s/up Authentication/up the "Authentication"/
- [ ] s/a http request/an HTTP request,/
- [ ] Punctuate this sentence, or s/like/like the following:/.

DimaSalakhov:
@seh Thanks for your review, everything was on point, I very much appreciate it!

In regards to the hash sum calculation ```sha256.Sum256()``` according to my very limited knowledge on the topic of timing attack, there are 2 suggestions to prevent them:
- use constant time algorithm to prevent bruteforcing the exact position of the first difference in byte arrays. That's exactly what subtle/ConstantTimeCompare does
- use byte arrays of the same length for the similar reason. In our case subtle/ConstantTimeCompare short circuit if arrays have a different length.
I might be wrong in that, but this was a premise for me to compare hashes rather than slices. You're right about a potential collision in hashes, but my assumption is that the chances that person would successfully use that for bruteforce are higher than a chance of hash collision. However, I don't have a data to prove it and would be happy to hear a security advice.

seh:
I agree with the first benefit of using `subtle.ConstantTimeCompare`. Note, though, that computing the hash requires various length-dependent operations, so while we eliminate length difference detection threat from the final byte comparison, we still face timing differences in computing the hashes.

However, if you compute the hashes for the expected user name and password values ahead of time, as [I had suggested earlier](https://github.com/go-kit/kit/pull/605#discussion_r135419258), then an attacker wouldn't be able to detect any relationship between computing the hashes of the expected and supplied values. You would then face only the threat of a hash collision yielding a false positive.

If the two hashes differ, we know that the supplied values do not match the expected values. If we wanted to be completely sure that a match in the hashes means the underlying values match, we'd then have to compare those underlying values directly—which brings us back to divulging differences in length via timing attack (even with `subtle.ConstantTimeCompare`), and wastes the utility of using the hash function.

I too would like to hear from others whether we should worry about the likelihood of a hash collision here.

auth/basic/middleware.go
```diff
@@ -13,49 +14,32 @@ import (
 	httptransport "github.com/go-kit/kit/transport/http"
 )
 
-// AuthError represents generic Authorization error
+// AuthError represents an authoriation error.
```
seh: - [ ] s/authoriation/authorization/

auth/basic/middleware.go
```diff
@@ -13,49 +14,32 @@ import (
...
 type AuthError struct {
 	Realm string
 }
 
-// StatusCode is an implemntation of StatusCoder interface in go-kit/http
+// StatusCode is an iimplementation of the StatusCoder interface in go-kit/http.
```
seh: - [ ] s/iimplementation/implementation/

auth/basic/middleware.go
```diff
@@ -64,16 +48,19 @@ func parseBasicAuth(auth string) (username, password string, ok bool) {
...
 		return
 	}
-	return cs[:s], cs[s+1:], true
+	return c[:s], c[s+1:], true
 }
 
-// AuthMiddleware returns a Basic Authentication middleware for a particular user and password
+// AuthMiddleware returns a Basic Authentication middleware for a particular user and password.
 func AuthMiddleware(requiredUser, requiredPassword, realm string) endpoint.Middleware {
+	requiredUserBytes := sha256.Sum256([]byte(requiredUser))
```
seh: - [ ] Convert this array to a slice here, rather than down at line 77 each time you use it.

auth/basic/middleware.go
```diff
@@ -64,16 +48,19 @@ func parseBasicAuth(auth string) (username, password string, ok bool) {
...
+	requiredPassBytes := sha256.Sum256([]byte(requiredPassword))
```
seh: - [ ] Convert this array to a slice here, rather than down at line 78 each time you use it.
- [ ] Consider using four extra characters to name the variable "requiredPasswordBytes."

auth/basic/README.md
```diff
@@ -13,4 +13,4 @@ httptransport.NewServer(
 	)
 ```
 
-For AuthMiddleware to be able to pick up Authentication header from a http request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```
\ No newline at end of file
+For AuthMiddleware to be able to pick up the Authentication header from a HTTP request we need to pass it through the context with something like ```httptransport.ServerBefore(httptransport.PopulateRequestContext)```.
```
seh: - [ ] We could go either way, but in American English, it sounds better to my ear to write "an HTTP."

DimaSalakhov:
@seh I agree with your statement about creating and comparing hashes.
As for the hash collision, I'm leaning more and more towards calling it possible, but unrealistic with the state of the modern hardware. Here is an article from Google about collision they've found in sha128: https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html
From what I see, it takes 1year, 110GPU and right algorithm to achieve it. SHA256 is twice the size and it would be a very long and expensive attack.

auth/basic/middleware.go
```diff
@@ -56,26 +56,31 @@ func parseBasicAuth(auth string) (username, password []byte, ok bool) {
...
 	return func(next endpoint.Endpoint) endpoint.Endpoint {
 		return func(ctx context.Context, request interface{}) (interface{}, error) {
 			auth := ctx.Value(httptransport.ContextKeyRequestAuthorization).(string)
-			givenUser, givenPass, ok := parseBasicAuth(auth)
+			givenUser, givenPassword, ok := parseBasicAuth(auth)
 			if !ok {
 				return nil, AuthError{realm}
 			}
 
 			// Equalize lengths of supplied and required credentials by hashing them.
```
seh: - [ ] Since this comment points out _what_ is happening but not why, consider removing this comment.

auth/basic/README.md
```diff
@@ -0,0 +1,16 @@
+`package auth/basic` provides a Basic Authentication middleware [Mozilla article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication).
```
peterbourgon: Please make this sentence make grammatical sense :) For example,

```
This package provides a Basic Authentication middleware.
Here is a [Mozilla article](...) with details.
```

auth/basic/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Usage
+
+```go
+import httptransport "github.com/go-kit/kit/transport/http"
+
+httptransport.NewServer(
+		endpoint.Chain(AuthMiddleware(cfg.auth.user, cfg.auth.password, "Example Realm"))(makeUppercaseEndpoint()),
```
peterbourgon: There's no reason to use the Chain helper here, I don't think?

auth/basic/middleware.go
```diff
@@ -0,0 +1,89 @@
...
+func (AuthError) Error() string {
+	return http.StatusText(http.StatusUnauthorized)
+}
+
+// Headers is an implementation of the Headerer interface in go-kit/http.
+func (e AuthError) Headers() http.Header {
+	return http.Header{
+		"Content-Type":           []string{"text/plain; charset=utf-8"},
+		"X-Content-Type-Options": []string{"nosniff"},
+		"WWW-Authenticate":       []string{fmt.Sprintf(`Basic realm=%q`, e.Realm)}}
```
peterbourgon: Please use a trailing comma and put the closing brace on the next line.

auth/basic/middleware.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// AuthMiddleware returns a Basic Authentication middleware for a particular user and password.
+func AuthMiddleware(requiredUser, requiredPassword, realm string) endpoint.Middleware {
+	requiredUserBytes := toHashSlice([]byte(requiredUser))
+	requiredPasswordBytes := toHashSlice([]byte(requiredPassword))
+
+	return func(next endpoint.Endpoint) endpoint.Endpoint {
+		return func(ctx context.Context, request interface{}) (interface{}, error) {
+			auth := ctx.Value(httptransport.ContextKeyRequestAuthorization).(string)
```
peterbourgon: Please check for the presence of the value in the context before type-asserting it to a string. If it's not there, the middleware should return an appropriate error.

peterbourgon:
With those changes, this looks good to me. Thanks for the initiative and sorry for my delays!

auth/basic/middleware.go
DimaSalakhov: I'm not sure it's necessary to check the presence of the value as type assertion will do it as well. From the authentication process view, a missing header is as invalid as empty.

peterbourgon:
Yep! That works 👍

PR #622, opened 2017-10-09
Remove dependency on juju

nelz9999:
As per the discussion on #582


nelz9999:
Ahhh.. `stringsvc3` depends on the old code... Will update!

josedonizetti:
@nelz9999 Not sure if you have the time, but something handy some others packages have, is a README with a simple example. First time using go-kit the READMEs for the log, and metrics were very useful.

peterbourgon:
Thanks a bunch!

PR #628, opened 2017-11-02
Create a node that creates a rule

heyanlong:
e.g. /path/name/name-rand
e.g. /micro-service/staging/staging-c5f4cb61de8d459592114d440e1dfc1a

### DIFF
--- a/sd/zk/client.go
+++ b/sd/zk/client.go
@@ -9,6 +9,9 @@ import (
 	"github.com/samuel/go-zookeeper/zk"
 
 	"github.com/go-kit/kit/log"
+	"io"
+	"crypto/rand"
+	"fmt"
 )
 
 // DefaultACL is the default ACL to use for creating znodes.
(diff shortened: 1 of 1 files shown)


peterbourgon:
Can you please explain your use case (requirement), and how this code change solves that requirement?

heyanlong:
The old code is so working
e.g. 
```golang
zkService := zk.Service{
    Path: "/micro-service/",
    Name: "staging",
    Data: []byte("172.16.2.23:10086")
}
```
Regiser node to ZK, create tow node to ZK 
```
1. Invalid node, node data is empty , call deregister func remove this node
/micro-service/staging
2. Ephemeral node ,node hours when session is broken, deregister func is invalid
/micro-service/_c_863ac654f07c5590fef371785e6a340e-staging0000000005
```
I think this is not very good, should be used at the end of the service Deregister delete node only

so, I changed the original code and let him work for the way
1. create one node to ZK
```
/micro-service/staging-72820bc03176868747612f9729ae1997
```
2. call deregister func remove node, not create ephemeral node

peterbourgon:
I'm really sorry, I can't understand the problem, nor how this code represents an improvement.

Issue #627, opened 2017-10-26
Consul SD causing 100% cpu usage if the service is down


Xzya:
Hey,

It seems that if the consul service is down, the Instancer will constantly try to get the list of services, causing 100% CPU usage until the service is back up. There should probably be a backoff mechanism which will only try after X amount of time. Here is how I'm using the client:

```go
var client consulsd.Client
{
	consulConfig := consulapi.DefaultConfig()
	consulConfig.WaitTime = 10 * time.Second
	consulConfig.Address = config.Consul.Address
	consulClient, err := consulapi.NewClient(consulConfig)
	if err != nil {
		return Endpoints{}, err
	}
	client = consulsd.NewClient(consulClient)
}

var (
	retryMax     = 1
	retryTimeout = 2 * time.Second
	tags         = []string{environment}
	passingOnly  = true
	endpoints    = Endpoints{}
	instancer    = consulsd.NewInstancer(client, logger, ServiceName, tags, passingOnly)
)
{
	factory := MainDBClientFactory(MakeRegisterEndpoint)
	endpointer := sd.NewEndpointer(instancer, factory, logger)
	balancer := lb.NewRoundRobin(endpointer)
	retry := lb.Retry(retryMax, retryTimeout, balancer)
	endpoints.RegisterEndpoint = retry
}
...
```

Terminal output:
`ts=2017-10-26T11:08:33.168160872Z caller=instancer.go:69 service=maindb tags=[local] err="Get http://192.168.99.100:8500/v1/health/service/maindb?passing=1&tag=local&wait=10000ms: dial tcp 192.168.99.100:8500: getsockopt: connection refused"
ts=2017-10-26T11:08:33.171253831Z caller=instancer.go:69 service=maindb tags=[local] err="Get http://192.168.99.100:8500/v1/health/service/maindb?passing=1&tag=local&wait=10000ms: dial tcp 192.168.99.100:8500: getsockopt: connection refused"
ts=2017-10-26T11:08:33.175242518Z caller=instancer.go:69 service=maindb tags=[local] err="Get http://192.168.99.100:8500/v1/health/service/maindb?passing=1&tag=local&wait=10000ms: dial tcp 192.168.99.100:8500: getsockopt: connection refused"
[...]

PR #635, opened 2017-12-03
Add backoff package and fix Consul CPU usage

nicot:
Justification for jitter and growth factor: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/.

Add backoff to the Consul instancer loop.

Fixes https://github.com/go-kit/kit/issues/627.


rjeczalik:
If I may chime in with one comment about backoff package api.

Overall, the `ExponentialBackoff` design looks similar [to this one](https://github.com/cenkalti/backoff/blob/master/exponential.go#L54). What I do not find good, as a user of the latter, is:

-  lack of meaningful zero-value - the zero-value ExponentialBackoff should use defaults, right now it won't work
- requirement to call Reset() each time one changes ExponentialBackoff interval (which is not obvious and not documented)

From the minor stuff I think test should not call directly an unexported methods of the type being tested.

peterbourgon:
Sorry I didn't see this originally, late last year was hectic time for me :(

backoff/backoff.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ExponentialBackoff provides jittered exponential durations for the purpose of
+// avoiding flodding a service with requests.
+type ExponentialBackoff struct {
+	Interval time.Duration
+	Max      time.Duration
+
+	currentInterval atomic.Value
+	cancel          <-chan struct{}
+}
```
peterbourgon: Mixing exported and non-exported fields _can_ be OK, but then you also mix direct and constructor initialization for this type, and combined I think that's a recipe for trouble...

backoff/backoff.go
```diff
@@ -0,0 +1,70 @@
...
+
+// next provides the exponential jittered backoff value. See
+// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+// for rationale.
+func (b *ExponentialBackoff) next() time.Duration {
+	current := b.currentInterval.Load().(time.Duration)
+	d := float64(current * 2)
+	jitter := rand.Float64() + 0.5
+	return time.Duration(d * jitter)
+}
```
peterbourgon: It seems like this jitter bit is the real meat of the change. Rather than adding a new package, types, API, package docs, etc. — can you instead put this jitter calculation into the old exponential func, statelessly?

util/conn/manager.go
```diff
@@ -132,14 +132,6 @@ func dial(d Dialer, network, address string, logger log.Logger) net.Conn {
 	return conn
 }
 
-func exponential(d time.Duration) time.Duration {
-	d *= 2
-	if d > time.Minute {
-		d = time.Minute
-	}
-	return d
-}
```
peterbourgon: I'm thinking a 1-line change can be enough, something like this, feel free to make corrections...

```go
func exponential(d time.Duration) time.Duration {
    d *= 2
    if d > time.Minute {
        d = time.Minute
    }
    return d * (rand.Float64() + 0.5)
}
```

peterbourgon:
@nicot Ping? I'd love to be able to close the corresponding issue :)

nicot:
Thanks for the review @rjeczalik and @peterbourgon.

I like your simple version a lot more @peterbourgon. There's a little bit of casting funkiness because duration is an int64 and we want to multiply it by a float, so let me know if there is a better way to do that.

Do you think the Exponential function belongs in the util/conn package? Or should it be somewhere more accessible?

nicot:
@peterbourgon ping back at you =)

sd/consul/instancer.go
```diff
@@ -70,6 +73,8 @@ func (s *Instancer) loop(lastIndex uint64) {
 			s.cache.Update(sd.Event{Err: err})
 		default:
 			s.cache.Update(sd.Event{Instances: instances})
+			time.Sleep(d)
+			d = conn.Exponential(d)
```
peterbourgon: Sorry for not catching this earlier. We want to implement an upper limit e.g. 30s, and have a successful reconnect reset the `d` value to its default state.

sd/consul/instancer.go
```diff
@@ -59,6 +61,7 @@ func (s *Instancer) loop(lastIndex uint64) {
 	var (
 		instances []string
 		err       error
+		d         time.Duration = time.Millisecond * 10
```
peterbourgon: `d = 10 * time.Millisecond` is preferable here.

peterbourgon:
Still need an upper limit.

nicot:
The upper limit is in the Exponential function -- Do you want both?

peterbourgon:
Oh! Gosh. Brainfart. Thanks!

Issue #639, opened 2017-12-10
Add full example for kitgen 


Raffo:
Hi, I was having a look at the PR for `kitgen`, great stuff! I noticed though that there is no full example of how to use it to build a real service. Do you think it would make sense to add? 
Additionally, I wonder why the `makeEndpoint` functions are not exported... I would use them from the main.go in a similar fashion like that: 

```
package main

import (
	"net/http"
	"os"

	kit_http "test/foo/http"

	"test/foo/endpoints"
	"test/foo/service"

	"github.com/go-kit/kit/log"
)

func main() {
	logger := log.NewLogfmtLogger(os.Stderr)

	var svc service.Service
	svc = service.ServiceStruct{}

	e := endpoints.Endpoints{}

	e.Concat = endpoints.MakeConcatEndpoint(svc)
	e.Sum = endpoints.MakeSumEndpoint(svc)

	handler := kit_http.NewHTTPHandler(e)

	http.Handle("/", handler)
	logger.Log("msg", "HTTP", "addr", ":8080")
	logger.Log("err", http.ListenAndServe(":8080", nil))
}

```

I'm probably wrong and there is another way to do that, and I think that a full example with a `main.go` would help to understand that.

Raffo:
@peterbourgon I see that you labeled the issue, if I can get an answer to my questions in the original comment, I would be interested in contributing the examples personally ^_^

peterbourgon:
> I noticed though that there is no full example of how to use it to build a real service. Do you think it would make sense to add?

This would be a great addition to the README in that directory. PR happily accepted.

>  wonder why the makeEndpoint functions are not exported... I would use them from the main.go in a similar fashion like that:

Yes, they should be exported! An oversight. PR happily accepted.

Raffo:
I created a PR to fix the unexported functions ( https://github.com/go-kit/kit/pull/641 ), will do the docs as well later on.

retnuh:
As someone new to go-kit and having watched a few talks that keep referencing "kitgen", I'd be happy to beta-test any examples/tutorials!

Raffo:
@nyarly I was trying to understand the way to work with `kitgen` and it looks like the service in the `service` folder is **always** overwritten. Can you confirm that this behaviour is intended? Also, I could not find the `*.tmpl` files mentioned in `ast_templates.go`, did you forget to commit them or is there something I am missing? 

[...]

PR #641, opened 2017-12-12
Export makeEndpoint in kitgen

Raffo:
This PR addresses a comment regarding exported `makeEndpoint` function in #639: with this fix, we will make those function exported which allow to use them, for example like that: 

```
package main

import (
	"net/http"
	"os"

	kit_http "test/foo/http"

	"test/foo/endpoints"
	"test/foo/service"

	"github.com/go-kit/kit/log"
)

func main() {
	logger := log.NewLogfmtLogger(os.Stderr)

	var svc service.Service
	svc = service.ServiceStruct{}

	e := endpoints.Endpoints{}

	e.Concat = endpoints.MakeConcatEndpoint(svc)
	e.Sum = endpoints.MakeSumEndpoint(svc)

	handler := kit_http.NewHTTPHandler(e)

	http.Handle("/", handler)
	logger.Log("msg", "HTTP", "addr", ":8080")
	logger.Log("err", http.ListenAndServe(":8080", nil))
}
```

### DIFF
--- a/cmd/kitgen/method.go
+++ b/cmd/kitgen/method.go
@@ -143,7 +143,7 @@ func (m method) encoderFunc() ast.Decl {
 }
 
 func (m method) endpointMakerName() *ast.Ident {
-	return id("make" + m.name.Name + "Endpoint")
+	return id("Make" + m.name.Name + "Endpoint")
 }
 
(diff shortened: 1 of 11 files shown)


peterbourgon:
Woop! Thanks!

Issue #643, opened 2017-12-15
Multiple same-name Headers are ignored


klaus:
In `transport/http/server.go` i have found 2 places where the headers are not correctly set to all values the http.Header might be holding.

http.Header is a `map[string][]string` and `.Get(k)` only returns the first of those fields from the []string array. I think to correctly handle it, we should do as the docs suggest:

```
// Get gets the first value associated with the given key.
// It is case insensitive; textproto.CanonicalMIMEHeaderKey is used
// to canonicalize the provided key.
// If there are no values associated with the key, Get returns "".
// To access multiple values of a key, or to use non-canonical keys,
// access the map directly.
func (h Header) Get(key string) string {
	return textproto.MIMEHeader(h).Get(key)
}
```

which would be sthg along the lines ...
```
		for k,hdrs := range headerer.Headers() {
			for _,h := range hdrs {
                             w.Header().Add(k, h))
                        }
		}
```

https://github.com/go-kit/kit/blob/1b4cf21e19ea1dae31038b02c4982eea5b4214b7/transport/http/server.go#L148

peterbourgon:
Yes, fully agreed. An easy fix, where's the PR? ;)

klaus:
https://github.com/go-kit/kit/pull/644

PR #644, opened 2017-12-18
Allow multiple HTTP-Headers in Server

klaus:
This is the awaited pull request for the Multi-HTTP Header issue https://github.com/go-kit/kit/issues/643

### DIFF
--- a/log/level/level.go
+++ b/log/level/level.go
@@ -177,7 +177,7 @@ func DebugValue() Value { return debugValue }
 
 var (
 	// key is of type interfae{} so that it allocates once during package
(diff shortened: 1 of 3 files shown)


peterbourgon:
Looks great! Minor nit, but could we add a test for this behavior? I don't want to regress 😇

klaus:
@peterbourgon sorry, X-mas freeze. will attend to it next couple of days

klaus:
@peterbourgon sorry for the long delay, I hope these tests will suffice, they do fail if you run it against the non-patched code

peterbourgon:
Thanks! 💯

PR #659, opened 2018-02-02
HTTP server NoOpRequestDecoder

### LINKED ISSUE #657: Allow nil HTTP Server Decoders
Currently `http. NewServer()` requires that you pass in a decoder otherwise it will panic in `ServeHTTP()`. This is slightly annoying for routes that have no input to decode. For instance GET `/users`, or something similar. For now I just created a common no-op decoder that I use in these cases, but it would be nice to simply pass `nil`.

The change is rather easy, and I am happy to provide a PR. Just wanted to make sure I wasn't missing some use case as to why it is this way first.

jdolce:
Adds NoOpRequestDecoder for HTTP request that do not need to be decoded.

Fixes #657

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -136,6 +136,12 @@ type ErrorEncoder func(ctx context.Context, err error, w http.ResponseWriter)
 // provided in the context under keys with the ContextKeyResponse prefix.
 type ServerFinalizerFunc func(ctx context.Context, code int, r *http.Request)
 
+// NopRequestDecoder is a DecodeRequestFunc that can be used for requests that do not
+// need to be decoded, and simply returns nil, nil.
+func NopRequestDecoder(ctx context.Context, r *http.Request) (interface{}, error) {
+	return nil, nil
+}
+
 // EncodeJSONResponse is a EncodeResponseFunc that serializes the response as a
 // JSON object to the ResponseWriter. Many JSON-over-HTTP services can use it as
 // a sensible default. If the response implements Headerer, the provided headers
--- a/transport/http/server_test.go
+++ b/transport/http/server_test.go
@@ -393,6 +393,28 @@ func TestEnhancedError(t *testing.T) {
 	}
 }
 
+func TestNoOpRequestDecoder(t *testing.T) {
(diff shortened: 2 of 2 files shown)


transport/http/server.go
```diff
@@ -136,6 +136,12 @@ type ErrorEncoder func(ctx context.Context, err error, w http.ResponseWriter)
 // provided in the context under keys with the ContextKeyResponse prefix.
 type ServerFinalizerFunc func(ctx context.Context, code int, r *http.Request)
 
+// NoOpRequestDecoder is a DecodeRequestFunc that can be used for requests that do not
+// need to be decoded, and simply returns nil, nil.
+func NoOpRequestDecoder(ctx context.Context, r *http.Request) (interface{}, error) {
```
ChrisHines: I know it might look a bit strange, but I think in Go we usually spell this `NopRequestDecoder`. See for precedent:

https://golang.org/pkg/io/ioutil/#NopCloser
https://godoc.org/github.com/go-kit/kit/log#NewNopLogger
https://godoc.org/golang.org/x/text/transform#NopResetter

transport/http/server_test.go
```diff
@@ -393,6 +393,33 @@ func TestEnhancedError(t *testing.T) {
...
+
+	server := httptest.NewServer(handler)
+	defer server.Close()
+	go http.Get(server.URL)
+
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("timeout waiting for finalizer")
+	}
```
ChrisHines: I think this test is doing too much. It seems to me that it would be perfectly valid to test this without spinning up an http server and extra goroutines. Consider calling `handler.ServeHTTP` directly. You can use `httptest.NewRequest` and `httptest.NewRecorder` to construct appropriate arguments for `ServeHTTP` and do it all sequentially with no need for worrying about timeouts, done channels or stopping a server.

jdolce:
@ChrisHines Thanks for the review. I made the requested changes. Cheers.

peterbourgon:
Fun! Thanks! I guess we ought to do a release soon...

Issue #660, opened 2018-02-05
Metrics - Cloudwatch - Histogram


nelz9999:
Heya folks!

I'm using the metrics package with CloudWatch, and I'd like to propose an addition.

The background:
I can see some use for how the current implementation uses in-memory histogram to emit stuff to CloudWatch, but really I don't need that... I can get the percentiles I need via native CloudWatch. (https://aws.amazon.com/about-aws/whats-new/2016/11/amazon-cloudwatch-adds-support-for-percentiles/)

So, in the short-term, I've just been emitting my latency-related metrics as `Counter`s, but that feels like I'm fighting the expressiveness added by `go-kit`. (I.e. "I expect this metric to be of a latency/duration type.")

What would you all feel about me adding something like a `NewNativeHistogram(...)` that conforms to the `Histogram` API, but actually just puts measurements into the `Counter` buckets?

Thanks!

peterbourgon:
Happy to take such a PR! Please make sure to include some basic test coverage.

nelz9999:
Urgh... This is not working. 😞 

On my first attempt, the initial implementation seemed to work okay... But then I got blocked trying to write the unit test. (The test harnesses don't really have a way of "crossing the streams" between counter-like behavior and histogram-like behavior.)

My second attempt, I tried to just create a converter on the `cloudwatch.Counter` type, but I realized I couldn't make the `With(...)` function work, unless I changed the signature to return the concrete struct rather than the interface.

At this point I'm thinking that I'll just make my own Counter->Histogram wrapper, so that I don't expose y'all to my weird corner case.

PR #665, opened 2018-02-17
Enable using counters as the other types

nelz9999:
As a potential answer to my issue #660 ...

Background: I see that the current implementation for `cloudwatch` tries to be helpful at write-time for the metrics: for Histograms it does the aggregation in memory and spits out a couple of entries when sending up to AWS; for Gauges it uses the latest entry in the series per attempt to write.

But, to my understanding, this is actually contrary to how to use CloudWatch. I believe you just send the raw entries up to the server, and Counter/Gauge/Histogram meaning is applied to the data sets at read-time.

So, the rawest ability to send all the data in the `cloudwatch` package is via using the Counter type. But, I would _like_ to keep the contextual intent provided by the go-kit abstractions. ("This value will be read as a gauge."; "This value will be read as a latency."; etc...)

So, this PR represents my attempt at putting Histogram/Gauge decoration on the Counter workhorse. I did it as its own `convert` package (spiritual sibling to the `discard` and `multi` packages) that can be applied to any of the other types, since I don't know if any of them suffer from similar usage/implementation patterns.

So far I've only implemented Counter<->Histogram and Counter<->Gauge. (Leaving out Histogram<->Gauge because I'm not sure if it's a good idea, or if I'm just blindly going for "completeness".)

My main question: does this look like a valid approach? Am I polluting go-kit while trying to address my personal corner case?

### DIFF
--- a/metrics/convert/convert.go
+++ b/metrics/convert/convert.go
@@ -88,3 +88,48 @@ func (gc gaugeCounter) With(labelValues ...string) metrics.Counter {
 func (gc gaugeCounter) Add(delta float64) {
 	gc.g.Set(delta)
 }
+
+type histogramGauge struct {
+	h metrics.Histogram
+}
+
+// NewHistogramAsGauge returns a Gauge that actually writes the
+// value on an underlying Histogram
+func NewHistogramAsGauge(h metrics.Histogram) metrics.Gauge {
+	return histogramGauge{h}
+}
+
+// With implements Gauge.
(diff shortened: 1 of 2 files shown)


nelz9999:
I added Gauge<->Histogram converters too, just for completeness, not because I have a concrete use case in mind.

nelz9999:
CircleCI build failure seems to be unrelated...

nelz9999:
Gonna un-WIP this PR. It seems like it will address my needs, and it's got test coverage percentage in the 90s...

PR #671, opened 2018-03-06
Zipkin tracing support (Native v2)

basvanbeek:
This adds native Zipkin support for the HTTP and gRPC transports.

It is a more idiomatic implementation benefiting people using Zipkin in their polyglot microservice environment and lowers the amount of necessary boilerplate in Go kit compared to the OpenTracing implementation by allowing for global service middleware setup.


peterbourgon:
Gave it a quick review. Thanks for updating the tests and examples.

PR #680, opened 2018-03-15
Transport for nats

kirooha:
The PR contains transport for NATS (messaging system, https://nats.io)
I saw a request with the similar functional https://github.com/go-kit/kit/pull/623
Since the PR is still open, I want to suggest my PR.


peterbourgon:
Oh, cool.. ! I would want to also see Publisher type (corresponding to Client), can that be added?

kirooha:
yes, sure! I am adding

examples/stringsvc4/main.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"log"
+	"strings"
+	"flag"
+
+	"github.com/go-kit/kit/endpoint"
+	natstransport "github.com/kirooha/kit/transport/nats"
```
MrEhbr: github.com/kirooha/kit/transport/nats - maybe need to change to github.com/go-kit/kit/transport/nats

transport/nats/publisher.go
```diff
@@ -0,0 +1,138 @@
...
+		if err := p.enc(ctx, &msg, request); err != nil {
+			return nil, err
+		}
+
+		for _, f := range p.before {
+			ctx = f(ctx, &msg)
+		}
+
+		if !p.async {
+			resp, err := p.publisher.Request(p.subject, msg.Data, p.timeout)
```
nussjustin: Since you already have a context you can use `RequestWithContext` instead and remove the timeout field.

transport/nats/publisher.go
```diff
@@ -0,0 +1,138 @@
...
+			response, err := p.dec(ctx, resp)
+			if err != nil {
+				return nil, err
+			}
+
+			return response, nil
+
+		}
+
+		if p.reply == "" {
```
nussjustin: You can just use `PublishMsg` and than remove the rest of the method.

transport/nats/encode_decode.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// DecodeRequestFunc extracts a user-domain request object from a publisher
+// request object. It's designed to be used in NATS subscribers, for subscriber-side
+// endpoints. One straightforward DecodeRequestFunc could be something that
+// JSON decodes from the request body to the concrete response type.
+type DecodeRequestFunc func(context.Context, *nats.Msg) (request interface{}, err error)
+
+
+// EncodeRequestFunc encodes the passed request object into the HTTP request
```
nussjustin: This comment still talks about HTTP

transport/nats/request_response_funcs.go
```diff
@@ -0,0 +1,22 @@
...
+// request context. In Subscribers, RequestFuncs are executed prior to invoking the
+// endpoint.
+type RequestFunc func(context.Context, *nats.Msg) context.Context
+
+// SubscriberResponseFunc may take information from a request context and use it to
+// manipulate a Publisher. SubscriberResponseFuncs are only executed in
+// subscribers, after invoking the endpoint but prior to publishing a reply.
+type SubscriberResponseFunc func(context.Context, *nats.Conn) context.Context
+
+// ClientResponseFunc may take information from an HTTP request and make the
```
nussjustin: HTTP

transport/nats/subscriber.go
```diff
@@ -0,0 +1,167 @@
...
+// of error handling, including logging in more detail, should be performed in a
+// custom SubscriberErrorEncoder which has access to the context.
+func SubscriberErrorLogger(logger log.Logger) SubscriberOption {
+	return func(s *Subscriber) { s.logger = logger }
+}
+
+// ServeMsg provides nats.MsgHandler.
+func (s Subscriber) ServeMsg(nc *nats.Conn) func(msg *nats.Msg) {
+	return func(msg *nats.Msg) {
+		ctx := context.Background()
```
nussjustin: The context is never cancelled

transport/nats/publisher.go
```diff
@@ -0,0 +1,138 @@
...
+
+		if err := p.enc(ctx, &msg, request); err != nil {
+			return nil, err
+		}
+
+		for _, f := range p.before {
+			ctx = f(ctx, &msg)
+		}
+
+		if !p.async {
```
nussjustin: The endpoint tries to handle both simple publish operations and requests which feels kinda odd. Also the !async case ignores the p.reply topic.

I suggest to remove the async case and the custom reply topic for know and only do a simple `Request(WithContext)`. We can always add support for messages without reply later through options or a different type.

transport/nats/request_response_funcs.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"context"
+
+	"github.com/nats-io/go-nats"
+)
+
+// RequestFunc may take information from a publisher request and put it into a
+// request context. In Subscribers, RequestFuncs are executed prior to invoking the
+// endpoint.
+type RequestFunc func(context.Context, *nats.Msg) context.Context
```
nussjustin: SubscriberRequestFunc

transport/nats/publisher.go
```diff
@@ -0,0 +1,138 @@
...
+	dec DecodeResponseFunc,
+	options ...PublisherOption,
+) *Publisher {
+	c := &Publisher{
+		publisher: publisher,
+		subject:   subject,
+		reply:     reply,
+		enc:     enc,
+		dec:     dec,
+		before:  []RequestFunc{},
```
nussjustin: No need to initialize this, Go already handles nil correct for all the cases needed.

transport/nats/subscriber.go
nussjustin: ctx should be cancelled when the handler returns

transport/nats/publisher.go
```diff
@@ -0,0 +1,138 @@
...
+			return response, nil
+
+		}
+
+		if p.reply == "" {
+			if err := p.publisher.Publish(msg.Subject, msg.Data); err != nil {
+				return nil, err
+			}
+
+			return nil, nil
```
nussjustin: The funcs in p.after should probably still be called here and below with a nil *nats.Msg

transport/nats/encode_decode.go
kirooha: oh, great! thanks a lot! I didn't hope to so fast review) Actually, it was a draft and I supposed to modify comments and test code. One more time, thanks a lot! I will fix your remarks as soon as possible.

kirooha:
@nussjustin all remarks are fixed

kirooha:
there's some weird behavior of ci in this pr. I did 3 force push without any changes (amend commit that added blank lines). 2 builds were failed and the latest is succeeded. I mean ci/circleci and continuous-integration/travis-ci/pr

kirooha:
@nussjustin tests have been added

kirooha:
@nussjustin fixed wrong paths, squashed commits to the only one with a more detailed comment. do you have any notes?

nussjustin:
Could you please add a test for PublisherTimeout? From what I see it's the only publisher option that's not tested right now (as reported by `go tool cover`).

Otherwise LGTM so far.

kirooha:
done

kirooha:
thank you! awesome note, fixed

PR #690, opened 2018-03-23
Add DisableEscapeHTML option for json logger

soh335:
Hi.
I add DisableEscapeHTML option for json logger. If pass ```log.DisableEscapeHTML(true)``` to NewJSONLogger, we can get more readability.

Thanks.

### DIFF
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -10,14 +10,29 @@ import (
 
 type jsonLogger struct {
 	io.Writer
+	disableEscapeHTML bool
 }
 
 // NewJSONLogger returns a Logger that encodes keyvals to the Writer as a
 // single JSON object. Each log event produces no more than one call to
 // w.Write. The passed Writer must be safe for concurrent use by multiple
 // goroutines if the returned Logger will be used concurrently.
-func NewJSONLogger(w io.Writer) Logger {
-	return &jsonLogger{w}
+func NewJSONLogger(w io.Writer, options ...JSONLoggerOption) Logger {
+	l := &jsonLogger{
+		Writer: w,
+	}
+	for _, option := range options {
+		option(l)
+	}
+	return l
+}
+
+// JSONLoggerOption sets a parameter for json logger
+type JSONLoggerOption func(*jsonLogger)
+
+// DisableEscapeHTML disable to escape &, <, >.
+func DisableEscapeHTML(v bool) JSONLoggerOption {
+	return func(j *jsonLogger) { j.disableEscapeHTML = v }
 }
 
 func (l *jsonLogger) Log(keyvals ...interface{}) error {
@@ -31,7 +46,9 @@ func (l *jsonLogger) Log(keyvals ...interface{}) error {
 		}
 		merge(m, k, v)
 	}
-	return json.NewEncoder(l.Writer).Encode(m)
+	enc := json.NewEncoder(l.Writer)
+	enc.SetEscapeHTML(!l.disableEscapeHTML)
+	return enc.Encode(m)
 }
 
 func merge(dst map[string]interface{}, k, v interface{}) {
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -10,7 +10,7 @@ import (
 
 type jsonLogger struct {
 	io.Writer
-	disableEscapeHTML bool
+	escapeHTML bool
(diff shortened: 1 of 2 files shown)


soh335:
oops. sorry I break ```func(io.Writer) log.Logger``` interface. 

I think there are some ideas.

* add like a ```NewJSONLoggerWithOption``` ( other constructor function dont contain WithOption. it seems to be bad)
* create Option interface and define constructor function as  ```func(w: io.Writer, options ...Options)``` ( need to change other loggers )
* call SetHTMLEscape(false) in jsonLogger without option. ( breaking change )
* other idea ?

ChrisHines:
My initial thought is to call SetHTMLEscape(false) in jsonLogger without option. I'm not sure how much that is really a breaking change. The output will still be JSON compliant. I wouldn't consider the browser as a primary target for the output of jsonLogger.

soh335:
>I wouldn't consider the browser as a primary target for the output of jsonLogger.

I think so too. so no one care about this change, call SetHTMLEscape(false) in jsonLogger seems to be better than others.

nicolaiskogheim:
`enc.SetEscapeHTML(!l.disableEscapeHTML)` (log/json_logger.go:50)

This would read better if the property was named `escapeHTML` instead, I think.

soh335:
@peterbourgon how about this ?

log/json_logger.go
```diff
@@ -10,14 +10,29 @@ import (
 
 type jsonLogger struct {
 	io.Writer
+	disableEscapeHTML bool
```
peterbourgon: This should be 

```go
escapeHTML bool
```

defaulting to false.

log/json_logger.go
```diff
@@ -10,14 +10,29 @@ import (
...
+		option(l)
+	}
+	return l
+}
+
+// JSONLoggerOption sets a parameter for json logger
+type JSONLoggerOption func(*jsonLogger)
+
+// DisableEscapeHTML disable to escape &, <, >.
+func DisableEscapeHTML(v bool) JSONLoggerOption {
```
peterbourgon: This should be 

```go
func SetEscapeHTML(on bool) JSONLoggerOption {
```

mirroring [json/Encoder.SetEscapeHTML](https://golang.org/pkg/encoding/json/#Encoder.SetEscapeHTML).

log/json_logger.go
```diff
@@ -31,7 +46,9 @@ func (l *jsonLogger) Log(keyvals ...interface{}) error {
 		}
 		merge(m, k, v)
 	}
-	return json.NewEncoder(l.Writer).Encode(m)
+	enc := json.NewEncoder(l.Writer)
+	enc.SetEscapeHTML(!l.disableEscapeHTML)
```
peterbourgon: This should be

```go
enc.SetEscapeHTML(l.escapeHTML)
```

soh335:
@peterbourgon thanks for reviewing and change to SetEscapeHTML.
/log/level/level_test.go and log/term/example_test.go is failed because log.NewJSONLogger is not ```func(io.Writer) log.Logger``` interface.

Should I change like this ? (I think its not smart...)

```diff
-                       format: log.NewJSONLogger,
+                       format: func(w io.Writer) log.Logger { return log.NewJSONLogger(w) },
```

```diff
-       logger := term.NewLogger(os.Stdout, log.NewJSONLogger, colorFn)
+       logger := term.NewLogger(os.Stdout, func(w io.Writer) log.Logger { return log.NewJSONLogger(w) }, colorFn)
```

soh335:
Is it mean call ```SetEscapeHTML(false)``` in json logger and call like ```html.EscapeString(v)``` function in HTMLEscapeLogger ?
I want to just call ```enc.SetEscapeHTML``` (its simple), But I know keep interface is also important...

soh335:
@ChrisHines @peterbourgon it seems to be difficult both to keep ```func(io.Writer) log.Logger```  and add logger options currently. And if dont allow to call enc.SetEscapeHTML without option in json logger, maybe this pr cant be merged. so I will close.

ChrisHines:
I haven't seen any objections to just changing the existing JSONLogger to always call `enc.SetEscapeHTML(false)` and leave the exported API untouched. I would be OK with that change.

PR #705, opened 2018-04-27
Fix race condition in NATS transport tests

nussjustin:
The tests for PR #680 have a small race that leads to sporadic test failures (https://travis-ci.org/go-kit/kit/jobs/372140473#L1732).

Closing subscriptions and connections is asynchronous, so when the next test runs and tries to receive messages there may still be an old connection subscribed to the same topic used by the next test. 

Since the tests are using queue subscriptions only one client will ever get a message, meaning that a message may never arrive on the newer connection.

This PR refactors the logic for opening and closing the connections to check for open subscriptions before opening new connections and, if there are any, waiting a bit before failing.

Other options are to use different topic or queue names or even not using queue subscriptions, but checking the server seemed safer to me and requires less changes to the tests.

### DIFF
--- a/transport/nats/publisher_test.go
+++ b/transport/nats/publisher_test.go
@@ -1,13 +1,13 @@
 package nats_test
 
 import (
-	"testing"
 	"context"
-	"time"
 	"strings"
+	"testing"
+	"time"
 
-	"github.com/nats-io/go-nats"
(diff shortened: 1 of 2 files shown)


peterbourgon:
Hehe thanks

PR #708, opened 2018-05-01
Update README.md with English version of website

Olliepop:
Because the rest of the readme is in English, it makes sense that the *Additional reading* section also be English, if possible.


peterbourgon:
Huh, didn't notice that. Thanks!

PR #712, opened 2018-05-10
fix for etcdv3 client panic when only using sd.Registrar

basvanbeek:
When using etcdv3 client on a service only registering and deregistering itself, the client would panic on close. 

This fixes the issue and adds an integration test for testing this use case.


basvanbeek:
seriously can't figure out what is wrong with the test on tip :(  the test output gives no clues and running on tip locally succeeds

PR #715, opened 2018-05-30
Extend an existing contextual logger rather than replacing it

cvanderschuere:
The current documentation states that 
> If logger is also a contextual logger created by With or WithPrefix, keyvals is appended to the existing context.

Prior to this change, that was not actually the case and a contextual logger was being overwritten every time.


cvanderschuere:
This does not do what I had intended it to

PR #719, opened 2018-06-11
Tracing middlewares for OpenCensus

basvanbeek:
This PR adds tracing middlewares for OpenCensus.

It contains custom handlers for Go kit gRPC and HTTP Server and Client transports as well as a Go kit endpoint middleware.

Included in this PR is the codification of the Failer interface in the endpoint package. It's use was already demonstrated in the addsvc package but now made part of the endpoint package as it now is used in middlewares like OpenCensus to be able to inspect if a response type value holds a business error.

Finally some small fixes to integration tests for Consul and Zookeeper to have the integration tests be skipped if the set-up environment variables are not there. This to bring it in line with the other SD packages.


sd/zk/integration_test.go
```diff
@@ -18,13 +17,16 @@ var (
 
 func TestMain(m *testing.M) {
 	zkAddr := os.Getenv("ZK_ADDR")
-	if zkAddr == "" {
-		log.Fatal("ZK_ADDR is not set")
+	if zkAddr != "" {
+		host = []string{zkAddr}
 	}
-	host = []string{zkAddr}
+	m.Run()
```
peterbourgon: This subverts the intent, which is to explicitly not run the test unless ZK_ADDR is set.

endpoint/endpoint.go
```diff
@@ -26,3 +26,17 @@ func Chain(outer Middleware, others ...Middleware) Middleware {
...
+// implement Failer.
+// Response encoders can check if responses are Failer, and if so if they've
+// failed encode them using a separate write path based on the error.
+// Endpoint middlewares can test if a response type failed and also act or
+// report upon it.
+//
+// The addsvc example shows Failer's intended usage.
+type Failer interface {
+	Failed() error
+}
```
peterbourgon: Something rings... _weird_ to me about the names Failer/Failed, but I've thought about it for awhile and can't come up with anything better. Can everyone reading this PR have a brainstorm for a minute or two? It may be this is the best option, but I'd like to explore the space a little bit, collectively...

endpoint/endpoint.go
```diff
@@ -26,3 +26,17 @@ func Chain(outer Middleware, others ...Middleware) Middleware {
...
+// Failer is an interface that should be implemented by response types that
+// hold error properties as to separate business errors from transport errors.
+// If the response type can hold business errors it is highly advised to
+// implement Failer.
+// Response encoders can check if responses are Failer, and if so if they've
+// failed encode them using a separate write path based on the error.
+// Endpoint middlewares can test if a response type failed and also act or
+// report upon it.
+//
+// The addsvc example shows Failer's intended usage.
```
peterbourgon: This could be cleaned up.

> Failer may be implemented by Go kit response types that contain business logic error details. If Failed returns a non-nil error, the Go kit transport layer may interpret this as a business logic error, and may encode it differently than a regular, successful response.
>
> It's not necessary for your response types to implement Failer, but it may help for more sophisticated use cases. The addsvc example shows how Failer should be used by a complete application.

sd/zk/integration_test.go
peterbourgon: I understand this now.

tracing/opencensus/endpoint_test.go
```diff
@@ -0,0 +1,159 @@
...
+	span4 = "SPAN-4"
+	span5 = "SPAN-5"
+)
+
+var (
+	err1 = errors.New("someError")
+	err2 = errors.New("otherError")
+	err3 = errors.New("someBusinessError")
+	err4 = errors.New("otherBusinessError")
+)
```
peterbourgon: Super minor nit: can these be more canonical, e.g. `errors.New("some error")`?

tracing/opencensus/grpc.go
```diff
@@ -0,0 +1,156 @@
...
+				span.End()
+			}
+		},
+	)
+
+	return func(c *kitgrpc.Client) {
+		clientBefore(c)
+		clientFinalizer(c)
+	}
+
```
peterbourgon: Minor nit: extraneous newline.

tracing/opencensus/http.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return func(c *kithttp.Client) {
+		clientBefore(c)
+		clientAfter(c)
+		clientFinalizer(c)
+	}
+}
+
+// HTTPServerTrace enables OpenCensus tracing of a Go kit HTTP transport server.
+func HTTPServerTrace(options ...TracerOption) kithttp.ServerOption {
```
peterbourgon: Really cool use of transport/http.ServerOption, lots of stuff packed in to these functions.

tracing/opencensus/http.go
```diff
@@ -0,0 +1,174 @@
...
+	)
+
+	serverFinalizer := kithttp.ServerFinalizer(
+		func(ctx context.Context, code int, r *http.Request) {
+			if span := trace.FromContext(ctx); span != nil {
+				span.SetStatus(ochttp.TraceStatus(code, http.StatusText(code)))
+
+				if rs, ok := ctx.Value(kithttp.ContextKeyResponseSize).(int64); ok {
+					span.AddAttributes(
+						trace.Int64Attribute("http.response_size", rs),
```
peterbourgon: Why is response size "special" in this way? Can other attributes be special, too?

tracing/opencensus/grpc.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+			return ctx
+		},
+	)
+
+	serverFinalizer := kitgrpc.ServerFinalizer(
+		func(ctx context.Context, err error) {
+			if span := trace.FromContext(ctx); span != nil {
+				s, ok := status.FromError(err)
+				if ok {
```
peterbourgon: Nit: combined `if` expr is possible here.

```go
if s, ok := status.FromError(err); ok {
    // ...
} else { 
    // ...
}
```

tracing/opencensus/grpc.go
```diff
@@ -0,0 +1,156 @@
...
+
+	for _, option := range options {
+		option(&cfg)
+	}
+
+	serverBefore := kitgrpc.ServerBefore(
+		func(ctx context.Context, md metadata.MD) context.Context {
+			var (
+				spanContext trace.SpanContext
+				ok          bool
```
peterbourgon: Nit: given the distance between declaration and use, can this have a slightly more descriptive name?

tracing/opencensus/grpc.go
```diff
@@ -0,0 +1,156 @@
...
+
+			return trace.NewContext(ctx, span)
+		},
+	)
+
+	clientFinalizer := kitgrpc.ClientFinalizer(
+		func(ctx context.Context, err error) {
+			if span := trace.FromContext(ctx); span != nil {
+				s, ok := status.FromError(err)
+				if ok {
```
peterbourgon: Nit: combined `if` expr is possible here.

```go
if s, ok := status.FromError(err); ok {
    // ...
} else { 
    // ...
}
```

tracing/opencensus/http_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"go.opencensus.io/plugin/ochttp/propagation/tracecontext"
+	"go.opencensus.io/trace"
+	"go.opencensus.io/trace/propagation"
+
+	"github.com/go-kit/kit/endpoint"
+	ockit "github.com/go-kit/kit/tracing/opencensus"
+	kithttp "github.com/go-kit/kit/transport/http"
+)
+
+func TestHttpClientTrace(t *testing.T) {
```
peterbourgon: Minor nit: TestHTTP

tracing/opencensus/http_test.go
```diff
@@ -0,0 +1,178 @@
...
+		name string
+		err  error
+	}{
+		{"", nil},
+		{"CustomName", nil},
+		{"", errors.New("dummy-error")},
+	}
+
+	for _, tr := range traces {
+
```
peterbourgon: Minor nit: extraneous newline.

tracing/opencensus/http.go
basvanbeek: I'm not sure what you're noticing as special here. Can you explain?

peterbourgon: Sorry, that _was_ a bit obtuse. I guess your code here responds to [these lines of code here](https://github.com/go-kit/kit/blob/master/transport/http/server.go#L90-L91), and I had mistakenly thought we were adding more to the context than we actually were. Disregard.

tracing/opencensus/endpoint_options.go
```diff
@@ -0,0 +1,42 @@
+package opencensus
+
+import "go.opencensus.io/trace"
+
+// EndpointOptions holds the options for tracing an endpoint
+type EndpointOptions struct {
+	// IgnoreBusinessError if set to true will not treat a business error
+	// identified trough the endpoint.Failer interface as a span error.
```
odeke-em: Minor nit: s/trough/through/g

tracing/opencensus/grpc.go
```diff
@@ -0,0 +1,155 @@
...
+
+	kitgrpc "github.com/go-kit/kit/transport/grpc"
+)
+
+const propagationKey = "grpc-trace-bin"
+
+// GRPCClientTrace enables OpenCensus tracing of a Go kit gRPC transport client.
+func GRPCClientTrace(options ...TracerOption) kitgrpc.ClientOption {
+	cfg := TracerOptions{
+		sampler: trace.AlwaysSample(),
```
odeke-em: By default each trace is always being sampled, perhaps let's put that as a comment in the doc right above?
```go
// By default each trace is sampled, however you can change this behavior
// by using the TracerOption WithSampler
```
or something like that

tracing/opencensus/http.go
```diff
@@ -0,0 +1,174 @@
...
+		clientAfter(c)
+		clientFinalizer(c)
+	}
+}
+
+// HTTPServerTrace enables OpenCensus tracing of a Go kit HTTP transport server.
+func HTTPServerTrace(options ...TracerOption) kithttp.ServerOption {
+	cfg := TracerOptions{
+		sampler:       trace.AlwaysSample(),
+		httpPropagate: &b3.HTTPFormat{},
```
odeke-em: Unless we'll always be using Zipkin/B3 propagation, could we add a TracerOption for httpPropagation? And provide a comment that it is useful only for
HTTPPropagation?

basvanbeek: We do have an option for it... Just found out I do need to change something on the TracerOptions struct... the items need to be exported for WithTracerConfig to make any sense.

odeke-em: Never mind, sorry for the noise. I hadn't seen the WithHTTPPropagation.

basvanbeek:
Nits have been addressed and PR has been rebased. Only open item is our discussion on Failed/Failer.

Do we have any better ideas for naming or should we proceed with the existing one even though it feels imperfect.

peterbourgon:
I'm convinced there's a better name for this, but I've thought about it extensively, and I can't come up with one. Let's go ahead.

basvanbeek:
I'll open an issue for the Failed interface nomenclature problem.

PR #720, opened 2018-06-11
fix pact test due to breaking api change

basvanbeek:

PR #721, opened 2018-06-16
Remove redundant statements in profilesvc example

sudo-suhas:
In the profilesvc example, both the request method and URL path are set in each specific encoder. This is not required for the request method as that is specified correctly in the call to `httptransport.NewClient`. This PR removes the redundant statements.


sudo-suhas:
CI failures are note related to changes made here, see #720.

peterbourgon:
Can you rebase on master?

sudo-suhas:
Done.

sudo-suhas:
Amended the commit message as it was not getting formatted correctly:

image

PR #723, opened 2018-06-18
Fix function names in JWT examples

zachwalton:
Small docs change to fix incorrect function names


peterbourgon:
Ah great, thanks.

PR #724, opened 2018-06-19
brought etcdv3 registrar only test to its own run

basvanbeek:
The etcdv3 test sometimes fails on CircleCI. Hopefully this fixes the problem. It is a cleaner way of testing anyway.


basvanbeek:
Failing on the Eureka test now :(

PR #730, opened 2018-07-16
Adds documentation for OpenCensus in tracing readme and package doc

basvanbeek:
Noticed OpenCensus did not have a package doc nor a section in our tracing readme.


tracing/README.md
```diff
@@ -39,6 +39,15 @@ change in configuration.
 The following tracing back-ends are known to work with Go kit through the
 OpenTracing interface and are highlighted in the [addsvc] example.
 
+## OpenCensus
```
yurishkuro: looks like the wrong place in the readme, doesn't work with the previous sentence.

basvanbeek: you are right, moving the section

PR #734, opened 2018-07-26
Moves zipkin-go-opentracing to openzipkin-contrib.

jcchavezs:
`zipkin-go-opentracing` was moved to `openzipkin-contrib`.


basvanbeek:
thanks @jcchavezs

PR #742, opened 2018-08-03
jsonrpc add id from request to response

wangzuo:
According to https://www.jsonrpc.org/specification:

> **id**
> This member is REQUIRED.
> It MUST be the same as the value of the id member in the Request Object.
> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it > MUST be Null.

This pull request add req.ID to response object.

### DIFF
--- a/transport/http/jsonrpc/request_response_types.go
+++ b/transport/http/jsonrpc/request_response_types.go
@@ -35,6 +35,16 @@ func (id *RequestID) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (id *RequestID) MarshalJSON() ([]byte, error) {
+	if id.intError == nil {
+		return json.Marshal(id.intValue)
+	} else if id.floatError == nil {
+		return json.Marshal(id.floatValue)
+	} else {
+		return json.Marshal(id.stringValue)
+	}
+}
+
 // Int returns the ID as an integer value.
 // An error is returned if the ID can't be treated as an int.
 func (id *RequestID) Int() (int, error) {
@@ -59,6 +69,7 @@ type Response struct {
 	JSONRPC string          `json:"jsonrpc"`
 	Result  json.RawMessage `json:"result,omitempty"`
 	Error   *Error          `json:"error,omitempty"`
+	ID      *RequestID      `json:"id"`
 }
 
 const (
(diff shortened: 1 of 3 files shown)


wangzuo:
Not sure why the tests failed

transport/http/jsonrpc/request_response_types_test.go
```diff
@@ -109,3 +109,43 @@ func TestCanUnmarshalNullID(t *testing.T) {
...
+		want := JSON
+		bol, _ := json.Marshal(resp)
+		got := string(bol)
+		if got != want {
+			t.Fatalf("'%s': want %s, got %s.", c.expType, want, got)
+		}
+	}
+}
+
+func TestCanMarshalNullID(t *testing.T) {
```
basvanbeek: why a separate test for null? wouldn't this fit the TestCanMarshalID test (adding a case for it)

wangzuo: test updated

basvanbeek:
@peterbourgon can you review? i’ll be offline the next 2 weeks.

transport/http/jsonrpc/request_response_types.go
```diff
@@ -35,6 +35,16 @@ func (id *RequestID) UnmarshalJSON(b []byte) error {
...
 
+func (id *RequestID) MarshalJSON() ([]byte, error) {
+	if id.intError == nil {
+		return json.Marshal(id.intValue)
+	} else if id.floatError == nil {
+		return json.Marshal(id.floatValue)
+	} else {
+		return json.Marshal(id.stringValue)
+	}
+}
```
peterbourgon: My sense is this would be better expressed without the short-circuit logic, as

```go
switch {
case id.intError == nil && id.floatError != nil:
    return json.Marshal(id.intValue)
case id.intError != nil && id.floatError == nil:
    return json.Marshal(id.floatValue)
default:
    return json.Marshal(id.stringValue)
}
```

What do you think?

wangzuo: I dont think so, intError and floatError can both be nil

peterbourgon: Yes, if intErr == nil and floatErr == nil, the final default case is triggered.

wangzuo: For integer value, intErr and floatErr are nil and the default case will return empty string.

peterbourgon: Ah, OK, I understand. Then the original implementation is best.

peterbourgon:
Thank you!

transport/http/jsonrpc/request_response_types.go
wangzuo: Awesome!

Issue #681, opened 2018-03-16
Question: AMQP transport implementation?


Vetcher:
Hey, I would like to add amqp transport with an eye on #680 and help of https://github.com/streadway/amqp
Is it OK idea?

Issue #295, opened 2016-06-15
Discussion: package pubsub


peterbourgon:
I'd like to start thinking about a design for package pubsub/msgbus/...

Important use cases:
- An async transport for existing/typical RPC messaging
- A foundation for [event sourcing](http://microservices.io/patterns/data/event-sourcing.html)
- A way to manage [streaming](https://github.com/go-kit/kit/issues/213)
- Others?

Would like support for:
- Kafka
- RabbitMQ
- NATS
- NSQ
- AWS SQS
- GCP PubSub
- Weave Mesh
- Others?

Prior art: 
- [NYTimes/gizmo/pubsub](https://godoc.org/github.com/NYTimes/gizmo/pubsub)
- [micro/go-micro/broker](https://godoc.org/github.com/micro/go-micro/broker) 
- [Micro on NATS](https://blog.micro.mu/2016/04/11/micro-on-nats.html)
- Others?

Open questions:
- What are the core abstractions and interfaces?
- Is streaming significantly different than event sourcing and transport use cases?
- What is the lowest common denominator for serialization?
- Others?

basvanbeek:
Shall we start a document for this or are you going to edit the above message to keep track of state?

intellidynamics:
Peter,

 

Task-oriented message passing architectures are great for flexibility, wiring together functional blocks or micro-services through subscriptions.  These tasks send/route messages of a negotiated format with identified payloads through a mesh of subscriptions, adding value as the information flows and routing onward to downstream subscribers.  We’d love to use Go Kit to do this, else we’ll write our own, it’s not too hard really.  We’ve been doing this successfully for 16 years.  Tasks/services have a collection of subscribers and a subscription object is relatively simple, upon which events are raised containing the transmitted message with its envelope information and payload.  Tasks, or a supervisory agent, can place their subscription object into the collection of other objects and begin receiving.  They can unsubscribe by removing the subscription object from the message sender’s collection.  

 

I would avoid a “bus” concept (centralized pipe) but use independent synchronous / asynchronous connections between tasks or services.

 

I would seek to keep the pub/sub as a transport that is independent of what is being exchanged as much as possible.  What the message is, it’s meaning, its format is the snarly part, so I’d avoid mingling it into the transport.   I think it should support fan-out (multiple subscribers) at a minimum and fan-in is sometimes needed too (use case: synchronizing data sets coming from different sources).  

 

[...]

PR #746, opened 2018-08-07
AMQP Transport 

mattfung:
Hello, 

I noticed that here had been interest in *AMQP* transport similiar to NATS (#681) as well as talks of pubsub support (#295) so I thought I might like to share an implementation of a simple AMQP client & server transport.

Comments/Reviews would be appreciated.

### DIFF
--- a/transport/amqp/doc.go
+++ b/transport/amqp/doc.go
@@ -1,0 +1,2 @@
+// Package amqp implements an AMQP transport.
+package amqp
--- a/transport/amqp/encode-decode.go
+++ b/transport/amqp/encode-decode.go
@@ -1,0 +1,22 @@
+package amqp
+
+import (
+	"context"
+	"github.com/streadway/amqp"
+)
+
+// DecodeRequestFunc extracts a user-domain request object from
+// an AMQP Delivery object. It is designed to be used in AMQP Subscribers.
+type DecodeRequestFunc func(context.Context, *amqp.Delivery) (request interface{}, err error)
+
+// EncodeRequestFunc encodes the passed request object into
+// an AMQP Publishing object. It is designed to be used in AMQP Publishers.
+type EncodeRequestFunc func(context.Context, *amqp.Publishing, interface{}) error
+
+// EncodeResponseFunc encodes the passed reponse object to
+// an AMQP Publishing object. It is designed to be used in AMQP Subscribers.
+type EncodeResponseFunc func(context.Context, *amqp.Publishing, interface{}) error
+
+// DecodeResponseFunc extracts a user-domain response object from
+// an AMQP Delivery object. It is designed to be used in AMQP Publishers.
+type DecodeResponseFunc func(context.Context, *amqp.Delivery) (response interface{}, err error)
--- a/transport/amqp/publisher.go
+++ b/transport/amqp/publisher.go
@@ -1,0 +1,152 @@
+package amqp
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/streadway/amqp"
+)
+
+// The golang AMQP implementation requires the []byte representation of
+// correlation id strings to have a maximum length of 255 bytes.
+const maxCorrelationIdLength = 255
+
+// Publisher wraps an AMQP channel and queue, and provides a method that
+// implements endpoint.Endpoint.
+type Publisher struct {
+	ch      Channel
+	q       *amqp.Queue
+	enc     EncodeRequestFunc
+	dec     DecodeResponseFunc
+	before  []RequestFunc
+	after   []PublisherResponseFunc
+	timeout time.Duration
+}
+
+// NewPublisher constructs a usable Publisher for a single remote method.
+func NewPublisher(
+	ch Channel,
+	q *amqp.Queue,
+	enc EncodeRequestFunc,
+	dec DecodeResponseFunc,
+	options ...PublisherOption,
+) *Publisher {
+	p := &Publisher{
+		ch:      ch,
+		q:       q,
+		enc:     enc,
+		dec:     dec,
+		timeout: 10 * time.Second,
+	}
+	for _, option := range options {
+		option(p)
+	}
+	return p
+}
+
+// PublisherOption sets an optional parameter for clients.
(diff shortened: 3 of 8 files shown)


peterbourgon:
Cool, this looks reasonable at a first glance. As a next step, please run `go vet`, `golint`, and `staticcheck` and fix the relevant errors. A couple of problems that stood out to me: underscores in variable names, and superfluous newlines in function declarations.

mattfung:
May bad with the underscores.. bad python habits.
The problems should be fixed now, lemme know if anything can be improved

jstoja:
Hey @peterbourgon @mattfung any chance that this moves forward? This would be a great addition along NATS!

jstoja:
@mattfung I had some issues with running the tests on my end. Rebasing solve all the tests, would you be able to do that?

mattfung:
make a PR to https://github.com/streadway/amqp with channel listener functionality :)

jokes aside, Id be interested on improvements

jstoja:
Haha yes indeed!
Let’s think about that in another issue ;)

It think this is definitely mergeable! Great job!

jstoja:
@peterbourgon Do you have any feedback on this for merging it? I've used it for a professional project as a demonstration (so with small usage), and it ran several days without any issues.

peterbourgon:
That’s great feedback, thanks. I just need to block off an hour to review.
I’ll do that in the next 2 days. Sorry for the delay; I have a lot on my
plate these days :/

jstoja:
No worries :) just trying to know if help was needed on that!

mattfung:
Sounds good, I'll get them fixed within a day's time.
thx again for putting in the time to do this

mattfung:
Committed the changes. I'm guessing the failed tests occurred because I didn't rebase the code with the latest `master` branch?

Anyhow, let me know if anything needs to be changed.

peterbourgon:
Yes, this seems fine; can you rebase on master?

peterbourgon:
Dang nice

TLogik:
@mattfung I am new to gokit and struggling a bit getting AMQP up and working.
Besides the test do you have a sample app or something that shows how to use the AMQP setup.
I am starting prototyping moving some of out NET services to GO and we rely heavily on RabbitMQ.
Any help is very much appreciated.

mattfung:
@TLogik  for now, heres an implementation of stringsvc4 pub & sub
https://github.com/cloudpbx/kit/tree/amqp/examples/stringsvc4/amqp

Hope this helps.

jstoja:
Awesome! Would be great to have it in this repository too, no?

mattfung:
Thats the plan, my only reservation currently is that the commit makes some breaking changes with the current `stringsvc4` implementation. When I resolve that, I'll create a new merge request.

TLogik:
> Hope this helps.

It helped a lot. Much appreciated.

TLogik:
Hi again @mattfung 
I am using your example as a template, trying to get the behaviour i need.
Questions arise as i work with the implementation.
I find that in the Subscriber.go, func ServeDelivery, the channel for consuming is also used for publishing.
That is somewhat ok, except for at common requirement for subscribing from one Vhost and publishing to another Vhost.
AFAIK you need to create a new connection to each Vhost, create individual channels etc.

i do not seem to be able to setup the internal publisher, in the subscriber.go, for an alternate channel - can you confirm that is not possible in the current implementation?

mattfung:
@TLogik correct, current implementation does not allow for a subscriber to subscribe to one channel (& consequently, Vhost) and publish to another. I am curious on how other frameworks handle this problem if you happen to know of any.

Currently, publishers and subscribers are (for simplicity's sake) tied to one channel. That can change if it is a significant limitation. 

An alternative approach would be to have a subscriber for vhost A trigger a publisher for vhost B.

TLogik:
@mattfung currently we use individual instances of subscriber and publisher with individual connstrings.
I was looking into using go-kit as the toolbox of choice to create a standard to metrics, health, service registration etc, but it is a little hard to extend the transport layer.
i am currently working on a service that needs to batch process messages consumed from MQ and then inserted into SOLR. It is a lot faster to insert in SOLR via batching than individual requests.
This will require a custom amqp implementation to support this.

freemanoid:
@mattfung thanks for the amqp implementation, I've adopted it and it works perfectly! I [see][1] that consuming function automatically publishes some "response". It's not needed for me and I don't see any possibility to turn it off. Can you explain why did you added it? Maybe I'm using amqp wrong.

[1]: https://github.com/go-kit/kit/commit/12210fb6ace19e0496167bb3e667dcd91fa9f69b#diff-8767a31b727acc5ec358b2e23985c87fR114

peterbourgon:
If you're not publishing responses, you're not doing RPC.

freemanoid:
> If you're not publishing responses, you're not doing RPC.

Do you mean this implementation is not about AMQP, but RPC over AMQP?

mattfung:
@freemanoid  Good point, one way AMQP communication is not uncommon. Currently the closest thing to a "no-op" is to use the [NopEncoder](https://github.com/go-kit/kit/blob/8a8a1bb8700078c22421bc35bdcfb140b57c1b3a/transport/amqp/subscriber.go#L163-L170) such that an empty response is sent back.

There are two solutions to implement true "no-op" subscribers, one is to make the "encoder" both encode and send, and use `EncodeNopResponse` (though Im not sure if thats a [good idea](https://github.com/go-kit/kit/pull/746#discussion_r230945887) )

The other option is to add an options/ change the parameter to `NewSubscriber`

peterbourgon:
@freemanoid 

> Do you mean this implementation is not about AMQP, but RPC over AMQP?

Yes. Go kit is concerned exclusively with RPC.

> **Non-goals**
> - Supporting messaging patterns other than RPC (for now) — e.g. MPI, pub/sub, CQRS, etc.

---

@mattfung

> The other option is to add an options/ change the parameter to `NewSubscriber`

For the record, if this is easy, I'd happily take a PR.

madshov:
@mattfung Do you plan any work around adding an option/change parameter to NewSubscriber to no-op subscribers?

mattfung:
@madshov sorry for the slow response, PR is posted #850

PR #755, opened 2018-08-30
Remove dependency on github.com/go-stack/stack

ChrisHines:
The main reason we used github.com/go-stack/stack was because it handled several subtle idiosyncrasies about acquiring correct stack frame information from Go's runtime package that existed until Go 1.7. We don't need it for that anymore.

go-stack/stack also provides some nice features to help format stack frame information that we don't expose in the log package. The `log.Valuer` facility provides a mechanism for others to use those features of go-stack/stack if they want to, but I don't think there is a compelling reason to use go-stack/stack to implement `log.Caller` anymore. One less external dependency without any loss of features is nice.

### DIFF
--- a/log/value.go
+++ b/log/value.go
@@ -1,9 +1,10 @@
 package log
 
 import (
+	"runtime"
+	"strconv"
+	"strings"
 	"time"
-
-	"github.com/go-stack/stack"
 )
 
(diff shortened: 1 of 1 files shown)


peterbourgon:
Ah, can you rebase on master?

ChrisHines:
Sure, I can rebase later tonight.

Issue #752, opened 2018-08-16
Logrus adapter for Logger


AlmogBaku:
I suggest we'll provide a Logrus adapter for our log interface.

This adapter will mainly be used to support features of gokit, i.e. "gokit's http-server logging"

AlmogBaku:
simple example:
```
type serverLogger struct {
	*logrus.Entry
}
func (l serverLogger) Log(keyvals ...interface{}) error {
	if len(keyvals)%2 == 0 {
		logger := l.Entry
		for i:=0; i<len(keyvals); i+=2 {
			logger = logger.WithField(spew.Sdump(keyvals[i]), keyvals[i+1])
		}
		logger.Error()
	} else {
		l.Error(keyvals)
	}
	return nil
}
```

aswinmprabhu:
I am new to Golang and this is my attempt at this issue. Any feedback would be appreciated.

AlmogBaku:
In the meanwhile, you can use the code I attached above. I'll send a PR later on

aswinmprabhu:
Shall I write tests and make a PR?

AlmogBaku:
In our use-case, we decided to use `logrus` organization-wide. However, we using some components of go-kit(such as `kithttp.NewServer`) which require a `kit.logger`.

I think it would make sense to provide as a part of the kit an adapter for logrus (like we have adapters for other stuff here)

dolum:
go-kit http server only requires for a logger to implement this interface.
```
type Logger interface {
	Log(keyvals ...interface{}) error
}
```
You can use any struct you want if it has ```Log(keyvals ...interface{}) error```

AlmogBaku:
I'm not suggesting to use Logrus, but to provide an adapter to our logger.
See the code here I attached above as a reference.

ChrisHines:
I am willing to review a PR that adds a package with a logrus adapter. Suggestions for the package path:

* github.com/go-kit/kit/log/logrus
* github.com/go-kit/kit/log/kitlogrus

aswinmprabhu:
I will make a PR after writing the tests

aswinmprabhu:
#759

PR #759, opened 2018-09-05
Add logrus adapter for Logger interface (#752)

aswinmprabhu:


log/logrus_logger.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+type logrusLogger struct {
+	*logrus.Logger
+}
+
+// NewLogrusLogger returns a logger that logs the keyvals to Writer
+// with a time stamp at the logrus.InfoLevel
+func NewLogrusLogger(w io.Writer) Logger {
+	newLogger := logrus.New()
```
ChrisHines: I believe the main use case for this package is to adapt a `*logrus.Logger` used as the main logger in an application so it can be passed to code that expects a Go kit `log.Logger`. For that use case this constructor would be more flexible if it were declared as `NewLogrusLogger(logger *logrus.Logger) log.Logger`.

This form allows the main application to configure logrus however it wants and then simply adapt it to the Go kit `log.Logger` interface.

log/logrus_logger.go
```diff
@@ -0,0 +1,34 @@
+package log
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/sirupsen/logrus"
```
ChrisHines: We don't want the base Go kit `log` package to have a dependency on logrus. The logrus adapter should be in its own package in a subdirectory. I suggest it go in `github.com/go-kit/kit/log/logrus`.

log/logrus/logrus_logger.go
```diff
@@ -0,0 +1,31 @@
+package log
```
ChrisHines: Now that the code is moved into the log/logrus package, the package name here should also be logrus.

log/logrus/logrus_logger.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/sirupsen/logrus"
+)
+
+type logrusLogger struct {
+	*logrus.Logger
+}
+
+// NewLogrusLogger takes a *logrus.Logger and returns
+//a logger that stisfies the go-kit log.Logger interface
```
ChrisHines: - Missing a space after the `//` and a period (.) at the end of the sentence.
- s/stisfies/satisfies

log/logrus/logrus_logger.go
```diff
@@ -0,0 +1,31 @@
...
+func NewLogrusLogger(logger *logrus.Logger) log.Logger {
+	return &logrusLogger{logger}
+}
+
+func (l logrusLogger) Log(keyvals ...interface{}) error {
+	if len(keyvals)%2 == 0 {
+		fields := logrus.Fields{}
+		for i := 0; i < len(keyvals); i += 2 {
+			fields[fmt.Sprint(keyvals[i])] = keyvals[i+1]
```
ChrisHines: `keyvals[i+1]` will panic if `len(keyvals)` is odd. Use `nil` as the field value instead.

ChrisHines: I see that the `if len(keyvals)%2 == 2` protects against the panic now, but why does the `else` clause just call `l.Info(keyvals)` in that case?

log/logrus/logrus_logger.go
ChrisHines: Please add at least a sentence of package documentation.

log/logrus/logrus_logger.go
aswinmprabhu: l.Info(keyvals) just prints out the arguments as values without keys.
Eg - 
logger.Log("hello", "world", "hello") gives
time="13-09-2018 18:23:14" level=info msg="[hello world hello]"

log/logrus/logrus_logger.go
aswinmprabhu: Can you please clarify on what you mean by package documentation? I looked at other packages like endpoint.go and metrics.go and didn't see anything other that function and type docs.

log/logrus/logrus_logger.go
ChrisHines: That approach is not compatible with the `log.Logger` interface defined by Go Kit:

>Logger is the fundamental interface for all log operations. Log creates a log event from keyvals, a variadic sequence of alternating keys and values.

We don't want to encourage people to use it without keys because that will be incompatible with other implementations of the interface.

Please look at how the `jsonLogger` handles this situation here: https://github.com/go-kit/kit/blob/master/log/json_logger.go#L23.

log/logrus/logrus_logger.go
```diff
@@ -1,6 +1,9 @@
-package log
+// Package logrus package provides an adapter to the
+// go-kit log.Logger interface.
```
ChrisHines: The second instance of the word "package" is redundant, please remove.

log/logrus/logrus_logger.go
```diff
@@ -11,21 +14,23 @@ type logrusLogger struct {
 	*logrus.Logger
 }
 
+var errMissingValue = errors.New("(MISSING)")
+
 // NewLogrusLogger takes a *logrus.Logger and returns
-//a logger that stisfies the go-kit log.Logger interface
+// a logger that satisfies the go-kit log.Logger interface.
```
ChrisHines: Consider this alternate wording:

```go
```

PR #761, opened 2018-09-13
Add context.Context support for stringsvc2 example to actually compile

bx2:


peterbourgon:
Would you mind making a PR to https://github.com/go-kit/go-kit.github.io ?

bx2:
sure, would you like to have it with a context in the service or without?

(In the first version of the `stringsvc` the context is not used at all. In the 2nd version it is removed completely - in the service.go)

peterbourgon:
Since "Threading a context" is listed as an advanced concept in the docs, it was probably a mistake that it was added at all to stringsvc1. Ideally, it should be removed from all docs and code, stringsvc1 through stringsvc3.

bx2:
ok, will do.

PR #767, opened 2018-10-01
metrics/internal/convert: use method value insetad of lambda

Quasilyte:
Method value signature matches expected callback signature,
so there is no need in explicit lambda wrapping.


Quasilyte:
CI failure looks unrelated to the change:
```
#!/bin/bash -eo pipefail
go get -t github.com/go-kit/kit/...

# github.com/miekg/dns
../../miekg/dns/msg_helpers.go:271:8: undefined: strings.Builder
../../miekg/dns/serve_mux.go:43:9: undefined: strings.Builder
Exited with code 2
```

PR #768, opened 2018-10-01
sd/lb: replace len(x)<=0 with len(x)==0

Quasilyte:
Len by definition never returns negative values.
Make check more precise and remove redundant uncertainty
from the code.


peterbourgon:
Sorry, old habits die hard.

PR #776, opened 2018-10-18
Fix 'Client.Endpoint' to not 'cancel' when bufferedStream

### LINKED ISSUE #773: 'Client.Endpoint()'  returns 'context canceled' while reading the Body
Hello! I have a question which IDK if it's a missus on my part and there is another way to do it.

I'm writing a client which consumes a MS that returns binary data. I'm using the `kithttp.BufferedStream(true)` while `NewClient` to make sure the body is not closed so I can read it when I want and not have it buffered.

This works fine with my implementation of the `decode*` that I use on the `NewClient` except if the body is longer than `3966` which then stops reading from the body.

This is my `decode*`:

```go
func decode(_ context.Context, r *http.Response) (interface{}, error) {
        pr, pw := io.Pipe()

        go func() {
                defer r.Body.Close()
                defer pw.Close()
                i, err := io.Copy(pw, r.Body)
                fmt.Println(i, err)
        }()
        // IOR is a io.Reader
        return response{IOR: pr}, nil
}
```

The intention is to only read from the `r.Body` when I need it, that's why I'm using a `io.Pipe`, and also because I have to `r.Body.Close()` when it ends (if not, just passing the `r.Body` to the `response.IOR` would work, but then the `r.Body` wold not closed, no?).

This code works for "small" `r.Body` but when the it's bigger than `3966` the `io.Copy` stops and the output form the print is `3966 context canceled`. So basically the https://github.com/go-kit/kit/blob/4914a6960de6e1a81504d2b8a9dd9f207d20c757/transport/http/client.go#L95 is called and the body lost.

Maybe I'm overcomplicating it using the `io.Pipe` but I thought it was the best solution for what I had in mind.

Maybe the `defer cancel()` should be called once the `r.Body.Close()` is called as it's I have the `BufferedStream` option no? IDK if that's even possible haha.

Thanks!

xescugc:
Fixes #773 by adding a wrapper to the `Resonse.Body` when the `bufferedStream` is specified.

For now it's a PoC as it may change.

The biggest **drawback** is that old users of `bufferedStream` that where not closing the `Resopnse.Body` may end up with an open connection :cry: . The only **solution** I can have for that is to have another option to tell the client that the user will close the connection, so all this logic I did would be placed in an if condition of that other option (which maybe cleaner, but with the bug on `bufferedStream` will still be there).

To make the test fail I had to fake a "big" body (`string(make([]byte, 6000))`) and "fake work" (`time.Sleep`), then those where fixed with the implementation.

Depending on how does this evolve, I could include the [jsonrpc.Client.Endpoint](https://github.com/go-kit/kit/blob/master/transport/http/jsonrpc/client.go#L143) to this PR or a new PR, as it has the same/similar implementation so I suppose the same "bug" (did not try it).

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -85,7 +85,7 @@ func ClientFinalizer(f ...ClientFinalizerFunc) ClientOption {
 
 // BufferedStream sets whether the Response.Body is left open, allowing it
 // to be read from later. Useful for transporting a file as a buffered stream.
-// That body has to be Closed to propery end the request
+// That body has to be Closed to propery end the request.
 func BufferedStream(buffered bool) ClientOption {
 	return func(c *Client) { c.bufferedStream = buffered }
 }
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -85,6 +85,7 @@ func ClientFinalizer(f ...ClientFinalizerFunc) ClientOption {
 
 // BufferedStream sets whether the Response.Body is left open, allowing it
 // to be read from later. Useful for transporting a file as a buffered stream.
+// That body has to be Closed to propery end the request
 func BufferedStream(buffered bool) ClientOption {
 	return func(c *Client) { c.bufferedStream = buffered }
 }
@@ -132,6 +133,8 @@ func (c Client) Endpoint() endpoint.Endpoint {
 			return nil, err
 		}
 
+		// If we expect a buffered stream, we don't cancel the context when the endpoint returns.
+		// Instead, we should call the cancel func when closing the response body.
 		if c.bufferedStream {
 			resp.Body = bodyWithCancel{ReadCloser: resp.Body, cancel: cancel}
 		} else {
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"encoding/xml"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -92,7 +93,6 @@ func BufferedStream(buffered bool) ClientOption {
 func (c Client) Endpoint() endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		ctx, cancel := context.WithCancel(ctx)
-		defer cancel()
 
 		var (
 			resp *http.Response
@@ -112,10 +112,12 @@ func (c Client) Endpoint() endpoint.Endpoint {
 
 		req, err := http.NewRequest(c.method, c.tgt.String(), nil)
 		if err != nil {
+			cancel()
 			return nil, err
 		}
 
 		if err = c.enc(ctx, req, request); err != nil {
+			cancel()
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


transport/http/client_test.go
```diff
@@ -99,7 +99,7 @@ func TestHTTPClient(t *testing.T) {
 
 func TestHTTPClientBufferedStream(t *testing.T) {
 	var (
-		testbody = "testbody"
+		testbody = string(make([]byte, 6000))
```
peterbourgon: Perhaps a comment explaining the choice of 6000? Or, better yet, factor it out into a const, e.g.

```go
var (
    bodysize = 6000 // trigger the foo in the bar package
    testbody = bytes.Repeat('a', bodysize)
)
```

transport/http/client.go
```diff
@@ -126,11 +128,15 @@ func (c Client) Endpoint() endpoint.Endpoint {
 		resp, err = c.client.Do(req.WithContext(ctx))
 
 		if err != nil {
+			cancel()
 			return nil, err
 		}
 
-		if !c.bufferedStream {
+		if c.bufferedStream {
+			resp.Body = bodyWithCancel{ReadCloser: resp.Body, cancel: cancel}
```
peterbourgon: A comment here to motivate this wrapper would also be welcome. Something like this, feel free to rephrase if I got something wrong:

```go
// If we expect a buffered stream, we don't cancel the context when the endpoint returns.
// Instead, we should call the cancel func when closing the response body.
```

transport/http/client_test.go
xescugc: I has not an specific meaning by itself, it big enough to make the test fail hehe. If it where to be over 9000 then we would have to modify the test as the `_, err = response.Body.Read(b)` would not return `io.EOF` but on the second read would do it (I explained this on the main issue hehe).

I added the documentation too :+1:

xescugc:
I've also added documentation on the `BufferedStream` to clarify that the Body has to be closed to properly clean the Request.

transport/http/client_test.go
```diff
@@ -98,8 +98,12 @@ func TestHTTPClient(t *testing.T) {
 }
 
 func TestHTTPClientBufferedStream(t *testing.T) {
+	// bodysize has a size big enought to make the resopnse.Body not an instant read
+	// so if the response is cancelled it wount be all readed and the test would fail
+	// The 6000 has not a particular meaning, it big enough to fulfill the usecase.
+	const bodysize = 6000
```
peterbourgon: I guess my question was more like: why is 6000 big enough, and 9000 too big? Is there a constant or default buffer size or something defined in the stdlib net/http package? If so, can you refer to that?

transport/http/client.go
```diff
@@ -84,6 +85,7 @@ func ClientFinalizer(f ...ClientFinalizerFunc) ClientOption {
 
 // BufferedStream sets whether the Response.Body is left open, allowing it
 // to be read from later. Useful for transporting a file as a buffered stream.
+// That body has to be Closed to propery end the request
```
peterbourgon: Please add a period :)

transport/http/client_test.go
xescugc: I would also want to know that ... but I've tried to search for it, where the body is setted on a response and how is the `io.ReadCoser` is implemented on it, but It was impossible to find it for me, I would really like to have that answer because then I could provide a more accurate answer on what it's happening.

Also the 9000 it's not "too big" but it simply makes the `io.Read` works differently then. From the `io.Read`

> [...] An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF. [...]

So the next read to the body would return the `io.EOF` instead of the first one (I did try it while testing, but It's not something realted to `go-kit` but to the stdlib so IMO no test needed for this).

I think is on the way the body is read, I think that it's read by "chunks" from the "sender" instead of buffering all the body.

ZedLinden:
@xescugc Nice solution with the `bodyWithCancel` struct.  I ran across this PR while searching for solutions to a similar problem that I was having.  Thanks for the great idea!

PR #782, opened 2018-10-24
Add error return to the Registrar interface functions

### LINKED ISSUE #780: SD - Registrator error handling
Is there any reason for not returning any error on the registration of a service (https://github.com/go-kit/kit/blob/ec0cc13a9b00d1214c9b7ae63d7b750626b54c6d/sd/registrar.go#L11) ?

From my point of view, if a service doesn't manage to correctly register, then it might be nice to retry later. Binding to events to watch if you're correctly registered seems overkilled compared to just checking this error.

What are your thoughts?

jstoja:
The Register and Deregister method are logging errors but not
returning any error, which makes error handling not possible.
The implementation of the error is meant to reflect error from the
client api, not from what we consider an error (re-registering, ...).

* Some test cases haven't been altered since errors are expected.
* Eureka error handling is purely an error return from the client, trying to register several time is not considered as an error.

Closes #780.

### DIFF
--- a/sd/consul/registrar.go
+++ b/sd/consul/registrar.go
@@ -3,9 +3,8 @@ package consul
 import (
 	"fmt"
 
-	stdconsul "github.com/hashicorp/consul/api"
-
 	"github.com/go-kit/kit/log"
+	stdconsul "github.com/hashicorp/consul/api"
 )
 
 // Registrar registers service instance liveness information to Consul.
@@ -26,19 +25,23 @@ func NewRegistrar(client Client, r *stdconsul.AgentServiceRegistration, logger l
 }
 
 // Register implements sd.Registrar interface.
-func (p *Registrar) Register() {
-	if err := p.client.Register(p.registration); err != nil {
+func (p *Registrar) Register() error {
+	err := p.client.Register(p.registration)
+	if err != nil {
 		p.logger.Log("err", err)
-	} else {
-		p.logger.Log("action", "register")
+		return err
 	}
+	p.logger.Log("action", "register")
+	return nil
 }
 
 // Deregister implements sd.Registrar interface.
-func (p *Registrar) Deregister() {
-	if err := p.client.Deregister(p.registration); err != nil {
+func (p *Registrar) Deregister() error {
+	err := p.client.Deregister(p.registration)
+	if err != nil {
 		p.logger.Log("err", err)
-	} else {
-		p.logger.Log("action", "deregister")
+		return err
 	}
+	p.logger.Log("action", "deregister")
+	return nil
 }
--- a/sd/etcd/registrar.go
+++ b/sd/etcd/registrar.go
@@ -4,9 +4,8 @@ import (
 	"sync"
 	"time"
 
-	etcd "go.etcd.io/etcd/client"
-
 	"github.com/go-kit/kit/log"
+	etcd "go.etcd.io/etcd/client"
 )
 
 const minHeartBeatTime = 500 * time.Millisecond
@@ -69,8 +68,9 @@ func NewRegistrar(client Client, service Service, logger log.Logger) *Registrar 
 
 // Register implements the sd.Registrar interface. Call it when you want your
 // service to be registered in etcd, typically at startup.
-func (r *Registrar) Register() {
-	if err := r.client.Register(r.service); err != nil {
+func (r *Registrar) Register() error {
+	err := r.client.Register(r.service)
+	if err != nil {
 		r.logger.Log("err", err)
 	} else {
 		r.logger.Log("action", "register")
@@ -78,6 +78,7 @@ func (r *Registrar) Register() {
 	if r.service.TTL != nil {
 		go r.loop()
 	}
+	return err
 }
 
 func (r *Registrar) loop() {
@@ -104,17 +105,18 @@ func (r *Registrar) loop() {
 
 // Deregister implements the sd.Registrar interface. Call it when you want your
 // service to be deregistered from etcd, typically just prior to shutdown.
-func (r *Registrar) Deregister() {
-	if err := r.client.Deregister(r.service); err != nil {
+func (r *Registrar) Deregister() error {
+	err := r.client.Deregister(r.service)
+	if err != nil {
 		r.logger.Log("err", err)
 	} else {
 		r.logger.Log("action", "deregister")
 	}
-
 	r.quitmtx.Lock()
 	defer r.quitmtx.Unlock()
 	if r.quit != nil {
 		close(r.quit)
 		r.quit = nil
 	}
+	return err
 }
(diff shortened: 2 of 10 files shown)


sd/etcd/registrar.go
```diff
@@ -69,15 +68,17 @@ func NewRegistrar(client Client, service Service, logger log.Logger) *Registrar
...
+	err := r.client.Register(r.service)
+	if err != nil {
 		r.logger.Log("err", err)
 	} else {
 		r.logger.Log("action", "register")
 	}
 	if r.service.TTL != nil {
 		go r.loop()
 	}
+	return err
```
peterbourgon: If Register returns an error, it shouldn't do other things, like `go r.loop`, don't you think?

sd/consul/registrar.go
```diff
@@ -26,19 +25,23 @@ func NewRegistrar(client Client, r *stdconsul.AgentServiceRegistration, logger l
...
 // Register implements sd.Registrar interface.
-func (p *Registrar) Register() {
-	if err := p.client.Register(p.registration); err != nil {
+func (p *Registrar) Register() error {
+	err := p.client.Register(p.registration)
+	if err != nil {
 		p.logger.Log("err", err)
-	} else {
-		p.logger.Log("action", "register")
+		return err
```
peterbourgon: Don't log an error you also return.

sd/consul/registrar.go
```diff
@@ -26,19 +25,23 @@ func NewRegistrar(client Client, r *stdconsul.AgentServiceRegistration, logger l
...
 // Deregister implements sd.Registrar interface.
-func (p *Registrar) Deregister() {
-	if err := p.client.Deregister(p.registration); err != nil {
+func (p *Registrar) Deregister() error {
+	err := p.client.Deregister(p.registration)
+	if err != nil {
 		p.logger.Log("err", err)
-	} else {
-		p.logger.Log("action", "deregister")
+		return err
```
peterbourgon: Don't log an error you also return.

sd/etcd/registrar.go
```diff
@@ -104,17 +105,18 @@ func (r *Registrar) loop() {
...
 		r.logger.Log("action", "deregister")
 	}
-
 	r.quitmtx.Lock()
 	defer r.quitmtx.Unlock()
 	if r.quit != nil {
 		close(r.quit)
 		r.quit = nil
 	}
+	return err
```
peterbourgon: Same thing here, if Deregister returns an error, it shouldn't do other stuff.

I think this reveals a fundamental design tension: what should an application do if its service discovery Register or Deregister action fails? Can it do anything useful with that error besides log it? Arguably, Register failure might want to terminate the service, in some circumstances, but if Deregister fails, can/should we do anything?

sd/etcdv3/registrar.go
```diff
@@ -66,22 +66,24 @@ func NewRegistrar(client Client, service Service, logger log.Logger) *Registrar
 
 // Register implements the sd.Registrar interface. Call it when you want your
 // service to be registered in etcd, typically at startup.
-func (r *Registrar) Register() {
+func (r *Registrar) Register() error {
 	if err := r.client.Register(r.service); err != nil {
 		r.logger.Log("err", err)
-		return
+		return err
```
peterbourgon: Log or return, not both.

sd/etcdv3/registrar.go
```diff
@@ -93,4 +95,5 @@ func (r *Registrar) Deregister() {
 		close(r.quit)
 		r.quit = nil
 	}
+	return err
```
peterbourgon: Similar questions about Deregister as above.

jstoja:
Hey thanks for the review and the very pertinent comments!

The reason for returning an error is exactly the use-case you pointed out.
* When the registration fails, the error might be useful to implement a retry, or simply correctly exit. I agree that this is arguably useful, since we could also query the registry to check if we're correctly there.

* When the de-registration fail, we could also retry it. Some service registry systems like Consul would just mark the service as un-healthy and [only remove it from the pool after a some time](https://www.consul.io/api/agent/check.html#deregistercriticalserviceafter).

I would prefer having an error returned and do something about it (or just not using it), than having nothing and try to find out somehow if my action succeeded.

The log/return thing was let like this because I don't wanted to change the actual logging pattern, but I'll change this as soon as I have time to.

sd/etcd/registrar.go
jstoja: Actually the loop tries to register every TTL, so I feel that I still should let the loop be launched. I don't know how this should be handled.

peterbourgon:
Until now we've taken the approach that an error in Register or Deregister isn't fatal, and that we should log it, but keep doing whatever it is we're doing in the background, in the hopes that it eventually resolves. I understand that this PR supposes it would be nice to let the application know when these things fail, and in the abstract I agree, but we've discovered that changing the semantics of errors in these methods is more subtle than we anticipated. 

By convention, a non-nil error means the function failed, and any other value returned by the it is undefined; I don't think we can break that convention. Should we have another "channel" (not necessarily a Go chan) on which to report errors to the application? Is that added complexity in every constructor, or Register/Deregister call, worth the value it brings? In case of error, should we keep any registration loop goroutines running, or terminate them?

I don't have good answers to these questions. I could be convinced there are good answers to them, but I don't think what we have right now is it.

jstoja:
I think the first issue here is that there are 2 registration workflow present here. The (1) first one is a simple service register where it will hold the service until a potential healthcheck fail for a defined amount of time, or it simply un-register. The (2) second one is a an insert in the k/v store with a TTL, that needs a loop to be kept present.
Even if both have similar semantics from go-kit point of view (the current service being registered somewhere), the expectation of how it's running is different.
In the case of a TTL with a loop (2), if the service is not registered on the first call, it might be in a later one. In the case of the service register (1), if the first call fails, then we won't ever be registered if we're not retrying in the application (where we currently have no feedback if it did succeed or not).
Maybe we could try to:
1. add a loop on the (1) (even if that feels a bit weird) to be sure that we're registered.
2. add a new method to check the registration object and see if we're correctly registered (in the case of the (1) we can have the actual service object and in (2) the actual key with the current tick ?). I'm not very fond of calling a method that doesn't return any error and then calling another one to get the error.
3. improve the documentation with some boilerplate code checking directly in the service discovery system if we're correctly registered. This would mean no change in the Registrar API but more application code.

yramanovich:
I agree that the Register and Deregister methods don't return any errors it's kind strange and @jstoja proposal makes sense. The problem is that different implementations follow different ways. In case of consul it's easy - we can just return error and that's it. If to speak about eureka it's more complicated: uses loops, heartbeats and so on. 
So i see 2 ways: 
1) we can just try to register and return error if fails - developer should handle error and write retry logic. 
2) implement loop style and return <-chan error, so it'll be possible to see errors.

PR #785, opened 2018-10-31
Fix logging

michurin:
I suggest such pull request can be opened without issue and honors :-)


basvanbeek:
thanks @michurin. even the smallest of fixes are highly appreciated!

PR #790, opened 2018-11-07
Add FinalizerFunc to NATS transport

kirooha:
I want NATS transport to have FinalizerFunc, as well as HTTP transport has. For example, to count working goroutines which handles request from publishers.


peterbourgon:
Sure! Thanks!

PR #793, opened 2018-11-12
Support logrus.Field logger in the logrus adapter

sagikazarmark:
Logrus exposes an interface `FieldLogger`. It is usually better to hint against this interface, because the main logger instance returns a different type for loggers preannotated with structured context (namely `*Entry`). This way one can do this:

```go
logger := kitlogrus.NewLogrusLogger(logrusLogger.With("component", "my_component"))
```

### DIFF
--- a/log/logrus/logrus_logger.go
+++ b/log/logrus/logrus_logger.go
@@ -11,13 +11,13 @@ import (
 )
 
 type logrusLogger struct {
-	*logrus.Logger
+	logrus.FieldLogger
 }
 
 var errMissingValue = errors.New("(MISSING)")
 
 // NewLogrusLogger returns a go-kit log.Logger that sends log events to a Logrus logger.
-func NewLogrusLogger(logger *logrus.Logger) log.Logger {
+func NewLogrusLogger(logger logrus.FieldLogger) log.Logger {
 	return &logrusLogger{logger}
 }
 


log/logrus/logrus_logger.go
```diff
@@ -11,13 +11,13 @@ import (
...
 type logrusLogger struct {
-	*logrus.Logger
+	logrus.FieldLogger
 }
 
 var errMissingValue = errors.New("(MISSING)")
 
-// NewLogrusLogger returns a go-kit log.Logger that sends log events to a Logrus logger.
-func NewLogrusLogger(logger *logrus.Logger) log.Logger {
+// NewLogrusLogger returns a go-kit log.FieldLogger that sends log events to a Logrus logger.
```
ChrisHines: The function still returns a go-kit log.Logger. It's the input type that has changed to accept the logrus.FieldLogger type.

The phrase "... to a Logrus logger" still seems accurate and perhaps more to the point. Ending with "... to a logrus.FieldLogger" would be more precise, but just repeating what the code says. I'm not a Logrus user, so I'm not sure which phrasing is more intuitive for that user base. Thoughts?

```suggestion
// NewLogrusLogger returns a go-kit log.Logger that sends log events to a Logrus logger.
```

sagikazarmark: Ah, good catch, it was kinda automatic change. 🙂

sagikazarmark:
Thanks!

log/logrus/logrus_logger.go
giautm: image
From Github suggest great platform. 👍

PR #798, opened 2018-11-20
Fix wrong sprintf verbs

dvrkps:


log/log_test.go
```diff
@@ -60,7 +60,7 @@ func TestContextMissingValue(t *testing.T) {
 // Valuers, regardless of how many times With has been called.
 func TestContextStackDepth(t *testing.T) {
 	t.Parallel()
-	fn := fmt.Sprintf("%n", stack.Caller(0))
+	fn := fmt.Sprintf("%v", stack.Caller(0))
```
ChrisHines: `%n` is a valid format verb for the return from `stack.Caller(0)`. See: https://godoc.org/github.com/go-stack/stack#Call.Format.

That said, a while ago I removed the dependency on that package from the non-test code in this package. We could also do that for the test code but we would have to either reimplement parts of it locally or take a different approach in this test.

PR #801, opened 2018-11-21
Add support for gomodules with minimum compatible versions

ValeryPiashchynski:


ChrisHines:
I suggest we open an issue to discuss what Go kit's strategy for supporting modules before we merge something like this. We may want to go with separate modules for things like `log`, `metrics`, etc.

We need to be careful and think through our plan first.

ValeryPiashchynski:
Yep, agree with you

peterbourgon:
In the medium term, I think I have convinced myself that I'd like to break out each of the Go kit packages that have successfully weathered the test of time to their own repositories, and version them independently. This would include 

- transport/* — whether as individual kithttp, kitgrpc, etc. repos, or in a single transport repo, TBD 
- endpoint — at least until generics lands, then this will likely go away in a refactor
- log
- tracing
- kitgen
- _probably_ metrics
- _maybe_ auth

In the short term, I have no strong feelings, but would appreciate it if someone could lay out the consequences of (for example) merging a PR like this one.

PR #802, opened 2018-11-21
Add amqp.Delivery to the request func

ValeryPiashchynski:
Add amqp.Delivery to the request func in order to have the ability to change incoming message before decode function in case of zipkin context. To add zipkin context to delivery message before it goes to the decode function


peterbourgon:
Any objections from anyone watching?

krzysztof-gzocha:
This would be pretty useful. Any idea when it could be merged?

Issue #804, opened 2018-11-23
Add error tag to spans on OT endpoint middleware


jlordiales:
Hi!
I think it would be useful for the OpenTracing endpoint middleware (https://github.com/go-kit/kit/blob/master/tracing/opentracing/endpoint.go) to inspect the `error` returned by the next endpoint and set the `error` tag to true/false accordingly (it could also optionally set the `error.object` log field).
In addition to just inspecting the returned `error` it could also check if the `response` struct implements `endpoint.Failer` and apply the same logic.

If you think this would be useful I'd be happy to create a PR, just wanted to get your thoughts first.

peterbourgon:
@basvanbeek?

basvanbeek:
Makes sense to add. The OpenCensus middleware already does somethjng similar with the status object and also inspects the endpoint.Failer checks.

jlordiales:
Great, I'll create a PR this week.

PR #805, opened 2018-11-26
Tag spans on error (from endpoint or service) in OpenTracing middleware

jlordiales:
Adds error handling for the Open Tracing endpoint middleware, following the recommended conventions: https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table

Closes https://github.com/go-kit/kit/issues/804

### DIFF
--- a/tracing/opentracing/endpoint.go
+++ b/tracing/opentracing/endpoint.go
@@ -27,7 +27,15 @@ func TraceServer(tracer opentracing.Tracer, operationName string) endpoint.Middl
 			defer serverSpan.Finish()
 			otext.SpanKindRPCServer.Set(serverSpan)
 			ctx = opentracing.ContextWithSpan(ctx, serverSpan)
-			return next(ctx, request)
+
+			response, err := next(ctx, request)
+
+			spanErr := effectiveSpanError(response, err)
+			if spanErr != nil {
+				markSpanWithError(serverSpan, spanErr)
+			}
+
+			return response, err
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


tracing/opentracing/endpoint.go
```diff
@@ -27,7 +26,16 @@ func TraceServer(tracer opentracing.Tracer, operationName string) endpoint.Middl
 			defer serverSpan.Finish()
 			otext.SpanKindRPCServer.Set(serverSpan)
 			ctx = opentracing.ContextWithSpan(ctx, serverSpan)
-			return next(ctx, request)
+
+			response, err := next(ctx, request)
+			if err != nil {
+				markSpanWithError(serverSpan, err)
+			}
+			if res, ok := response.(endpoint.Failer); ok && res.Failed() != nil {
```
peterbourgon: Does `else if` better capture the semantics here?

tracing/opentracing/endpoint.go
```diff
@@ -3,10 +3,9 @@ package opentracing
 import (
 	"context"
 
+	"github.com/go-kit/kit/endpoint"
```
basvanbeek: please revert change in import grouping, we typically do:

1. standard library
2. external dependencies
3. go-kit internal dependencies

tracing/opentracing/endpoint_test.go
```diff
@@ -4,13 +4,31 @@ import (
 	"context"
 	"testing"
 
+	"github.com/go-kit/kit/endpoint"
```
basvanbeek: See previous comment on import grouping

jlordiales:
@peterbourgon @basvanbeek fixed the imports and squashed. Thanks for the feedback!

peterbourgon:
For the record, goimports changed recently and automatically applies 2-level import grouping, stdlib and non-stdlib. I’m not happy about it but that’s how most code is gonna end up formatted from here forward.

jlordiales:
Yeah that's how I ended up with those changes here, running `goimports` automatically on save. It does seem to keep the separation if you explicitly add a new line between groups (or maybe that's my IDE, not sure)

jlordiales:
@basvanbeek I think this is ready for a second quick review after the import changes were reverted.

basvanbeek:
I think I'd like it to be more inline with what has been done in the OpenCensus middleware. Distinction is made there between business errors (Failer), load balancer errors (lb.RetryError) and other (non go kit) errors. 
If you look at that implementation you also see that consumers can chose to ignore business errors (in OT this would mean not setting error = true) as they view business errors as a successful response / behavior of the system.

peterbourgon:
@jlordiales Friendly ping? :)

jlordiales:
@peterbourgon sorry, completely dropped the ball on this one. I'll look into it this week.

PR #809, opened 2018-12-04
[transport/http] adding request and response encoders for Protobuf

### LINKED ISSUE #808: Add support for encoding Protobuf requests and responses to transport/http
I often use Protobuf-over-HTTP and I feel this repo will be a better home for the functions than Gizmo.

Related PR incoming, feel free to reject!

jprobinson:
Bringing these over from our `marvin` framework now that we can use Gizmo + go-kit in Google App Engine (hooray!)

I'm not 100% sure what we should do in the case of error while marshalling the protobuf so just commenting with a frowny face for now. Any thoughts for https://github.com/go-kit/kit/compare/master...jprobinson:encode-proto#diff-05760750cf2e3aac59bd773896ee182fR202?

This resolves #808.


transport/http/server.go
```diff
@@ -169,6 +171,45 @@ func EncodeJSONResponse(_ context.Context, w http.ResponseWriter, response inter
...
+	w.WriteHeader(code)
+	if code == http.StatusNoContent {
+		return nil
+	}
+	if res == nil {
+		return nil
+	}
+	b, err := proto.Marshal(res)
+	if err != nil {
+		// returning an error here will cause a second header write, swallowing :(
```
peterbourgon: I think you have to return the error and trigger the second header write, otherwise there's no way clients are going to know this didn't work. Am I being naïve?

transport/http/client.go
```diff
@@ -5,12 +5,14 @@ import (
...
 	"encoding/json"
 	"encoding/xml"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 
 	"github.com/go-kit/kit/endpoint"
+	"github.com/golang/protobuf/proto"
```
ChrisHines: Do we want to add this dependency here? I would venture a guess that `github.com/go-kit/kit/transport/http` is one of the most commonly used packages within Go kit and I would also venture a guess that pairing it with protobuf does not account for the majority of use cases.

Is it possible to offer the protobuf functionality as a separate package (I don't know as I don't have time to review the rest of this PR thoroughly at the moment)?

To be clear, my objection is not strenuous. Although `github.com/golang/protobuf/proto` is a large package, it doesn't bring with it any transitive dependencies outside the standard library. But this would be the first non-Go kit and non-standard library dependency here, so I think it's worth being conservative if possible.

peterbourgon: I had the same hesitation, for the record.

basvanbeek: Agreed... I'd love to have this contribution, but I see this more in a subpackage of `transport/http`.

transport/http/server.go
jprobinson: Nah, I've been on the fence but leaning this way. After this nudge, we're returning the error.

transport/http/client.go
jprobinson: I'm down to push these to a subpackage. Any recommendations on package name? `proto`? `httpproto`?

jprobinson:
I've pushed updates with the new functions in a new subpackage: `transport/http/proto`

jprobinson:
I've been thinking about the package name a big and `proto` may be a bit annoying as it matches the match proto package. Any thoughts on naming it `pb` or `httppb`?

peterbourgon:
Go kit consumers are already essentially required to make use of import aliases to get `transport/http` into their lexical scope. I think for now it's best to leverage that assumption, keep with the naming conventions, and stick with `proto`. (When we get around to a repo and package restructuring, I'll probably stick `kit` as a prefix to most package names.)

peterbourgon:
Cool, I like this well enough. Any objections? I'll merge in a couple days otherwise.

jprobinson:
I don't mean to be a bugger, but any chance we can merge this one soon-ish? I'm working on some internal docs/boilerplates and would love to include this lil tidbit.

peterbourgon:
Oh! I thought I'd done it. Mea culpa.

jprobinson:
Thanks, Peter!

PR #818, opened 2019-01-05
Fix Consul SD deregister service if id empty

vlastv:
Consul allows [register service without `ID`](https://www.consul.io/api/agent/service.html).

> `ID (string: "")` - Specifies a unique ID for this service. This must be unique per agent. This defaults to the `Name` parameter if not provided.

Then, deregister must be by `Name`.


peterbourgon:
Sure. Can you add a test?

peterbourgon:
Still need that test.

MaruHyl:
@vlastv hello, are you still work on this pr? If not, can I continue your work?

peterbourgon:
@MaruHyl Go for it.

PR #821, opened 2019-01-09
Change span injection/extraction from HTTP request

nboukeffa:
Today, `TextMap` format is used when injecting or extracing *span* from HTTP requests.

This format will treat the HTTP header value (that contains the span) as is (without decoding). Conversely the `HTTPHeaders` format will try to URL decode the value first.

The issue is that some opentracing clients like [java jaeger client](https://github.com/jaegertracing/jaeger-client-java/blob/f5f6ad1f4eaf606aba07562b7543958ec4658ebe/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java#L206) is URL encoding the HTTP header value before sending it over the wire. Then when the server receive the request and try to extract the span from it it will fail if `TextMap` format is used istead `HTTPHeaders` format.

For example the java jaeger client will send something like this: `Uber-Trace-Id: 9ec076efd89825a%3A9c876c01059545b2%3Ab9ec076efd89825a%3A1` but the golang jaeger client is expecting this `Uber-Trace-Id: 9ec076efd89825a:9c876c01059545b2:b9ec076efd89825a:1` when the `TextMap` format is used.

In this case the call results to the following [error](https://github.com/jaegertracing/jaeger-client-go/blob/master/context.go#L111): `String does not match tracer state format`

Because the `http.go` file is used precisely for dealing with span in a HTTP context, I propose to default the span format to `HTTPHeaders`

### DIFF
--- a/tracing/opentracing/http.go
+++ b/tracing/opentracing/http.go
@@ -6,7 +6,7 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/opentracing/opentracing-go"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


basvanbeek:
Thanks for the PR. Looks good to me so merging.

Issue #841, opened 2019-02-27
Async (non-RPC) AMQP transport


madshov:
Hi All,

Any documentation/examples that exist around AMQP transport, subscriber/publisher?

Thanks

madshov:
More specifically, I'm just interested in using RabbitMQ as a simple message broker. Receive a message and ACK it, if OK. I can see that the ServeDelivery() on the subscriber will respond to incoming messages, like RPC fashion. I'm not interested in the sending the response. Can I ignore it or could that cause any problems?

peterbourgon:
[See previous discussion here](https://github.com/go-kit/kit/pull/746#issuecomment-454909761). As I understand it, it would be possible to add an option to transport/amqp to not require a response, but it's not currently implemented.

madshov:
Thanks Peter.
So if I understand you correctly, within go-kit, you are only considering AMQP transport if it does RPC?
If I want true one-way communication, it would be something I have to build on the side. That is, if I don't choose to send back an empty response with the `EncodeNopResponse()`.

/mads

peterbourgon:
At the moment, yes. But it sounds like it would be relatively easy to enhance the AMQP package to not require a response.

peterbourgon:
I think #850 makes this fixed. If that's not the case, please re-open with details.

FabianTe:
@madshov I am interested in how you handled one-way communication. I'd like to implement a Go-based worker for a legacy system that uses RabbitMQ as a worker queue. In this scenario I also don't want to send a response.

Did you sucessfully implement this?

madshov:
Hi @FabianTe,
Yes I have successfully implemented @mattfung additions in go-kit v0.9.0 (i believe)

PR #850, opened 2019-03-19
adds NopResponse to AMQP transport

mattfung:
In response to [previous discussion](https://github.com/go-kit/kit/pull/746#issuecomment-455607315) and #841, here is a commit that allows users to configure AMQP Subscriber to not issue a response to the sender. Similiarly, one is also able to configure AMQP Consumer to not wait for a response.

Comments/ Suggestions are more than welcome.


peterbourgon:
This leaves a bad taste in my mouth, because we're not doing RPC anymore. But I guess it fits the model.

PR #851, opened 2019-03-23
wide log tooling

nelz9999:
Was just trying to think how I'd do the one-wide-log-per-request thing, and thought others might benefit from it as well.

Also considered doing it as a `NewJSONArrayLogger`, but figured doing it at the writer level would reduce the amount of time that would need to be locked in mutex.

### DIFF
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -87,3 +87,28 @@ func safeError(err error) (s interface{}) {
 	s = err.Error()
 	return
 }
+
+type jsonArrayWriter struct {
+	events []json.RawMessage
+}
+
+// NewJSONArrayWriter returns an io.Writer that accumulates log events
+// in memory, so that you can emit them all as sub-events within a single
+// outer event, by passing the writer itself as a log value. This can
+// help to achieve the modern observability
+// recommendation to emit one very wide event per request.
+//
+// The accumulation happens by appending to a slice, so if you
+// need to guarantee goroutine-safe logging, wrap it with NewSyncWriter.
+func NewJSONArrayWriter() io.Writer {
+	return &jsonArrayWriter{[]json.RawMessage{}}
+}
+
+func (w *jsonArrayWriter) Write(p []byte) (int, error) {
+	w.events = append(w.events, json.RawMessage(append([]byte{}, p...)))
+	return len(p), nil
+}
+
+func (w *jsonArrayWriter) MarshalJSON() ([]byte, error) {
+	return json.Marshal(w.events)
+}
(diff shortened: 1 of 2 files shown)


ChrisHines:
I am not familiar with this idea. Do you have any references that I can read about the "one-wide-log-per-request thing"?

nelz9999:
Sure. The most I'm hearing about this is from the Honeycomb folks. (Disclaimer: I'm a friend of several of them.)

Charity goes into it in this thread: https://twitter.com/mipsytipsy/status/1042817542648082432
And they have this up too: https://www.honeycomb.io/blog/how-are-structured-logs-different-from-events/

But I got to thinkin', might this have been a useful pattern in a different context with different tools? (Like, e.g. at a previous job where we used Splunk.) I think it might've been useful. At least enough to try to think how I'd do it with the go-kit logger. Which is how we got to this PR. :grin:

peterbourgon:
Related work: https://github.com/peterbourgon/ctxlog

peterbourgon:
A more-general version of that package, in one code block:

```go
type ctxdata struct {
	data map[string]string
}

func (d *ctxdata) set(key, val string) {
	d.data[key] = val
}

func (d *ctxdata) get(key, def string) string {
	if val, ok := d.data[key]; ok {
		return val
	}
	return def
}

func newContextData(ctx context.Context) (context.Context, *ctxdata) {
	d := &ctxdata{data: map[string]string{}}
	return context.WithValue(ctx, ctxdatakey, d), d
}

func contextDataFrom(ctx context.Context) *ctxdata {
	v := ctx.Value(ctxdatakey)
	if v == nil {
		return nil
	}
	d, ok := v.(*ctxdata)
	if !ok {
		return nil
	}
	return d
}

type ctxdatakeytype struct{}

var ctxdatakey ctxdatakeytype
```

And usage:

```go
func middleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		ctx, d := newContextData(r.Context())
		defer func() {
			// e.g. log.Print(d.get("key"))
		}()
		next.ServeHTTP(w, r.WithContext(ctx))
	})
}

func myHandler(w http.ResponseWriter, r *http.Request) {
	contextDataFrom(r.Context()).set("key", "value")
	// ...
}
```

peterbourgon:
Oh, I just saw this was a PR, not an issue. Something _like_ this may make sense to be part of package log—or, probably, a sub-package—but I think it's not quite this. I don't like the coupling to JSON, and it feels too low-level, somehow.

nelz9999:
Hrm. Based on Peter's feedback, I just went looking to see if I could make this more widely usable, outside of JSON. 

I considered converting `jsonArrayWriter` to something like `accumulatingWriter`, and then have the `MarshalJSON` func JIT convert the lines to `RawJSON`.

But then I started looking into the `logfmt` library. It seems that library explicitly eschews any collections as values: https://github.com/go-logfmt/logfmt/blob/master/encode.go#L223 Which I take to mean that this approach is not useful in a `logfmt` scenario.

So I can close out this PR, since there doesn't seem to be an appetite for it.

Issue #725, opened 2018-07-04
How about web socket tranports ?


Astone-Chou:
https://github.com/gorilla/websocket

peterbourgon:
I'll happily review a PR.

PR #859, opened 2019-04-10
WebSocket Support

blainsmith:
I saw a lot of requests kicking around about WebSocket support so I thought I would take an initial pass at what something like that would look like. A lot of this code is a WIP, but the happy path in the addsvc example does work.

The README has a lot of the information to cover the concepts about handling the initial WebSocket connection and defining the Subprotocol to route incoming messages to the proper Endpoints. Because WebSockets are agnostic about what messages look like it was up to leveraging the Subprotocol to define what the server will accept and know how to decode and encode. I left the implementation of Subprotocol decoding/encoding up to the user of the package, but the addsvc does make use of a simple STOMP protocol to illustrate how to wire up a sever.

So now I am just gauging interest if this is something people would use and/or this should belong in the core kit. Questions/comments/concerns welcome!

### DIFF
--- a/examples/addsvc/cmd/addsvc/addsvc.go
+++ b/examples/addsvc/cmd/addsvc/addsvc.go
@@ -46,6 +46,7 @@ func main() {
 		grpcAddr       = fs.String("grpc-addr", ":8082", "gRPC listen address")
 		thriftAddr     = fs.String("thrift-addr", ":8083", "Thrift listen address")
 		jsonRPCAddr    = fs.String("jsonrpc-addr", ":8084", "JSON RPC listen address")
+		wsAddr         = fs.String("ws-addr", ":8085", "Web Socket listen address")
 		thriftProtocol = fs.String("thrift-protocol", "binary", "binary, compact, json, simplejson")
 		thriftBuffer   = fs.Int("thrift-buffer", 0, "0 for unbuffered")
 		thriftFramed   = fs.Bool("thrift-framed", false, "true to enable framing")
@@ -168,6 +169,7 @@ func main() {
 		grpcServer     = addtransport.NewGRPCServer(endpoints, tracer, zipkinTracer, logger)
 		thriftServer   = addtransport.NewThriftServer(endpoints)
 		jsonrpcHandler = addtransport.NewJSONRPCHandler(endpoints, logger)
+		wsHandler      = addtransport.NewWSHandler(endpoints, logger)
 	)
 
 	// Now we're to the part of the func main where we want to start actually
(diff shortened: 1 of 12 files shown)


peterbourgon:
Whoa this looks cool. Can you say a few more words about the motivations or use cases? Or maybe just link to some of the tickets/comments you're talking about?

blainsmith:
Sure thing. Let me collect some more thoughts when I have more spare thinking time tonight and will post it here. Just wanted to let you know I got this.

peterbourgon:
Friendly ping :)

blainsmith:
:derp:

There was a request for socket support, but was closed due to inactivity or no PR. https://github.com/go-kit/kit/issues/725

## Use Case
My personal interest in it was I was recently working on an HTTP API for a multiplayer online game and it is more common to use WebSockets in that industry. While HTTP still has its place for certain scenarios game developers use WebSockets more for its real-time and bi-directional capabilities. Some common use cases are:

- Sending game play data for kills, deaths, item pick ups, points, etc.
- Initiating a matchmaking session
- Real-time leaderboards
- Achievement alerts

Some of these cases are no-op requests, but others are certainly request/response. The request/response cases might take seconds to complete depending on the case. For instance, a game client might send a `join-match` request and receive no immediate response, but some time in the future the game client might receive a `found-match` response so the client can take action.

## Implementation Choices
I chose to implement this using the WebSocket protocol instead of raw sockets because it would make it fit within the `Endpoint` pattern of HTTP and it also has [well-defined specification](https://tools.ietf.org/html/rfc6455) to follow, specifically around supplying message types and encoding via the [Subprotocol header](https://tools.ietf.org/html/rfc6455#section-11.3.4). The handler can choose to support multiple subprotocols and they are validated at the time of client connection. If the subprotocol is not support when the client requests to connect, it will receive an error and the connection will be denied. If successful, every messages must adhere to the subprotocol it established for the life of the connection. Subprotocol is similar to the `Accept` and `Content-Type` headers, but not only does it define the type of content, but also the structure.

I hope this covers more what I was thinking in supporting WebSockets and I am open to discuss further if this is something it should be support or if the implementation needs to go in a different direction. One thing I would like to investigate is not using a concrete WebSocket library and instead allow the implementer to chose which one is the best choice.

peterbourgon:
Thanks, I'll dig in and give some first thoughts in the next few days.

wardour:
These changes also open the door to gRPC streams (2-way)

I might have some bandwidth to help on this, nice work.

noprom:
Hi, @blainsmith , do you have any ideas about how to integrate [gobwas/ws](https://github.com/gobwas/ws) with `go-kit`?
Because according to this [benchmark](https://github.com/uNetworking/uWebSockets/tree/master/benchmarks), `gobwas/ws` is much faster than `gorilla/websocket`.

abeMedia:
Any update on this? Would love to see it merged!

crjacinro:
Can we integrate the gorilla/websockets in the meantime with gokit?

kallydev:
Any update on this? This PR is useful and I hope to merge as soon as possible. Also, go-kit hasn't had any commit for two months, what's happened recently?

@peterbourgon

peterbourgon:
It is unlikely that I'm going to merge this PR.

Issue #861, opened 2019-04-13
Transport level Error handlers


sagikazarmark:
#### Proposal

Currently most transport implementations accept a logger instance. When a transport level error occurs, they are usually handled by a code like this:

```
	if err != nil {
		s.logger.Log("err", err)
		s.errorEncoder(ctx, err, w)
		return
	}
```

I'm proposing to add a new concept in addition to/replacing server loggers: error handlers.

With an error handler in place, the above code would look like this:

```
	if err != nil {
		s.errorHandler.Handle(err)
		s.errorEncoder(ctx, err, w)
		return
	}
```

The interface for such handler:

```
type ErrorHandler interface {
    Handle(err error)
}
```

#### Reasoning

While logs might be the most common target of error events, error can also be considered a distinct domain in which developers might want to take different actions than logging, for example sending the error (with context, like stack trace) to an error tracking service or triggering some sort of alert.

Of course, logs remain a natural target (and an error handler which eventually logs the error would make perfect sense), but it takes extra effort to "reverse engineer" *every* log to see if it contains an error to handle it differently.

peterbourgon:
In general I support this idea. It would be important to not break the current API, probably by having a default error handler log to the transport's logger.

sagikazarmark:
> It would be important to not break the current API, probably by having a default error handler log to the transport's logger.

Exactly what I had in mind.

I will submit a PR shortly.

sagikazarmark:
I think this can be closed as the provided PR has been merged.

peterbourgon:
Yep!

Issue #792, opened 2018-11-12
Add correlation to logs


madshov:
Hi!
I would be nice to be able to add a correlation-id per request to all logs including the ones created in `ServeHTTP` method. Currently errors returned from an endpoint will not have any correlation-id:
```
response, err := s.e(ctx, request)
if err != nil {
	s.logger.Log("err", err)
	s.errorEncoder(ctx, err, w)
	return
}
```
It could just be the request-id from the context. I don't see how it's possible to add it in these cases. If it can be done, I would appreciate, if you would let me know.

sagikazarmark:
After #863 is merged, you will be able to put correlation information into the context and use it in the error handlers.

madshov:
This is now possible in v. 0.9.0 using a custom error log handler.

PR #863, opened 2019-04-16
Implement transport level error handlers

sagikazarmark:
This is a first implementation for #861 

I haven't provided any tests for now as the error loggers were not tested either, but I can write tests if necessary.

### ToDo

- [x] Update examples once the implementation is settled


log/error_handler.go
```diff
@@ -0,0 +1,16 @@
+package log
+
+// ErrorHandler is a transport error handler implementation which logs an error.
+type ErrorHandler struct {
+	logger Logger
+}
```
peterbourgon: This belongs in package transport, not package log.

log/error_handler_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"testing"
+
+	"github.com/go-kit/kit/log"
+)
+
+func TestErrorHandler(t *testing.T) {
+	var output []interface{}
+
+	logger := log.Logger(log.LoggerFunc(func(keyvals ...interface{}) error {
+		output = keyvals
```
peterbourgon: Probably

```go
output = append(output, keyvals...)
```

transport/grpc/server.go
```diff
@@ -68,8 +68,19 @@ func ServerAfter(after ...ServerResponseFunc) ServerOption {
...
-	return func(s *Server) { s.logger = logger }
+	return func(s *Server) {
+		if s.errorHandler == nil {
+			s.errorHandler = log.NewErrorHandler(logger)
+		}
+	}
+}
+
+// ServerErrorHandler is used to handle non-terminal errors. By default, no errors
+// are handled.
```
peterbourgon: I think it would be a little bit less ambiguous to say e.g. "By default, non-terminal errors are ignored."

transport/http/server.go
```diff
@@ -70,8 +70,22 @@ func ServerErrorEncoder(ee ErrorEncoder) ServerOption {
 // of error handling, including logging in more detail, should be performed in a
 // custom ServerErrorEncoder or ServerFinalizer, both of which have access to
 // the context.
+// Deprecated: Use ServerErrorHandler instead.
 func ServerErrorLogger(logger log.Logger) ServerOption {
-	return func(s *Server) { s.logger = logger }
+	return func(s *Server) {
+		if s.errorHandler == nil {
```
peterbourgon: Why the explicit check?

sagikazarmark: Because the ErrorLogger should be considered deprecated and if someone uses an ErrorHandler, it should take precedence.

peterbourgon: IMO ServerErrorLogger should now just be a helper function that calls e.g.

```go
func ServerErrorLogger(logger log.Logger) ServerOption {
    return func(s *Server) { s.errorHandler = LoggingErrorHandler(logger) }
}
```

No need for precedence. You can keep the deprecation warning.

sagikazarmark:
Addressed all comments

transport/error_handler.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func NewLogErrorHandler(logger log.Logger) *LogErrorHandler {
+	return &LogErrorHandler{
+		logger: logger,
+	}
+}
+
+func (h *LogErrorHandler) Handle(err error) {
+	_ = h.logger.Log("err", err)
```
peterbourgon: IMO no need for `_ =`

transport/error_handler.go
```diff
@@ -0,0 +1,26 @@
+package transport
```
peterbourgon: Now that `package transport` actually exists, can we get a `doc.go` with e.g.

```go
// Package transport contains helpers applicable to all supported transports.
package transport
```

peterbourgon:
If you rebase on master, the builds should be fixed.

sagikazarmark:
Thanks!

Rebased and applied the requested changes.

sagikazarmark:
Should I update the examples in this PR or in a separate one?

peterbourgon:
This one, if you don't mind.

sagikazarmark:
Not at all.

I'm wondering if we should somehow consider adding support for something like requested in #792 
Maybe the error handler should accept a context as well? Or we should package the context into the error somehow and make it available through an interface?

peterbourgon:
Seems like a good idea. At first glance, if the errorHandler had a signature like `func(context.Context, error)`, each transport could pass the context it already had, which would work, I think.

sagikazarmark:
I will add that as well then.

sagikazarmark:
Done and done.

sagikazarmark:
Anything else I can do here @peterbourgon ?

peterbourgon:
Thanks, this is a great addition!

sagikazarmark:
Thanks!

PR #865, opened 2019-04-21
Fix data race caused by sharing sd.Event between observers

toffaletti:
`sd/internal/instance/cache.go` causes sd.Event.Instances to be shared between multiple observers:

https://github.com/go-kit/kit/blob/master/sd/internal/instance/cache.go#L71

```
func (r registry) broadcast(event sd.Event) {
	for c := range r {
		c <- event
	}
}
```

but observers frequently directly modify the `sd.Event.Instances` they receive, for example `sd/endpoint_cache.go` contains this:

https://github.com/go-kit/kit/blob/master/sd/endpoint_cache.go#L54

```
...
	// Happy path.
	if event.Err == nil {
		c.updateCache(event.Instances)
...
func (c *endpointCache) updateCache(instances []string) {
	// Deterministic order (for later).
	sort.Strings(instances)
```

The fix is to deep copy `sd.Event` every time it is shared. This requirement should probably be called out in the documentation as well if this is the expected behavior.

Test output before the fix:
```
WARNING: DATA RACE
Read at 0x00c000126350 by goroutine 13:
  sort.(*StringSlice).Less()
      /usr/local/go/src/sort/sort.go:298 +0x92
  sort.doPivot()
      /usr/local/go/src/sort/sort.go:117 +0x564
  sort.quickSort()
      /usr/local/go/src/sort/sort.go:190 +0xa7
  sort.Sort()
      /usr/local/go/src/sort/sort.go:218 +0x8a
  github.com/go-kit/kit/sd.(*endpointCache).updateCache()
      /usr/local/go/src/sort/sort.go:314 +0x88
  github.com/go-kit/kit/sd.(*endpointCache).Update()
      /Users/jason/go/src/github.com/go-kit/kit/sd/endpoint_cache.go:54 +0x3f5
  github.com/go-kit/kit/sd.(*DefaultEndpointer).receive()
      /Users/jason/go/src/github.com/go-kit/kit/sd/endpointer.go:77 +0x132

Previous write at 0x00c000126350 by goroutine 12:
  sort.(*StringSlice).Swap()
      /usr/local/go/src/sort/sort.go:299 +0x144
  sort.medianOfThree()
      /usr/local/go/src/sort/sort.go:77 +0x14b
  sort.doPivot()
      /usr/local/go/src/sort/sort.go:105 +0x9b
  sort.quickSort()
      /usr/local/go/src/sort/sort.go:190 +0xa7
  sort.quickSort()
      /usr/local/go/src/sort/sort.go:197 +0x140
  sort.Sort()
      /usr/local/go/src/sort/sort.go:218 +0x8a
  github.com/go-kit/kit/sd.(*endpointCache).updateCache()
      /usr/local/go/src/sort/sort.go:314 +0x88
  github.com/go-kit/kit/sd.(*endpointCache).Update()
      /Users/jason/go/src/github.com/go-kit/kit/sd/endpoint_cache.go:54 +0x3f5
  github.com/go-kit/kit/sd.(*DefaultEndpointer).receive()
      /Users/jason/go/src/github.com/go-kit/kit/sd/endpointer.go:77 +0x132

Goroutine 13 (running) created at:
  github.com/go-kit/kit/sd.NewEndpointer()
      /Users/jason/go/src/github.com/go-kit/kit/sd/endpointer.go:38 +0x33c
  github.com/go-kit/kit/sd/internal/instance.TestDataRace()
      /Users/jason/go/src/github.com/go-kit/kit/sd/internal/instance/cache_test.go:109 +0x44c
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163

Goroutine 12 (running) created at:
  github.com/go-kit/kit/sd.NewEndpointer()
      /Users/jason/go/src/github.com/go-kit/kit/sd/endpointer.go:38 +0x33c
  github.com/go-kit/kit/sd/internal/instance.TestDataRace()
      /Users/jason/go/src/github.com/go-kit/kit/sd/internal/instance/cache_test.go:108 +0x3fc
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163
```

### DIFF
--- a/sd/internal/instance/cache.go
+++ b/sd/internal/instance/cache.go
@@ -41,8 +41,10 @@ func (c *Cache) Update(event sd.Event) {
 // State returns the current state of discovery (instances or error) as sd.Event
 func (c *Cache) State() sd.Event {
 	c.mtx.RLock()
-	defer c.mtx.RUnlock()
-	return c.state
+	event := c.state
+	c.mtx.RUnlock()
+	eventCopy := copyEvent(event)
+	return eventCopy
 }
 
 // Stop implements Instancer. Since the cache is just a plain-old store of data,
@@ -54,8 +56,10 @@ func (c *Cache) Register(ch chan<- sd.Event) {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 	c.reg.register(ch)
+	event := c.state
+	eventCopy := copyEvent(event)
 	// always push the current state to new channels
-	ch <- c.state
+	ch <- eventCopy
 }
 
 // Deregister implements Instancer.
@@ -70,7 +74,8 @@ type registry map[chan<- sd.Event]struct{}
 
 func (r registry) broadcast(event sd.Event) {
 	for c := range r {
-		c <- event
+		eventCopy := copyEvent(event)
+		c <- eventCopy
 	}
 }
 
@@ -80,4 +85,18 @@ func (r registry) register(c chan<- sd.Event) {
 
 func (r registry) deregister(c chan<- sd.Event) {
 	delete(r, c)
+}
+
+// copyEvent does a deep copy on sd.Event
+func copyEvent(e sd.Event) sd.Event {
+	// observers all need their own copy of event
+	// because they can directly modify event.Instances
+	// for example, by calling sort.Strings
+	if e.Instances == nil {
+		return e
+	}
+	instances := make([]string, len(e.Instances))
+	copy(instances, e.Instances)
+	e.Instances = instances
+	return e
 }
--- a/sd/internal/instance/cache_test.go
+++ b/sd/internal/instance/cache_test.go
@@ -1,10 +1,15 @@
 package instance
 
 import (
+	"context"
+	"fmt"
+	"io"
 	"reflect"
 	"testing"
 	"time"
 
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/sd"
 )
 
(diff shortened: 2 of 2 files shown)


sd/internal/instance/cache.go
```diff
@@ -52,10 +54,12 @@ func (c *Cache) Stop() {}
...
 func (c *Cache) Register(ch chan<- sd.Event) {
 	c.mtx.Lock()
-	defer c.mtx.Unlock()
 	c.reg.register(ch)
+	event := c.state
+	c.mtx.Unlock()
+	eventCopy := copyEvent(event)
 	// always push the current state to new channels
-	ch <- c.state
+	ch <- eventCopy
```
MaruHyl: I think it's better to put this in Lock. Otherwise the receiver may get stale intances.

toffaletti: I don’t think a mutex is solving that problem. You would need updates to arrive in such a way that there would be contention on the mutex. If that is the case, it is probably just as likely you’d have updates arrive in quick succession but miss contention.

MaruHyl: Lock can ensure that events enter `chan` in order, which means  consumers can consume in the right order.
Make a simple path in your case.
Thread1: Lock -> A event -> UnLock
Thread2: Wait ->         ...                      -> Lock -> B event(newer than A) -> UnLock
Once out of Lock, consumers may consume B first and then A, no one can determine the order.

toffaletti: I see your point now. Will fix.

sd/internal/instance/cache.go
```diff
@@ -81,3 +86,20 @@ func (r registry) register(c chan<- sd.Event) {
...
+func copyEvent(e sd.Event) sd.Event {
+	// observers all need their own copy of event
+	// because they can directly modify event.Instances
+	// for example, by calling sort.Strings
+	var instances []string
+	if len(e.Instances) != 0 {
+		instances = make([]string, len(e.Instances))
+		copy(instances, e.Instances)
+	}
+	eventCopy := sd.Event{
```
MaruHyl: Maybe it is not important to new a struct here, `e` is already a copy, just `e.Instances=instances` is the same.

toffaletti: Thanks for the thorough review, I've improved the implementation of copyEvent to also handle another potential edge case where a 0 len slice with some capacity could be shared and later appended to causing data races inside the pre-allocated memory.

peterbourgon:
Can you rebase this on master?

toffaletti:
@peterbourgon done

peterbourgon:
Vielen Dank_!_

PR #868, opened 2019-04-30
Some code improvements

sosiska:
* Rewrite some if-else chains as a switch statements. Based on Go style guide: https://golang.org/doc/effective_go.html#switch
* Simplify some functions.
* Improvement of len() comparisons style, because builtin len() function can't return value smaller than zero.
* strings.ToLower comparison changed to strings.EqualFold.
* **gofmt** _examples/addsvc/thrift/gen-go/addsvc/add_service-remote/add_service-remote.go_ file (was done automatically by my IDE  ¯\\_(ツ)_/¯ )

### DIFF
--- a/auth/jwt/transport.go
+++ b/auth/jwt/transport.go
@@ -77,7 +77,7 @@ func ContextToGRPC() grpc.ClientRequestFunc {
 
 func extractTokenFromAuthHeader(val string) (token string, ok bool) {
 	authHeaderParts := strings.Split(val, " ")
-	if len(authHeaderParts) != 2 || strings.ToLower(authHeaderParts[0]) != bearer {
+	if len(authHeaderParts) != 2 || !strings.EqualFold(authHeaderParts[0], bearer) {
 		return "", false
 	}
 
--- a/examples/addsvc/thrift/gen-go/addsvc/add_service-remote/add_service-remote.go
+++ b/examples/addsvc/thrift/gen-go/addsvc/add_service-remote/add_service-remote.go
@@ -4,183 +4,182 @@
 package main
 
 import (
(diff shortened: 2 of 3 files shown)


peterbourgon:
> Rewrite some if-else chains as a switch statements. Based on Go style guide: https://golang.org/doc/effective_go.html#switch

This is a style choice, one version isn't better or worse than the other, by default. Unless there is a clear win in legibility, changing these sorts of things is relatively unproductive code churn. I didn't see anything in the diff that would obviously warrant the changes, if you believe otherwise, please give justification.

> Improvement of len() comparisons style, because builtin len() function can't return value smaller than zero

These were done deliberately. Please revert.

With those changes, happy to merge.

sosiska:
> These were done deliberately. Please revert.

Why check the obviously impossible option?

peterbourgon:
len returns an int, ints can be negative, <= is safer.

sosiska:
> len returns an int, ints can be negative, <= is safer.

I do not agree, the source code is full of places where the result of the len() is compared with 0.

peterbourgon:
If I wrote the code, I used `<=`. If other people wrote the code, I probably didn't pick any nits about it.

sosiska:
> If I wrote the code, I used `<=`. If other people wrote the code, I probably didn't pick any nits about it.

https://github.com/go-kit/kit/blame/master/metrics/internal/lv/space.go#L82  ¯\\(ツ)/¯

sosiska:
> These were done deliberately. Please revert.

Rejected in https://github.com/go-kit/kit/commit/40b2404b3da771def8eb6b7af3dc2ec3194856e8

PR #870, opened 2019-05-03
more convenient log: add FullPathCaller and TrimPathPrefixCaller

darkdarkfruit:
In real app, we might need a log which records the full path or a path with prefix trimmed.

pros && cons:
Old default caller: might be appropriate, but also might be too succinct;
FullPathCaller: verbose(good for quick debugging and jumping ) but might be too long;
TrimPathPrefixCaller: user controlled file path.

### DIFF
--- a/log/value.go
+++ b/log/value.go
@@ -92,6 +92,35 @@ func Caller(depth int) Valuer {
 	}
 }
 
+// Caller returns a Valuer that returns a file and line from a specified depth
+// in the callstack. Users will probably want to use DefaultFullPathCaller.
+func FullPathCaller(depth int) Valuer {
+	return func() interface{} {
+		_, file, line, _ := runtime.Caller(depth)
+		//idx := strings.LastIndexByte(file, '/')
+		// using idx+1 below handles both of following cases:
+		// idx == -1 because no "/" was found, or
+		// idx >= 0 and we want to start at the character after the found "/".
+		return file + ":" + strconv.Itoa(line)
+	}
+}
+
+// Caller returns a Valuer that returns a file and line from a specified depth
+// in the callstack. Users will probably want to use DefaultTrimmedPrefixPathCaller
+// eg:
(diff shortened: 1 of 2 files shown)


ChrisHines:
Thanks for the ideas.

It is true that `DefaultCaller` is very simple and not sufficient for many applications, but that's OK because the `Valuer` mechanism provides a general purpose tool for each application to provide its own implementation. I am not convinced the base Go kit log package needs to include these variants.

What do you think @peterbourgon?

peterbourgon:
I don't have a strong opinion. Could a middle ground be amending the documentation on Caller to suggest users who aren't satisfied with the default behavior duplicate and modify the function in their code?

darkdarkfruit:
just for convenient

ChrisHines:
Adding to the documentation sounds good to me.

peterbourgon:
The suggestion was to remove FullPathCaller and TrimPathPrefixCaller declarations, and add documentation to the Caller function to suggest that users can implement their own custom e.g. FullPathCaller and TrimPathPrefixCaller functions in their own code.

darkdarkfruit:
IMHO,  the convenience out of the box might be better

ChrisHines:
This package is intended to have a minimal API that provides the necessary building blocks. We don't want to add to the public API unless the new feature provides significant benefit not easily provided other ways. The package README states this up front:

>`package log` provides a minimal interface for structured logging in services. It may be wrapped to encode conventions, enforce type-safety, provide leveled logging, and so on. It can be used for both typical application log events, and log-structured data streams.

It may be true that this PR provides some additional convenience for some uses. There are many features we could add for convenience but have chosen not to to keep a minimal API. The most common additional features (levels, colored output for terminals, integrations with other log packages) live in sub directories rather than the base package because not everyone needs those features.

I can imagine many different ways that the caller information can be formatted. I wrote [github.com/go-stack/stack](https://godoc.org/github.com/go-stack/stack) for that purpose several years ago. Yet I am quite happy to keep that functionality separate from this package, since they are easily combined when needed.

darkdarkfruit:
ok

Issue #867, opened 2019-04-29
ElasticSearch support for go-kit Logger


shkov:
Do you want to make it possible to send logs to elastic by interval? Logrus allows you to do this through a `logger.Hooks.Add(hook)` and by helper packages like https://github.com/sohlich/elogrus

ChrisHines:
I think it would be interesting to see some design ideas for this. Hopefully it could be implemented as a `log.Logger` wrapper or middleware. Is this something you would like to work on and contribute?

shkov:
Yes, it's interesting for me. I will try to implement this in the near future.

shkov:
I have some design doubts.
I am going to make a `log.Logger` wrapper, which will send logs to elastic by `BulkProcessor` https://github.com/olivere/elastic/wiki/BulkProcessor. Should it look like I pass the elastic client inside the wrapper, or should I connect to the elastic inside? I think the second option is better, since the connection loop is needed because the elastic client returns an error if elastic is not available. Maybe I should pass an interface with single method like sendBatch() inside?

peterbourgon:
Although I'm happy to review a design doc and/or PR, I'm really dubious that this kind of functionality belongs in core Go kit. My intuition is that it's much better as a third-party adapter.

PR #872, opened 2019-05-06
add log.Logger wrapper

shkov:
new log.Logger wrapper for https://github.com/go-kit/kit/issues/867

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -73,6 +73,31 @@ func WithPrefix(logger Logger, keyvals ...interface{}) Logger {
 	}
 }
 
+// Flusher is the interface for sending your logs to analytical systems so
+// you can analyze and aggregate them. Implementations may involve a buffer
+// cleared by interval or sending to any database directly
+// also they must be safe for concurrent use by multiple goroutines.
+type Flusher interface {
+	Flush(keyvals ...interface{}) error
+}
+
+// WithFlusher returns a new contextual logger with passed Flusher.
+//
+// The returned Logger replaces all value elements (odd indexes) containing a
+// Valuer with their generated value for each call to its Log method.
+func WithFlusher(logger Logger, flusher Flusher) Logger {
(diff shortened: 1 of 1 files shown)


peterbourgon:
Is this not an easier approach?

```go
type FlushingLogger struct {
	next log.Logger
	flusher Flusher
}

func (fl FlushingLogger) Log(keyvals ...interface{}) error {
	fl.next.Log(keyvals...)
	fl.flusher.Flush(keyvals...)
	return nil
}
```

shkov:
Do I understand correctly that if i don’t increase the length of keyvals, then there will be no problems with context stack depth consistency?

shkov:
I just feared it https://github.com/go-kit/kit/blob/master/log/log.go#L88

ChrisHines:
I don't want to see `log.context` grow more functionality.

You don't need to worry about the stack depth problem if all the wrapping layers are built up before `log.With` is used on them.

As for the functionality in this PR, the `Flush` method introduced here has the exact same signature as the `Log` method of the `log.Logger` interface. It's not clear to me how this adds any value. An implementation of `log.Logger` can do everything that an implementation of a `log.Flusher` can. A `log.Logger` implementation that adds some buffering or writes to a DB, or whatever works just fine without introducing any new abstractions.

shkov:
It turns out that I can implement my own `log.Logger`, which can write to elasticSearch. I'm not sure now that core of Go kit needed more functionality. Maybe you would like to see the implementation of sending logs to elastic inside of go-kit or something else?

peterbourgon:
My comment in #867 stands:

> Although I'm happy to review a design doc and/or PR [for Elasticsearch support], I'm really dubious that this kind of functionality belongs in core Go kit. My intuition is that it's much better as a third-party adapter.

More concretely, I don't really want to have to maintain Elasticsearch integration tests for the lifetime of Go kit.

PR #834, opened 2019-02-14
support for Go 1.11 Modules

gilgameshskytrooper:

PR #881, opened 2019-06-06
Update imports for NATS

wallyqs:
Hi, sorry for the inconvenience but the NATS project has recently renamed some of its repos. Most of the time Github redirects would help but in case a project is using Go modules, it can help to have the correct paths in the imports just in case.

Signed-off-by: Waldemar Quevedo <wally@synadia.com>

### DIFF
--- a/transport/nats/subscriber_test.go
+++ b/transport/nats/subscriber_test.go
@@ -9,8 +9,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/nats-io/gnatsd/server"
-	"github.com/nats-io/go-nats"
+	"github.com/nats-io/nats-server/server"
+	"github.com/nats-io/nats.go"
 
 	"github.com/go-kit/kit/endpoint"
 	natstransport "github.com/go-kit/kit/transport/nats"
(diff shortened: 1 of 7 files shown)


transport/nats/subscriber_test.go
```diff
@@ -9,8 +9,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/nats-io/gnatsd/server"
-	"github.com/nats-io/go-nats"
+	"github.com/nats-io/nats-server/server"
```
derekcollison: Probably should be `"github.com/nats-io/nats-server/v2/server"` now.

peterbourgon:
@wallyqs You wanna try updating re: above?

wallyqs:
@peterbourgon I'd like to update, but thought maybe should investigate more or maybe wait for something like this to merge https://github.com/go-kit/kit/pull/834 since if there are users not using modules yet and using go-kit then I think the import `github.com/nats-server/v2/server` may break for them.
I will take a look and try to update, for now since it is only for a pub/sub test the v1 version of the server would work ok since protocol backwards compatible.

derekcollison:
This is transparently supported back to Go 1.9.7 IIRC


________________________________

PR #898, opened 2019-08-06
fix sd/etcdv3 bug

wencan:


peterbourgon:
1. Which bug?
2. This is a breaking change. Can you make it.. not a breaking change? 😇

peterbourgon:
@wencan Friendly ping re: above :)

wencan:
If a service is registered with etcd after GetEntries is executed and before WatchPrefix is executed, kit/sd/etcdv3 will not get it. 
Ci looks broken.
I can build it successfully

peterbourgon:
> If a service is registered with etcd after GetEntries is executed and before WatchPrefix is executed, kit/sd/etcdv3 will not get it.

Can you please write a test case which fails on current master and passes on your branch?

> Ci looks broken. I can build it successfully

Rebase on master and it should be fixed.

wencan:
This difficulty
I need to optimize the client class
Do you allow me to do this?

peterbourgon:
Closing due to nonresponsive contributor.

PR #912, opened 2019-09-21
Remove extra newline

LasTshaMAN:
This PR shows how CI pipeline fails because `go-kit` relies on the latest version of its dependencies.


peterbourgon:
Acknowledged. Fixed on master.

PR #914, opened 2019-09-27
Disable HTML escaping in JSON Logger

treuherz:
In #690 @ChrisHines said he'd be open to disabling HTML escaping in the JSON Logger altogether, without changing the API, but it doesn't seem like anyone got around to contributing the change. If this is still something the maintainers would be open to it'd be great to see it put in place.

### DIFF
--- a/log/json_logger.go
+++ b/log/json_logger.go
@@ -31,7 +31,9 @@ func (l *jsonLogger) Log(keyvals ...interface{}) error {
 		}
 		merge(m, k, v)
 	}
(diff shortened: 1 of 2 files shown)


treuherz:
The failures seem to be a dependency-related issue. If that gets fixed I'd be happy to rebase and resubmit

peterbourgon:
Can you rebase on master?

treuherz:
Rebased, greens across the board

Issue #796, opened 2018-11-20
Chunked transfer encoding


rampage644:
I've faced an issue similiar to what is described in #391. Namely, other 3rd party RESTful service is unable to parse incoming JSON body.

In the end, it boiled down to the inability of existing nginx/uwsgi/django setup to properly parse HTTP request without `Content-Length` header (dunno what element in the chain should be responsible for this). 

What was happening under the hood is the way go-kit creates `http.Request` and initializes its body: since `body` is nil during construction, `r.ContentLength` is not set. The solution (as in mentioned issue) is to redefine `encodeJSONRequest` function to properly set required parameter.

Here we create a request with nil body: https://github.com/go-kit/kit/blob/master/transport/http/client.go#L114. This code is skipped due to nil body: https://github.com/golang/go/blob/master/src/net/http/request.go#L827

Please advise what is the best way of dealing with this issue.

rampage644:
This is how my encode functions looks like:
```go
func encodeJSONRequest(c context.Context, r *http.Request, request interface{}) error {

       // omitting other stuff from original function

	var b bytes.Buffer
	err := json.NewEncoder(&b).Encode(request)
	r.Body = ioutil.NopCloser(&b)
	// We set content length explicitly here since go-kit constructs request
	// with nil body and thus failing to initialize some fields.
	// We can omit setting it but this results in using chunked transfer-encoding
	// that cause problems on the nginx/uwsgi/django side - it just can't
	// deal with no Content-Length header set.
	r.ContentLength = int64(b.Len())
	return err
}
```

peterbourgon:
That seems correct to me. Is there something unsatisfying about this?

rampage644:
I thought if it's worth hiding this in the go-kit default encoding function. Or maybe change the way http request is created to allow it to be fully initialized.

rampage644:
Right now I feel this leaks low-level transport details into encoder function - "should I set only the body here or I need to figure out what is expected by the underlying low-level things". Strictly speaking, original code also initializes `GetBody` function as well, and it could be confusing just to duplicate code from `src/net/http/request`.

peterbourgon:
What alternative implementation possibly exists?

LasTshaMAN:
@peterbourgon

As I currently understand the problem, `go-kit` tries to hide some boilerplate details about `*http.Request` creation from the user, yet it doesn't fully succeed in the task,

as the result, we get hard-to-debug behavior ...

[...]

PR #916, opened 2019-09-28
(#796) User should be responsible for request encoding with client endpoint

LasTshaMAN:
TODO:
- check out whether some docs require updating as well (so that they will describe new API usage  instead of the deprecated one)
- increase test coverage (especially check the part about `Content-Length`)

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -13,85 +13,23 @@ import (
 	"github.com/go-kit/kit/endpoint"
 )
 
-// HTTPClient is an interface that models *http.Client.
-type HTTPClient interface {
-	Do(req *http.Request) (*http.Response, error)
-}
-
-// Client wraps a URL and provides a method that implements endpoint.Endpoint.
-type Client struct {
-	client         HTTPClient
-	method         string
-	tgt            *url.URL
-	enc            EncodeRequestFunc
-	dec            DecodeResponseFunc
-	before         []RequestFunc
-	after          []ClientResponseFunc
-	finalizer      []ClientFinalizerFunc
-	bufferedStream bool
-}
-
-// NewClient constructs a usable Client for a single remote method.
-func NewClient(
+// NewClientEndpoint returns a usable endpoint that invokes the remote endpoint.
+func NewClientEndpoint(
 	method string,
 	tgt *url.URL,
-	enc EncodeRequestFunc,
+	enc EncodeClientRequestFunc,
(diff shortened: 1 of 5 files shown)


peterbourgon:
Thanks, I'll work on the build failures.

peterbourgon:
Can you rebase on master?

LasTshaMAN:
Sure I can rebase, will do soon!

LasTshaMAN:
Rebased.

LasTshaMAN:
@peterbourgon, have you had a chance to take a look yet ?

peterbourgon:
Sorry for the delay. We can't break the transport/http API by removing the Client type. Can you not incorporate the changes into that type?

peterbourgon:
More concretely, I would want to see EncodeClientRequestFunc (probably better named CreateRequestFunc) as a field in the Client type, with a default implementation that has the current behavior (http.NewRequest) and an Option to set it to something else.

LasTshaMAN:
> We can't break the transport/http API by removing the Client type

I think this PR does not break it, only deprecates it (at least this is how I intended it to be).

> Can you not incorporate the changes into that type? More concretely, I would want to see EncodeClientRequestFunc (probably better named CreateRequestFunc) as a field in the Client type, with a default implementation that has the current behavior (http.NewRequest) and an Option to set it to something else.

We can go this route, but I see the following major obstacle - we initialize `Client` like this:
```go
func NewClient(
	method string,
	tgt *url.URL,
	enc EncodeRequestFunc,
	dec DecodeResponseFunc,
	options ...ClientOption,
) *Client {
...
```
that is, we **require** the API consumer to pass in `enc EncodeRequestFunc`. 

So, as API consumer if I want to create my requests through `CreateRequestFunc` option (like you are suggesting above) rather than through `EncodeRequestFunc` - what value for `enc` should I pass in `NewClient` ?

LasTshaMAN:
Hey, guys, can we follow-up on this ?

ppacher:
Hi, is there any progress here? I'd be cool to get this merged.

peterbourgon:
I'm sorry for my lack of action on this PR, I simply haven't had the mental bandwidth to devote to this actually somewhat substantial change. I will make time before the end of the week to get it into shape and merged. 

@LasTshaMAN — in the interest of expediency, would it be OK with you if I made a PR to your PR to simply implement whatever changes I felt were appropriate?

PR #926, opened 2019-10-18
Fix incorrect Summary/Histogram references

jeffwidman:


peterbourgon:
These comments are not incorrect. They refer to Go kit Histograms, which map to Prometheus Summaries.

jeffwidman:
Ah, I missed https://github.com/go-kit/kit/blob/9f5354e50d79d79d865f684fe139811cf309870f/metrics/prometheus/prometheus.go#L111

That makes more sense, I was wondering why the Prometheus implementation had both summaries and histograms and yet the generic version only had histograms.

PR #927, opened 2019-10-18
Stop intermediate casting to prometheus.Labels

jeffwidman:
Prometheus already offers a `WithLabelValues()` method that takes `lvs
...string` as input.

So stop doing an intermediate cast to `prometheus.Labels`, as we simply
don't need to do that.

### DIFF
--- a/metrics/prometheus/prometheus.go
+++ b/metrics/prometheus/prometheus.go
@@ -42,7 +42,7 @@ func (c *Counter) With(labelValues ...string) metrics.Counter {
 
 // Add implements Counter.
 func (c *Counter) Add(delta float64) {
-	c.cv.With(makeLabels(c.lvs...)).Add(delta)
+	c.cv.WithLabelValues(c.lvs...).Add(delta)
 }
 
 // Gauge implements Gauge, via a Prometheus GaugeVec.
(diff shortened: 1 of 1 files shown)


jeffwidman:
It's certainly confusing to me that these metrics have a `With()` method that appears to shadow the `prometheus.With()` method, but in actuality is shadowing the `prometheus.WithLabelValues()` method.

I actually probably would have preferred to have this PR change the public API to use `prometheus.With()` because of [this note on `WithLabelValues()`](https://godoc.org/github.com/prometheus/client_golang/prometheus#CounterVec.WithLabelValues):
> Note that for more than one label value, this method is prone to mistakes caused by an incorrect order of arguments. Consider `With(Labels)` as an alternative to avoid that type of mistake. For higher label numbers, the latter has a much more readable (albeit more verbose) syntax, but it comes with a performance overhead (for creating and processing the `Labels` map). See also the `GaugeVec` example.

Personally, I think the readability and minimizing the chance of mistakes on argument order is worth the tiny hit in perf for vast majority of projects.

peterbourgon:
Your new code is not equivalent to the code it replaces. WithLabelValues elides the keys.

jeffwidman:
Ah, I missed that. Thanks, your explanation helped me understood this bit of code better.

Issue #924, opened 2019-10-17
Weighted round robin


shkov:
My `sd.Endpointer` have endpoints which call different data centers. Round Robin choses endpoints on a cyclic basis but call of some instances, placed further than other ones, takes longer time. I want the call of further endpoint only to come when near endpoint returns error. Is there way to give a priority to endpoint? If there is no realization please help me with design.

wenkaler:
You can use this function. https://gist.github.com/wenkaler/b56305db53ffa8852b8876e3e4e5627d

shkov:
@wenkaler Actually it solves only my problem but this approach doesn't allow to balance traffic by weight between instances. Also i am not sure that relying on endpoints order is a good practice.

PR #928, opened 2019-10-18
Issue 924

wenkaler:
problem solution - https://github.com/go-kit/kit/issues/924


wenkaler:
I realized that my solution does not require a merge request.

PR #941, opened 2019-12-17
Close the response body after reading from it

jeffwidman:
According to https://golang.org/pkg/net/http/, the response body should
be closed after reading from it:

```
resp, err := http.Get("http://example.com/")
if err != nil {
        // handle error
}
defer resp.Body.Close()
body, err := ioutil.ReadAll(resp.Body)
// ...
```

Note: I saw a few other places in the repo this could probably be added as well, but starting with here as I'm a `go` n00b and there's a chance I'm misunderstanding and this isn't actually required...

### DIFF
--- a/metrics/prometheus/prometheus_test.go
+++ b/metrics/prometheus/prometheus_test.go
@@ -23,6 +23,7 @@ func TestCounter(t *testing.T) {
 
 	scrape := func() string {
 		resp, _ := http.Get(s.URL)
+		defer resp.Body.Close()
 		buf, _ := ioutil.ReadAll(resp.Body)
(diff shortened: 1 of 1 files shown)


peterbourgon:
In a test context it's less relevant, because when each test function terminates everything is torn down anyway. And in the cases you touched in your commit, it's not safe, because we don't check the error response and the http.Response might be nil.

So in I'm happy to have a review of other cases where we might not be closing the response body, but in this case (so far) I think it's probably incorrect.

jeffwidman:
That sounds reasonable, thank you for educating me.

Issue #698, opened 2018-04-05
Question for rate limiter


gigimushroom:
I am trying to set up rate limiter for my grpc:
limit := rate.NewLimiter(rate.Every(time.Minute), 10)	
e = ratelimit.NewErroringLimiter(limit)(e)
pb.RegisterFlowServer(s, server.NewBinding(e))
...

My client test code is to send 10 RPC, wait a minute, then send another 10 RPC.

However, my 2nd batch failed 9 out of 10.

Any ideas?

gigimushroom:
It seems the token will never refilled itself.
```
var nopEndpoint = func(context.Context, interface{}) (interface{}, error) { return struct{}{}, nil }

func TestXRateErroring(t *testing.T) {
	limit := rate.NewLimiter(rate.Every(time.Second), 3)
	testSuccessThenFailure(
		t,
		ratelimit.NewErroringLimiter(limit)(nopEndpoint),
		ratelimit.ErrLimited.Error())
}

func testSuccessThenFailure(t *testing.T, e endpoint.Endpoint, failContains string) {
	ctx, cxl := context.WithTimeout(context.Background(), 3*time.Second)
	defer cxl()

	// First request should succeed.
	if _, err := e(ctx, struct{}{}); err != nil {
		t.Errorf("unexpected: %v\n", err)
	}
	if _, err := e(ctx, struct{}{}); err != nil {
		t.Errorf("unexpected2: %v\n", err)
	}

	if _, err := e(ctx, struct{}{}); err != nil {
		t.Errorf("unexpected3: %v\n", err)
	}

	time.Sleep(time.Second)

	if _, err := e(ctx, struct{}{}); err != nil {
		t.Errorf("unexpected4: %v\n", err)
	}

	if _, err := e(ctx, struct{}{}); err != nil {
		t.Errorf("unexpected5: %v\n", err)
	}

	if _, err := e(ctx, struct{}{}); err != nil {
		t.Errorf("unexpected6: %v\n", err)
	}
}
```

Result:
```
2018-04-04 23:46:26.134630259 -0700 PDT m=+0.000362458 0001-01-01 00:00:00 +0000 UTC 3
num of burst left 3 num of tokens 2
2018-04-04 23:46:26.13476097 -0700 PDT m=+0.000493169 2018-04-04 23:46:26.134630259 -0700 PDT m=+0.000362458 2.000130711
num of burst left 3 num of tokens 1.0001307110000002
[...]

PR #943, opened 2019-12-28
Clarify rate limiter expressions in addsvc example

marselester:
Rate limiting examples might be a bit confusing when `rate.Every` method is used https://github.com/go-kit/kit/issues/698.
Naturally one would expect to see a throttling demo as a number of requests per unit of time

```go
// 100 requests per second with burst of 100 requests.
rate.NewLimiter(rate.Limit(100), 100)

// 5 requests per minute with burst of 5 requests.
rate.NewLimiter(rate.Limit(5/time.Minute.Seconds()), 5)
```

instead of defining an interval between requests.

```go
// 1 request per second with burst of 100 requests.
rate.NewLimiter(rate.Every(time.Second), 100)

// 5 requests per minute with burst of 5 requests.
rate.NewLimiter(rate.Every(time.Minute/time.Duration(5)), 5)
```

### DIFF
--- a/examples/addsvc/pkg/addendpoint/set.go
+++ b/examples/addsvc/pkg/addendpoint/set.go
@@ -35,6 +35,8 @@ func New(svc addservice.Service, logger log.Logger, duration metrics.Histogram, 
 	var sumEndpoint endpoint.Endpoint
 	{
 		sumEndpoint = MakeSumEndpoint(svc)
+		// Sum is limited to 1 request per second with burst of 1 request.
+		// Note, rate is defined as a time interval between requests.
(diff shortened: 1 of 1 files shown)


peterbourgon:
The different ways to express rate limits and burst capacity are properties of the rate limiting package, and it's expected that users will read those docs to learn more. It's not clear to me that your changes are strictly or even generally easier to understand, so I'm -1 on a wholesale change, but I'd welcome a change to one example, and comments (like you have in the comments) on both examples explaining what the expressions mean.

marselester:
I added comments to `addsvc` example since it has two rate limiters. It should be easier to spot the difference when comparing expressions side by side.

PR #948, opened 2020-01-11
Add a span name getter to opencensus endpoint options

sagikazarmark:
This is an experimental PR suggested by @basvanbeek in https://github.com/go-kit/kit/pull/946#issuecomment-573043128

It adds a span name getter function to the opencensus endpoint tracer config.

Other options might worth adding:

- StartOptions
- GetStartOptions

See the opencensus ochttp plugin for more.

### DIFF
--- a/tracing/opencensus/endpoint.go
+++ b/tracing/opencensus/endpoint.go
@@ -31,6 +31,10 @@ func TraceEndpoint(name string, options ...EndpointOption) endpoint.Middleware {
 
 	return func(next endpoint.Endpoint) endpoint.Endpoint {
 		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
+			if name == TraceEndpointDefaultName && cfg.GetSpanName != nil {
+				name = cfg.GetSpanName(ctx)
+			}
+
 			ctx, span := trace.StartSpan(ctx, name)
 			if len(cfg.Attributes) > 0 {
 				span.AddAttributes(cfg.Attributes...)
--- a/tracing/opencensus/endpoint.go
+++ b/tracing/opencensus/endpoint.go
@@ -31,13 +31,24 @@ func TraceEndpoint(name string, options ...EndpointOption) endpoint.Middleware {
 
 	return func(next endpoint.Endpoint) endpoint.Endpoint {
 		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
-			if name == TraceEndpointDefaultName && cfg.GetSpanName != nil {
-				name = cfg.GetSpanName(ctx)
+			var attributes []trace.Attribute
+
+			if cfg.GetSpanDetails != nil {
+				var newName string
+
+				newName, attributes = cfg.GetSpanDetails(ctx, name)
+
+				if newName != "" {
+					name = newName
+				}
 			}
 
 			ctx, span := trace.StartSpan(ctx, name)
 			if len(cfg.Attributes) > 0 {
 				span.AddAttributes(cfg.Attributes...)
+			}
+			if len(attributes) > 0 {
+				span.AddAttributes(attributes...)
 			}
 			defer span.End()
 
--- a/tracing/opencensus/endpoint.go
+++ b/tracing/opencensus/endpoint.go
@@ -32,9 +32,7 @@ func TraceEndpoint(name string, options ...EndpointOption) endpoint.Middleware {
 	return func(next endpoint.Endpoint) endpoint.Endpoint {
 		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
 			if cfg.GetName != nil {
-				newName := cfg.GetName(ctx, name)
-
-				if newName != "" {
+				if newName := cfg.GetName(ctx, name); newName != "" {
 					name = newName
 				}
 			}
--- a/tracing/opencensus/endpoint.go
+++ b/tracing/opencensus/endpoint.go
@@ -44,9 +44,8 @@ func TraceEndpoint(name string, options ...EndpointOption) endpoint.Middleware {
 			defer span.End()
 
 			if cfg.GetAttributes != nil {
-				attributes := cfg.GetAttributes(ctx)
-				if len(attributes) > 0 {
-					span.AddAttributes(attributes...)
+				if attrs := cfg.GetAttributes(ctx); len(attrs) > 0 {
+					span.AddAttributes(attrs...)
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


tracing/opencensus/endpoint_options.go
```diff
@@ -11,6 +15,11 @@ type EndpointOptions struct {
 	// Attributes holds the default attributes which will be set on span
 	// creation by our Endpoint middleware.
 	Attributes []trace.Attribute
+
+	// GetSpanName holds the function to use for generating the span name
+	// from the information found in the incoming Request.
+	// Defaults to the name that the middleware was initialized with.
+	GetSpanName func(ctx context.Context) string
```
sagikazarmark: Should we add a functional option for it? Like `WithSpanNameGetter`?

sagikazarmark:
What are your thoughts about this? Is this what you had in mind?

peterbourgon:
@basvanbeek

tracing/opencensus/endpoint_options.go
basvanbeek: Since the other options have them I would say whatever we decide to add here should have a corresponding functional option.

That said I feel like we're adding something with limited value and could improve.

Some gaps to consider:
1. we should allow for a name regardless of being set explictly or not to be overridable
2. we should allow for a name regardless of being set explictly or not to be kept as is
3. we should allow for attributes to be set given the metadata as found in ctx.

Given the above gaps, how about we do this:
```go
GetSpanDetails func(ctx context.Context, name string) (newName string, attrs []trace.Attribute)
```

sagikazarmark: Works for me, although that would be inconsistent with how the official OpenCensus plugins allow setting a name for a span.

It's also a bit error prone, because if you only want to return attributes and not a name you have to be aware that the returned name will always override any previously set one.

basvanbeek: I don't think that is a concern as long as you in the handling of the function test for empty (which could make it revert to previously set)

We're not trying to be inline with regular OpenCensus plugins here because the Endpoint middleware is distinctly different from those (they have no concept of a local span middleware)

sagikazarmark:
@basvanbeek I've updated the PR based on your comments.

However, I feel it would be better to split the function into two:

- GetSpanName
- GetAttributes

More functions, more work, but they are more expressive IMHO.

basvanbeek:
We can do the split... we can also say that if a returned name is empty on a call to `GetName` we keep the existing name for the endpoint

sagikazarmark:
Updated the PR accordingly. If you have any ideas for better naming, let me know.

tracing/opencensus/endpoint.go
```diff
@@ -31,12 +31,27 @@ func TraceEndpoint(name string, options ...EndpointOption) endpoint.Middleware {
 
 	return func(next endpoint.Endpoint) endpoint.Endpoint {
 		return func(ctx context.Context, request interface{}) (response interface{}, err error) {
+			if cfg.GetName != nil {
+				newName := cfg.GetName(ctx, name)
```
basvanbeek: cleaner:

```go
if newName := cfg.GetName(ctx, name); newName != "" {
    name = newName
}
```

sagikazarmark: That panics if GetName is nil

basvanbeek: I meant inside the test for nil

sagikazarmark: Ah, ok! Makes sense

tracing/opencensus/endpoint_options.go
```diff
@@ -11,6 +15,16 @@ type EndpointOptions struct {
 	// Attributes holds the default attributes which will be set on span
 	// creation by our Endpoint middleware.
 	Attributes []trace.Attribute
+
+	// GetName holds the function used for generating the span name
+	// based on the current name and from the information found in the incoming Request.
+	//
+	// If the returned name is empty, the existing name for the endpoint is kept.
+	GetName func(ctx context.Context, name string) string
```
peterbourgon: Is "the existing name for the endpoint" what is passed for the name parameter to GetName? If so,

> GetName is an optional function that can set the span name based on the existing name for the endpoint and information in the context. If the function is nil, or the returned name is empty, the existing name for the endpoint is used.

tracing/opencensus/endpoint_options.go
```diff
@@ -11,6 +15,16 @@ type EndpointOptions struct {
...
+
+	// GetAttributes holds the function used for extracting additional attributes
+	// from the information found in the incoming Request.
+	GetAttributes func(ctx context.Context) []trace.Attribute
```
peterbourgon: I believe attributes are just attached to the span(s), is that correct? Are these attributes used instead of, or in addition to, the Attributes field?

sagikazarmark: Yes, they are attached, in addition to any other attributes.

tracing/opencensus/endpoint.go
```diff
@@ -31,12 +31,25 @@ func TraceEndpoint(name string, options ...EndpointOption) endpoint.Middleware {
...
+
 			ctx, span := trace.StartSpan(ctx, name)
 			if len(cfg.Attributes) > 0 {
 				span.AddAttributes(cfg.Attributes...)
 			}
 			defer span.End()
 
+			if cfg.GetAttributes != nil {
+				attributes := cfg.GetAttributes(ctx)
+				if len(attributes) > 0 {
```
peterbourgon:
```go
if attrs := cfg.GetAttributes(ctx); len(attrs) > 0 {
	span.AddAttributes(attrs...)
}
```

tracing/opencensus/endpoint_options.go
peterbourgon: > GetAttributes is an optional function that can extract trace attributes from the context and add them to the span.

tracing/opencensus/endpoint.go
sagikazarmark: Done

tracing/opencensus/endpoint_options.go
```diff
@@ -11,6 +15,15 @@ type EndpointOptions struct {
...
 	// creation by our Endpoint middleware.
 	Attributes []trace.Attribute
+
+	// GetName is an optional function that can set the span name based on the existing name
+	// for the endpoint and information in the context.
+	//
+	// If the function is nil, or the returned name is empty, the existing name for the endpoint is used.
+	GetName func(ctx context.Context, name string) string
+
+	// GetAttributes is an optional function that can extract trace attributes from the context and add them to the span.
```
peterbourgon:
```suggestion
	// GetAttributes is an optional function that can extract trace attributes 
	// from the context and add them to the span.
```

PR #956, opened 2020-02-14
Fix typo: "thte" -> "the"

HCoban:


ChrisHines:
Thanks for fixing that.

PR #957, opened 2020-02-17
Add GitHub Actions workflow

sagikazarmark:
See them in action at https://github.com/sagikazarmark/kit/actions

### TODOs

- [x] Register the repo on Codecov at https://codecov.io/gh/go-kit/kit (@peterbourgon)
- [x] Add repo token at https://github.com/go-kit/kit/settings/secrets as `CODECOV_TOKEN` (@peterbourgon)
- [x] Remove `github-actions` branch target (@sagikazarmark)
- [x] Cleanup old build files once approved (@sagikazarmark)

### Notes

- Coverage upload does not work with forks when using GH secrets: https://github.com/codecov/codecov-action/issues/29


.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
peterbourgon: Any reason this can't be just

```yaml
on: push
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+        env:
+          ETCD_ADDR: http://localhost:${{ job.services.etcd.ports[2379] }}
+          CONSUL_ADDR: localhost:${{ job.services.consul.ports[8500] }}
+          ZK_ADDR: localhost:${{ job.services.zk.ports[2181] }}
+          EUREKA_ADDR: http://localhost:${{ job.services.eureka.ports[8761] }}/eureka
+        run: go test -v -race -coverprofile=coverage.coverprofile -covermode=atomic -tags integration ./...
+
+      - name: Upload coverage
+        uses: codecov/codecov-action@v1
+        if: always()
```
peterbourgon: ?

.github/workflows/ci.yml
sagikazarmark: It's not happening, because the workflow file is not in the repo yet. If both `push` and `pull_request` triggers are enabled, GH will run two actions for branches+PRs pushed by you. If that doesn't bother you, feel free to remove the branch filter.

.github/workflows/ci.yml
sagikazarmark: Upload coverage (if any), even if something went wrong in the previous steps. It might not make much sense in this case though. I usually run integration and unit tests separately and I upload coverage, even if integration tests fail. Feel free to remove it if it bothers you, but I think that's a good default.

peterbourgon: I mean, can't we just remove the if directive altogether?

sagikazarmark: In this case, it probably is. `if: always()` means the given block should run, even if a previous one failed. The default behavior is to stop after the first failure.

Since the previous step is running tests itself it probably has no added value. But if you ever add a step between coverage upload and test running, it would make sense to add it back, if you want to upload coverage, even if something went wrong. That's why I said it's a sane default.

peterbourgon: Ah, OK, I see. I don't think I want code coverage to be updated if tests fail. In that case, can it be removed?

.github/workflows/ci.yml
peterbourgon: Ah, I see. OK, this way is better 👍

.github/workflows/ci.yml
sagikazarmark: Done

sagikazarmark:
Anything else that needs to be done here?

peterbourgon:
Sorry! Let's try it. Just merge and see, right?

sagikazarmark:
Yeah

PR #961, opened 2020-02-21
Log CloudWatch response error

Trane9991:
Condition to log CloudWatch Response Error is broken, as a result, the error is never logged.
This PR fixes the condition


peterbourgon:
Thanks. Is it possible to write a test which failed on the previous behavior and works with the fixed behavior?

Trane9991:
@peterbourgon done!

peterbourgon:
Thank you!

PR #962, opened 2020-02-21
allow to use specific logrus level in the adaptor

Trane9991:


log/logrus/logrus_logger.go
```diff
@@ -12,13 +12,20 @@ import (
...
 
 // NewLogrusLogger returns a go-kit log.Logger that sends log events to a Logrus logger.
 func NewLogrusLogger(logger logrus.FieldLogger) log.Logger {
-	return &logrusLogger{logger}
+	return &logrusLogger{logger, logrus.InfoLevel}
+}
+
+// NewLogrusLoggerWithLevel returns a go-kit log.Logger that sends log events to a Logrus logger, which
+// will be logged with provided error level
+func NewLogrusLoggerWithLevel(logger logrus.FieldLogger, level logrus.Level) log.Logger {
```
ChrisHines: For consistency with other Go kit packages we should use the functional option pattern for this.

Trane9991: Done, please check :)

log/logrus/logrus_logger.go
```diff
@@ -14,18 +14,30 @@ type logrusLogger struct {
...
+	return l
 }
 
-// NewLogrusLoggerWithLevel returns a go-kit log.Logger that sends log events to a Logrus logger, which
-// will be logged with provided error level
-func NewLogrusLoggerWithLevel(logger logrus.FieldLogger, level logrus.Level) log.Logger {
-	return &logrusLogger{logger, level}
+// WithLevel allows to set specific log level
+// to log messages with
+func WithLevel(level logrus.Level) option {
```
ChrisHines: The option type should be exported. It is not a good practice to have exported APIs that depend on unexported types.

Also, consider rephrasing the doc comment on `WithLevel` to start with the phrase: "WithLevel configures a logrus logger to ..."

peterbourgon: I agree the Option type should be exported, but then its parameter should also be exported. logrus.LogrusLogger stutters, so it should probably be renamed to logrus.Logger, and the constructor to simply NewLogger.

log/logrus/logrus_logger.go
```diff
@@ -10,18 +10,38 @@ import (
...
+	for _, optFunc := range options {
+		optFunc(l)
+	}
+
+	return l
 }
 
-func (l logrusLogger) Log(keyvals ...interface{}) error {
+// WithLevel configures a logrus logger to set specific log
+// level to log messages with
```
ChrisHines: This comment starts out ok, but seems to leave something dangling at the end. How about this:

```
// WithLevel configures a logrus logger to log at level for all events.
```

log/logrus/logrus_logger_test.go
```diff
@@ -53,3 +55,66 @@ func TestLogrusLogger(t *testing.T) {
...
+	for _, tt := range tests {
+		buf := &bytes.Buffer{}
+		logrusLogger := logrus.New()
+		logrusLogger.Out = buf
+		logrusLogger.Level = tt.level
+		logrusLogger.Formatter = &logrus.JSONFormatter{}
+		logger := log.NewLogger(logrusLogger, log.WithLevel(tt.level))
+
+		t.Run(tt.name, func(t *testing.T) {
+			if err := logger.Log(time.Now().String(), "amazing", "tests"); err != nil {
```
ChrisHines: This `Log` call seems to be missing either a key or a value. I also wonder why pass these keyvals here when they aren't checked later. I don't think they are needed to test the `WithLevel` functionality, so maybe it's more clear to just remove them.

log/logrus/logrus_logger.go
```diff
@@ -10,18 +10,37 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type logrusLogger struct {
+type Logger struct {
 	logrus.FieldLogger
+	logrus.Level
```
ChrisHines: I just noticed that these fields are embedded. Do we want their methods added to the method set of `Logger`? Now that `Logger` is exported, that also means these fields are accessible outside this package. I don't think that is intended. It certainly opens the potential for other code to modify the level while other code is reading it in the `Log` method in a racy way.

I think these should have unexported field names.

What do you think @peterbourgon ?

peterbourgon: I agree.

peterbourgon:
```suggestion
	level logrus.Level
```

log/logrus/logrus_logger.go
```diff
@@ -10,18 +10,37 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type logrusLogger struct {
+type Logger struct {
 	logrus.FieldLogger
```
peterbourgon:
```suggestion
	field logrus.FieldLogger
```

log/logrus/logrus_logger.go
```diff
@@ -10,18 +10,37 @@ import (
...
 }
 
+type Option func(*Logger)
+
 var errMissingValue = errors.New("(MISSING)")
 
-// NewLogrusLogger returns a go-kit log.Logger that sends log events to a Logrus logger.
-func NewLogrusLogger(logger logrus.FieldLogger) log.Logger {
-	return &logrusLogger{logger}
+// NewLogger returns a go-kit log.Logger that sends log events to a Logrus logger.
```
peterbourgon:
```suggestion
// NewLogger returns a Go kit log.Logger that sends log events to a logrus.Logger.
```

Trane9991:
Sorry for the delay. Fixed!

ChrisHines:
LGTM, thanks for the contribution!

PR #971, opened 2020-03-20
transport/http: NewExplicitClient

### LINKED ISSUE #796: Chunked transfer encoding

peterbourgon:
Closes #796, alternative to #916.

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -21,9 +21,7 @@ type HTTPClient interface {
 // Client wraps a URL and provides a method that implements endpoint.Endpoint.
 type Client struct {
 	client         HTTPClient
-	method         string
-	tgt            *url.URL
-	enc            EncodeRequestFunc
+	req            CreateRequestFunc
 	dec            DecodeResponseFunc
 	before         []RequestFunc
 	after          []ClientResponseFunc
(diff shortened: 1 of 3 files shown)


transport/http/encode_decode.go
```diff
@@ -17,6 +17,12 @@ type DecodeRequestFunc func(context.Context, *http.Request) (request interface{}
 // encodes the object directly to the request body.
 type EncodeRequestFunc func(context.Context, *http.Request, interface{}) error
 
+// CreateRequestFunc creates an outgoing HTTP request based on the passed
+// request object. It's designed to be used in HTTP clients, for client-side
+// endpoints. It's a more powerful version of EncodeRequestFunc, and can be used
+// if more fine-grained control of the HTTP request is required.
+type CreateRequestFunc func(context.Context, interface{}) (*http.Request, error)
```
LasTshaMAN: Don't you want to mark `EncodeRequestFunc` as `// deprecated` linking [this issue](https://github.com/go-kit/kit/issues/796) for more information ?

transport/http/client.go
```diff
@@ -207,3 +198,22 @@ func EncodeXMLRequest(c context.Context, r *http.Request, request interface{}) e
 	r.Body = ioutil.NopCloser(&b)
 	return xml.NewEncoder(&b).Encode(request)
 }
+
+//
+//
+//
+
+func makeCreateRequestFunc(method string, target *url.URL, enc EncodeRequestFunc) CreateRequestFunc {
```
LasTshaMAN: Maybe add some meaningful comments here instead ? :)

transport/http/encode_decode.go
peterbourgon: No, because I don't consider it deprecated, this is just an alternative.

transport/http/client.go
peterbourgon: It's an unexported function, and I find it pretty self-explanatory.

peterbourgon:
Cool, thanks!

Issue #843, opened 2019-03-04
Go kit: the road ahead


peterbourgon:
# Go kit: the road ahead

- This is a tracking issue for broad changes to the Go kit project
- Expect edits and updates
- No ETA at this time

## Context

- The microservices ecosystem has matured since Go kit's introduction
- Many responsibilities have, sensibly, moved to the platform/infrastructure
- Go kit has, accordingly, reduced its scope of responsibilities in practice
- Go kit is less about specific components or integrations, more about good application architecture
- The next iteration of the project should reflect this reality
- Keep and refocus on the good and useful parts, deprecate the less-useful parts

## Keep

- package transport - necessary and valuable
- package endpoint - required for service/transport interop, pending generics
- package log - has clearly proved its value in the test of time
- package tracing - has also proved its value
- package examples - move to new or individual repos

## Undecided

- package metrics - pending broader re-think and refactor
- kitgen - ownership is unclear
- auth - I can't easily judge usefulness; seeking comment

## Sunset 

- package ratelimit - less useful in modern platforms, easy to bolt-on with generic middleware
- package circuitbreaker - less useful in modern platforms, easy to bolt-on with generic middleware
- package sd - less useful in modern platforms, the source of most test flakes, a maintenance headache

## Additional changes

- Rewrite website with new focus on architectural design and best practice
- New, expanded tutorial(s) based on lessons learned in my consulting work

## Coda: repos

- One option is to break each package (or package-group) into its own repo under the go-kit org
- Another is to retain the existing kit repo, and use version tagging
- The latter is slightly more appealing to me, for continuity and visibility reasons

## Coda: generics

- Generics, if/when they come, will change relationship between transport and service
- I expect the endpoint layer will no longer be necessary
- It's too early to say more

sagikazarmark:
> package metrics - pending broader re-think and refactor

With OpenCensus and OpenTelemetry the abstraction is kinda moved to the platform layer. I would say the same is true for tracing.


> package auth - I can't easily judge usefulness; seeking comment

I think the implemented auth methods are useful, but need a bit of work. They are also great reference implementations for how auth in a go-kit environment should work. Since auth (sadly) is rarely as simple as basic auth or JWT, chances are people will use customized versions (for example I extended the jwt package with additional token validation).


> Do transport- and service-level errors need to be kept so distinct?

This is a huge pain point for me. Not the separation of transport and service level errors, but the fact that the way they are handled is quite inconsistent.

[...]

PR #976, opened 2020-04-23
Metrics/librato

Trane9991:
This PR adds [librato](https://www.librato.com/) client.
Important, this is a client for [Librato **source** metrics](http://api-docs-archive.librato.com/) and **NOT** [new Librato **tagged** metrics](https://www.librato.com/docs/api/).
The problem that it is impossible to migrate fro **source** to **target** metrics without of creation of a new Librato account (and data loss), so plenty of people uses **source** metrics.
However, API seems pretty similar, so this package potentially may be modified to use any of the APIs.

Some additional fixes:

1. `FillCounter` could generate `n == 0` which didn't add any values to the counter, which made some tests unreliable.

### DIFF
--- a/metrics/librato/annotations.go
+++ b/metrics/librato/annotations.go
@@ -1,0 +1,47 @@
+package librato
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+const annotationsURL = "https://metrics-api.librato.com/v1/annotations"
+
+var ErrNoNameAnnotation = fmt.Errorf("Annotation must have name")
+
+// Annotation is a representation of librato annotation object
+// https://www.librato.com/docs/kb/visualize/annotations/
+type Annotation struct {
+	Title       string  `json:"title"`
+	Source      *string `json:"source"`
+	Description *string `json:"description"`
+	Links       []Link  `json:"links"`
+
+	StartTime *int64 `json:"start_time"`
+	EndTime   *int64 `json:"end_time"`
+}
+
+// Link is a representation of link object, that can be used in annotations
+// https://www.librato.com/docs/api/#update-an-annotation
+type Link struct {
+	Relationship string  `json:"rel"`
+	URL          string  `json:"href"`
+	Label        *string `json:"label"`
+}
+
+// PostAnnotation sends annotation to librato API right away
+// because Annotation to doesn't seem to support batching
+// http://api-docs-archive.librato.com/#create-an-annotation
+func (lb *Librato) PostAnnotation(body *Annotation, name string) error {
+	if name == "" {
+		return ErrNoNameAnnotation
+	}
+
+	b, err := json.Marshal(body)
+	if nil != err {
+		return err
+	}
+
+	return lb.makeRequest(bytes.NewBuffer(b), fmt.Sprintf("%s/%s", annotationsURL, name))
+}
--- a/metrics/librato/librato.go
+++ b/metrics/librato/librato.go
@@ -1,0 +1,557 @@
+package librato
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/generic"
+	"github.com/go-kit/kit/metrics/internal/lv"
+)
+
+const (
+	maxConcurrentRequests = 20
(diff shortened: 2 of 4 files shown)


peterbourgon:
> Comment to the `Add` function says:
> // Add locates the time series identified by the name and label values in
> // the vector space, and appends the delta to the last value in the list of
> // observations.
> 
> however, real behaviour is different. Instead of incrementing existing observation, it creates new incremented observation.

The comment is wrong but the behavior is correct: Add _should_ create a new incremented observation.

Trane9991:
OK, I will change it back.
Can you please explain a bit, why is it like that?

My use case:
I see that most of the Gauges implementation just take the `last(values)`.
Since librato allows us to send batch metrics, I wanted to keep this feature and allow to send the batch of metrics added by `gauge.Set(val)`, however, this makes `gauge.Add(val)` useless, because it will send the batch of ever incrementing metrics. 

So I was thinking of the way as `Add` should work (at least as I imagine it).
So basically if I wanted to have aggregated sum of values over time, I would use `Add(val)` and if I wanted to report batch of metrics, I would use `Sum(val)`.  But since `Add(val)` would generate array of ever-increasing metrics and it will be sent to librato. So that makes and `Add(val)` useless for the batch gauge implementation.

peterbourgon:
> Can you please explain a bit, why is it like that?

Each point in the lv vectorspace represents not a single value but a timeseries, a sequence of values collected over time. Operations on points in the vectorspace (Set, Add, Observe, ...) always _append_ values to the timeseries. The Add method is a bit unique in that the value that it appends is a function of the previous value, which is why that extra code is necessary.

> Since librato allows us to send batch metrics, I wanted to keep this feature and allow to send the batch of metrics added by gauge.Set(val), however, this makes gauge.Add(val) useless, because it will send the batch of ever incrementing metrics.

I don't quite understand the point you're making. Sending one metric or sending a batch of metrics shouldn't depend on how Go kit represents timeseries data in memory. Specifically, with a gauge, if all you care about is the current value at a given time, and you don't care about the history of those values, then you can just take the last value in the `observations []float64`.

peterbourgon:
Should you even be using the `lv` package? Does Librato provide Gauge, Counter, and Histogram-like data types that can easily be coerced to the Go kit interface expectations?

Trane9991:
Librato has the support of `Counters` and `Gauges`.
I took the implementation of `cloudwatch` as an example.
For the histogram, Librato doesn't have it, but I like the idea of having percentiles, so I also took an example of `cloudwatch` implementation where Histogram used to generate multiple gauges with percentiles.

peterbourgon:
So maybe just use Librato Counters and Gauges directly, without lv stuff?

Trane9991:
I also wanted to be able to do some client-side aggregation, to be able to save a bit on the number of metrics/requests sent to librato. All the lv stuff seemed handy for that purpose.

For some metrics, I need to be able to do `average` over the period of time, for some `sum` and for others to send them as is.

So for averages, I used `generic.NewSimpleHistogram()` on top of separate `averages lv.Space` and hoped to be able to use `gauge.Add()` for sum and `gauge.Set` for the rest.

One of the solutions I had in mind is creating similar `sums lb.Space` which will be aggregated as sum before sending, do you think that will be a better approach?

Trane9991:
I reverted that commit about `Add()` function, also, refactored the gauges a bit.

So, for `Gauge` I implemented used [Librato Gauge Aggregation](http://api-docs-archive.librato.com/#gauge-specific-parameters). However, for collecting metrics this way, the `gauge.Add()` function is meaningless, so I had to mock it, so `gauge.Set()` and `gauge.Add()` both will behave as `gauge.Set()`.

But I still need the Gauge that can be aggregated as sum of values over period of time. In ordered to do so, I added new space `sums  *lv.Space`. However, for that, I had to mock `gauge.Set()`, since in this case using it has no sense as well. So fo for this gauge `gauge.Set()` and `gauge.Add()` both will behave as `gauge.Add()`.

Please, let me know, what you think, and I will try to test it on real librato account tomorrow.

Trane9991:
Hey, @peterbourgon I tested this on the real account, added some comments and improvements and now it is ready! Please, take a look when you have time. Thanks!

Trane9991:
Hey @peterbourgon.
Kind reminder about this PR :)

sagikazarmark:
@Trane9991 I don't want to speak for @peterbourgon , but lately Go kit doesn't accept any more changes like this in the core. There is no reason why a library like this would have to live there.

So I'd suggest releasing this as a separate library and then open a PR to the README to link to your project.

Also see #843 for future plans.

peterbourgon:
Second @sagikazarmark. Sorry.

PR #982, opened 2020-05-15
util/conn: close old connection before reconnect

chikaku:


peterbourgon:
If the caller puts a non-nil error, the connection is assumed to be terminated. Still, this should be a no-op, so 🤷

Issue #986, opened 2020-05-21
Which version of jsonrpc protocol does go-kit support?


lazuka13:
Hello! I tried to use go-kit jsonrpc client implementation to work with flask-jsonrpc server and ran into a problem with "jsonrpc" field - it is set to an empty string, when flask-jsonrpc expects it to be an adequate value, like "2.0" or "1.1". Also, according to specification on https://www.jsonrpc.org/specification, only version 2.0 request should have "jsonrpc" field, version 1.0 doesn't have it.

So, my question is which version of jsonrpc protocol does go-kit provide, and if the answer is "2.0", then why does it leave "jsonrpc" field empty (and shouldn't there be some convenient method to set it)?

I don't really have much experience with go, so if I don't understand something really simple here - please, explain it to me. 😄

PR #990, opened 2020-06-12
Add Version to JSON-RPC client request

shirolimit:
This tiny PR addresses the issue with JSON-RPC client: https://github.com/go-kit/kit/issues/986

For some reason current implementation leaves `jsonrpc` field empty and there is no easy way to change it (there is a workaround with `ClientBefore` which is not a good option). Some server implementations require this field to be `2.0`.

According to the documentation go-kit implementation only supports v2.0, that is why I don't think this parameter should be configurable.


peterbourgon:
Thanks!

Issue #991, opened 2020-06-12
Enhancement: log.WithSuffix


vinayvinay:
Contrary to [`log.WithPrefix`](https://godoc.org/github.com/go-kit/kit/log#WithPrefix), which prepends keyvals to those passed to `Log`, we have a requirement to _append_ keyvals.

Especially if a human is reading logfmt output on the shell or in the web browser, it helps to have certain fields prefixed, e.g. `ts`, but certain fields are better when suffixed, e.g. `caller`, `environment`, so that more important information appears first, e.g.:

```
// WithPrefix: ts, caller, service, environment
// Log: event
ts=2020-06-12T13:20:52.417486636Z caller=run.go:217 service=push-service environment=staging event=step.started
```

could be better `WithSuffix`:
```
// WithPrefix: ts
// WithSuffix: caller, service, environment
// Log: event
ts=2020-06-12T13:20:52.417486636Z event=step.started caller=run.go:217 service=push-service environment=staging
```

Here's what [it might look like in code](https://github.com/go-kit/kit/commit/3067c4ab6964fde6f73aec0bdd7a89fe38e01173), which is just a draft, will polish it up and PR if there's interest. Do you think `WithSuffix` will come handy?

vinayvinay:
Hello @peterbourgon @ChrisHines, do you have any thoughts on this one?

peterbourgon:
I don't have strong opinions, but I'm not immediately convinced that this is useful enough to justify the complexity. I'm also a little concerned there might be a performance impact. Could you add a few test cases so I can see the behavior? Mixing and matching `log.WithPrefix`, `log.WithSuffix`, and `log.With`.

peterbourgon:
Another reason this is of indeterminate value is that not all output formats respect order (e.g. JSON).

ChrisHines:
My apologies for the late response.

My first reaction when I saw this feature request was skepticism. After a little thinking about what the implementation might look like I was less concerned  because in my mind it looked pretty much like what you wrote in your prototype and it doesn't feel too costly. I would like to see some benchmarks but it seems like it shouldn't add much overhead for people that never use it.

My biggest concern at this point is whether it pays for the cost of an additional exported function in the API. It may, since I cannot think of a good way to implement it externally and get semantics similar to to `With` and `WithPrefix`.

My vote is to proceed with a PR and benchmarking.

vinayvinay:
Thank you for these inputs, fair points. I will come back with a PR with benchmarks.

PR #992, opened 2020-06-18
log.WithSuffix, appends kvs to those passed to Log

vinayvinay:
enhancement suggested in: https://github.com/go-kit/kit/issues/991

**why?**

if a human is reading logfmt output on the shell or in the web browser, it helps to have certain fields prefixed, e.g. `ts`, but certain fields are better suffixed, e.g. `caller`, `environment`, so that more important information appears first.

benchmarks suggest an additional cost only if `WithSuffix` is used.

```
goos: darwin
goarch: amd64
pkg: github.com/go-kit/kit/log
BenchmarkDiscard-4               	32230156	        38 ns/op	      32 B/op	       1 allocs/op
BenchmarkOneWith-4               	 9647907	       122 ns/op	      96 B/op	       2 allocs/op
BenchmarkTwoWith-4               	 8935790	       134 ns/op	     160 B/op	       2 allocs/op
BenchmarkTenWith-4               	 5016836	       236 ns/op	     672 B/op	       2 allocs/op
BenchmarkOneWithPrefix-4         	 9907198	       123 ns/op	      96 B/op	       2 allocs/op
BenchmarkTenWithPrefix-4         	 5076309	       239 ns/op	     672 B/op	       2 allocs/op
BenchmarkOneWithSuffix-4         	 6432942	       189 ns/op	     128 B/op	       3 allocs/op
BenchmarkTenWithSuffix-4         	 4899711	       246 ns/op	     416 B/op	       3 allocs/op
BenchmarkOneWithPrefixSuffix-4   	 6111750	       197 ns/op	     160 B/op	       3 allocs/op
BenchmarkTenWithPrefixSuffix-4   	 2172066	       555 ns/op	    1952 B/op	       3 allocs/op
PASS
ok  	github.com/go-kit/kit/log	14.224s
```

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -16,8 +16,8 @@ type Logger interface {
 var ErrMissingValue = errors.New("(MISSING)")
 
 // With returns a new contextual logger with keyvals prepended to those passed
-// to calls to Log. If logger is also a contextual logger created by With or
-// WithPrefix, keyvals is appended to the existing context.
+// to calls to Log. If logger is also a contextual logger created by With,
+// WithPrefix, or WithSuffix, keyvals is appended to the existing context.
 //
 // The returned Logger replaces all value elements (odd indexes) containing a
 // Valuer with their generated value for each call to its Log method.
@@ -36,14 +36,16 @@ func With(logger Logger, keyvals ...interface{}) Logger {
 		// backing array is created if the slice must grow in Log or With.
 		// Using the extra capacity without copying risks a data race that
 		// would violate the Logger interface contract.
-		keyvals:   kvs[:len(kvs):len(kvs)],
-		hasValuer: l.hasValuer || containsValuer(keyvals),
+		keyvals:    kvs[:len(kvs):len(kvs)],
+		hasValuer:  l.hasValuer || containsValuer(keyvals),
+		sKeyvals:   l.sKeyvals,
+		sHasValuer: l.sHasValuer,
 	}
 }
 
 // WithPrefix returns a new contextual logger with keyvals prepended to those
 // passed to calls to Log. If logger is also a contextual logger created by
-// With or WithPrefix, keyvals is prepended to the existing context.
+// With, WithPrefix, or WithSuffix, keyvals is prepended to the existing context.
 //
 // The returned Logger replaces all value elements (odd indexes) containing a
 // Valuer with their generated value for each call to its Log method.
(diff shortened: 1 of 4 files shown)


log/log.go
```diff
@@ -67,9 +68,43 @@ func WithPrefix(logger Logger, keyvals ...interface{}) Logger {
...
+	n := len(l.suffixKeyvals) + len(keyvals)
+	if len(keyvals)%2 != 0 {
+		n++
+	}
+	kvs := make([]interface{}, 0, n)
+	kvs = append(kvs, keyvals...)
+	if len(kvs)%2 != 0 {
+		kvs = append(kvs, ErrMissingValue)
+	}
+	kvs = append(l.suffixKeyvals, kvs...)
```
vinayvinay, line 102: This behaviour is common across `WithPrefix` and `WithSuffix`, keen to hear your thoughts on whether we should extract a function, or keep it inlined?

log/log_test.go
```diff
@@ -189,3 +218,75 @@ func BenchmarkTenWith(b *testing.B) {
...
+	for i := 1; i < 10; i++ {
+		lc = log.WithPrefix(lc, "a", "first")
+		lc = log.WithSuffix(lc, "z", "last")
+	}
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		lc.Log("k", "v")
+	}
+}
```
vinayvinay, line 292: Wondered if these benchmarks are what you were looking for, or more/less are needed?

log/log_test.go
```diff
@@ -56,6 +56,35 @@ func TestContextMissingValue(t *testing.T) {
 	}
 }
 
+func TestWithPrefixAndSuffix(t *testing.T) {
+	t.Parallel()
+	buf := &bytes.Buffer{}
+	logger := log.NewLogfmtLogger(buf)
```
ChrisHines, line 62: I would prefer this test didn't depend on one of the formatting loggers. `WithPrefix` and `WithSuffix` have specific semantics related to the order of values passed to other loggers, but don't have any control over what those loggers do with the values or if the order is honored.

Look at the other tests in this file to see how they capture the `keyvals` slice in a `[]interface{}` variable and then verify the order in that slice.

log/log_test.go
ChrisHines: Yes, these are the benchmarks I was looking for, thanks.

log/log.go
```diff
@@ -113,13 +149,15 @@ func (l *context) Log(keyvals ...interface{}) error {
...
-		// that future log events will reevaluate the stored Valuers.
+		// If no keyvals were appended above then we must copy l.keyvals
+		// and l.suffixKeyvals so that future log events will reevaluate
+		// the stored Valuers.
 		if len(keyvals) == 0 {
-			kvs = append([]interface{}{}, l.keyvals...)
+			kvs = append([]interface{}{}, append(l.keyvals, l.suffixKeyvals...)...)
 		}
-		bindValues(kvs[:len(l.keyvals)])
+		bindValues(kvs[:(len(kvs))])
```
ChrisHines, line 160: This call to bindValues is not quite what we want.

Before adding `WithSuffix` we only passed in a subslice of `kvs` that included the keyvals stored in `context.keyvals` and excluded the new keyvals passed to the `Log` method.

The new version here includes the new keyvals passed to the `Log` method. I think there should be two calls to `bindValues`, one for the elements of `kvs` from `l.keyvals` and one for the elements from `l.suffixKeyvals`.

We should also consider separate `hasValuer` fields for each group of keyvals stored in the context and handling them separately in the `Log` method.

log/log_test.go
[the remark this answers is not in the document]
vinayvinay: Agree Chris, I've corrected that in https://github.com/go-kit/kit/pull/992/commits/e2fae8b60e89b24171a06ca3a48b1b826c6a43d1.

log/log.go
vinayvinay: Ah, got a bit confused there. Good spot. Addressed in: https://github.com/go-kit/kit/pull/992/commits/b8ab9fc3b52a767219010f528bed08d306acc918 and https://github.com/go-kit/kit/pull/992/commits/6746fef7864e26bd7399a9630cd09e844c9acda6 by making 2 calls to `bindValues` and creating a separate `hasValuer` for suffix.

vinayvinay:
@ChrisHines Thanks for your guidance. I agree to the feedback and have made changes accordingly. Please take another look. I've kept commits separate to make it easy for you to review, will squash later.

vinayvinay:
There isn't a notable change to benchmarks after recent changes.

log/log.go
```diff
@@ -93,9 +131,12 @@ func WithPrefix(logger Logger, keyvals ...interface{}) Logger {
 //       returning a newly constructed context with a merged keyvals rather
 //       than simply wrapping the existing context.
 type context struct {
-	logger    Logger
-	keyvals   []interface{}
-	hasValuer bool
+	logger     Logger
+	keyvals    []interface{}
+	hasValuer  bool
```
ChrisHines, line 136: I think we should move `hasValuer` down a couple lines so that it is after `sKeyvals` and before `sHasValuer`. Putting all the `bool` fields consecutively at the end avoids padding bytes in the middle of the struct that the compiler will add to maintain the alignment requirements of the bigger slice fields.

Yes, it is a micro-optimization, and yes it is a bit less readable, but sometimes these things matter and it none of the fields are exported, so I am not too worried about the slight loss in readability in this case.

log/log_test.go
```diff
@@ -145,6 +194,43 @@ func TestWithConcurrent(t *testing.T) {
 	}
 }
 
+func TestLogCopiesValuers(t *testing.T) {
```
ChrisHines, line 197: Nice test. Thanks for adding this.

log/log.go
vinayvinay: I wasn't aware of this, Chris. Thanks for explaining, I found it fascinating! I've updated the PR.

ChrisHines:
> One minor thing I'd like your opinion on: comments mention `With and WithPrefix` ([example](https://github.com/vinayvinay/kit/blob/e3d5e919/log/log.go#L114)), now that there's a `WithSuffix`, would you say it is OK keeping these comments as-is, or update them to mention this 3rd func, or something like `With* functions` instead?

Just saw this question. That's a good catch. We prefer our code comments to be of the highest quality in Go kit, so yes, please update that to add `WithSuffix` to the list. I would rather not use `With* functions` because comments should preferably be in clear human language.

vinayvinay:
Thank you for your guidance Chris. It has been a pleasure!

Issue #997, opened 2020-07-18
SD:Consul - Index handling is incorrect per consul docs


schmidtw:
The [loop](https://github.com/go-kit/kit/blob/master/sd/consul/instancer.go#L70) where index handling is incorrect.

Consul Docs
https://www.consul.io/api-docs/features/blocking#implementation-details

### Index goes backwards

In rare cases, the index can be returned that is before the last index.  In this case, the system can basically miss out on events for either a very long time or forever.  This can be pretty damaging if you depend on the consistency of the SD value & the value does not appropriately update.

### Index returned is zero

In rare cases the returned index can be zero.  In this case the code should backoff and try again to avoid a tight loop.

schmidtw:
I have a proposed fix for these two errors in my local repo.
https://github.com/schmidtw/kit/commit/e539a98781cb06b55e5f787254627b720f21e8ff

peterbourgon:
Happy to take a PR.

sagikazarmark:
Fixed in #999

PR #999, opened 2020-07-20
Improve inconsistent Consul SD index handling.

schmidtw:
Consul has guidelines for how the index movement should be handled.

https://www.consul.io/api-docs/features/blocking#implementation-details

Generally the index should always increase but if it does not increase this
patch implements the suggested solutions for recovery.


schmidtw:
This should fix #997

peterbourgon:
Thanks!

PR #1008, opened 2020-08-26
Move the ClientAfter calls to before the decode

directionless:
ClientAfter is documented as running _prior_ to being decoded. But, this
was only partly true. They were run prior to decode, but after the
jsonrpc unmarshalling,

As the callback uses the bare response object, and at least for me, is
used to debug what's seen on the wire, this seems incorrect.

This oddity does not exist in other transports, as there is not the duplicate decode step.

### DIFF
--- a/transport/http/jsonrpc/client.go
+++ b/transport/http/jsonrpc/client.go
@@ -196,15 +196,15 @@ func (c Client) Endpoint() endpoint.Endpoint {
 			defer resp.Body.Close()
 		}
 
+		for _, f := range c.after {
+			ctx = f(ctx, resp)
+		}
+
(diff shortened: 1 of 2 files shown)


peterbourgon:
Sounds good. Can you add a test which fails before this change and succeeds afterwards?

directionless:
Okay! I added tests.

Also did some cleanup on that test file:
* Removed the external network calls
*  added various `Close()` and `Parallel()` calls
* moved some of the wrappers into a table driven test

directionless:
Hi! I wanted to check in on this?

peterbourgon:
Apologies for the delay. Looks good. Thanks!

PR #1016, opened 2020-09-29
directly calling the methods of the service (booking) interface

Qalifah:
Since the Service interface was directly referenced, the service middleware can call its method directly


peterbourgon:
These changes would create infinite loops.

PR #1017, opened 2020-09-30
Fix: if register failed, return error for handling

QiaoHao9:
When I use go-kit's etcd, I find that I don't know if the `func Register()` was successful, so I need it  return an error as info.

### DIFF
--- a/sd/etcd/registrar.go
+++ b/sd/etcd/registrar.go
@@ -69,15 +69,18 @@ func NewRegistrar(client Client, service Service, logger log.Logger) *Registrar 
 
 // Register implements the sd.Registrar interface. Call it when you want your
 // service to be registered in etcd, typically at startup.
-func (r *Registrar) Register() {
+func (r *Registrar) Register() error {
 	if err := r.client.Register(r.service); err != nil {
 		r.logger.Log("err", err)
+		return err
 	} else {
 		r.logger.Log("action", "register")
 	}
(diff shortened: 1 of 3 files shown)


tinyRatP:
I'm not good at test, who can help me fix tests.

sd/etcd/registrar.go
```diff
@@ -69,15 +69,18 @@ func NewRegistrar(client Client, service Service, logger log.Logger) *Registrar
 
 // Register implements the sd.Registrar interface. Call it when you want your
 // service to be registered in etcd, typically at startup.
-func (r *Registrar) Register() {
+func (r *Registrar) Register() error {
```
peterbourgon, line 72: I don't think it's clear how to use this type/method. What should happen if the first Register fails? Should it still `go r.loop()` or not? If yes, does it really matter if the first register fails? If not, should the loop happen in a different way?

sd/etcd/registrar.go
```diff
@@ -104,9 +107,10 @@ func (r *Registrar) loop() {
 
 // Deregister implements the sd.Registrar interface. Call it when you want your
 // service to be deregistered from etcd, typically just prior to shutdown.
-func (r *Registrar) Deregister() {
+func (r *Registrar) Deregister() error {
```
peterbourgon, line 110: Same comments as with Register.

Issue #858, opened 2019-04-09
Feature request: AWS SQS transport


dhait:
This is different from pubsub #295 because it is specific to message handling from AWS SQS, not a generic pubsub solution.  I am envisioning something modeled very similarly to the AMQP transport. 
I was going to try to implement this; would there be interest from the team if we proposed it as a PR?

peterbourgon:
Would be happy to review.

0marq:
Hello,

I took the liberty of implementing the AWS SQS transport since this issue has not been update for almost a year and a half. I just submitted the PR to review (https://github.com/go-kit/kit/pull/1018), and linked it to this issue! 

Looking forward to hearing from you !

PR #1018, opened 2020-10-07
Add AWS' Simple Queue Service support for transport

0marq:
Hi,

@forepaas, we are currently starting a new project using go-kit and wanted to use AWS' Simple Query Service as a transport layer between some of our microservices.

I took the liberty of forking the project and implementing the support of SQS in go-kit's transport. This was discussed on issue https://github.com/go-kit/kit/issues/858 but was never implemented.

If you have any remarks, suggestions or questions please let me know.

### DIFF
--- a/transport/awssqs/consumer.go
+++ b/transport/awssqs/consumer.go
@@ -129,57 +129,59 @@ func ConsumerDeleteMessageAfter() ConsumerOption {
 }
 
 // ServeMessage serves an SQS message.
-func (c Consumer) ServeMessage(ctx context.Context, msg *sqs.Message) error {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+func (c Consumer) ServeMessage(ctx context.Context) func(msg *sqs.Message) error {
+	return func(msg *sqs.Message) error {
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
 
-	if len(c.finalizer) > 0 {
-		defer func() {
-			for _, f := range c.finalizer {
-				f(ctx, msg)
-			}
-		}()
-	}
+		if len(c.finalizer) > 0 {
+			defer func() {
+				for _, f := range c.finalizer {
+					f(ctx, msg)
+				}
+			}()
+		}
 
-	for _, f := range c.before {
-		ctx = f(ctx, cancel, msg)
-	}
+		for _, f := range c.before {
+			ctx = f(ctx, cancel, msg)
+		}
 
-	req, err := c.dec(ctx, msg)
-	if err != nil {
-		c.errorHandler.Handle(ctx, err)
-		c.errorEncoder(ctx, err, msg, c.sqsClient)
-		return err
-	}
+		req, err := c.dec(ctx, msg)
+		if err != nil {
+			c.errorHandler.Handle(ctx, err)
+			c.errorEncoder(ctx, err, msg, c.sqsClient)
+			return err
+		}
 
-	response, err := c.e(ctx, req)
-	if err != nil {
-		c.errorHandler.Handle(ctx, err)
-		c.errorEncoder(ctx, err, msg, c.sqsClient)
-		return err
-	}
+		response, err := c.e(ctx, req)
+		if err != nil {
+			c.errorHandler.Handle(ctx, err)
+			c.errorEncoder(ctx, err, msg, c.sqsClient)
+			return err
+		}
 
-	responseMsg := sqs.SendMessageInput{}
-	for _, f := range c.after {
-		ctx = f(ctx, cancel, msg, &responseMsg)
-	}
+		responseMsg := sqs.SendMessageInput{}
+		for _, f := range c.after {
+			ctx = f(ctx, cancel, msg, &responseMsg)
+		}
 
-	if !c.wantRep(ctx, msg) {
+		if !c.wantRep(ctx, msg) {
+			return nil
+		}
+
+		if err := c.enc(ctx, &responseMsg, response); err != nil {
+			c.errorHandler.Handle(ctx, err)
+			c.errorEncoder(ctx, err, msg, c.sqsClient)
+			return err
+		}
+
+		if _, err := c.sqsClient.SendMessageWithContext(ctx, &responseMsg); err != nil {
+			c.errorHandler.Handle(ctx, err)
+			c.errorEncoder(ctx, err, msg, c.sqsClient)
+			return err
+		}
 		return nil
 	}
-
-	if err := c.enc(ctx, &responseMsg, response); err != nil {
-		c.errorHandler.Handle(ctx, err)
-		c.errorEncoder(ctx, err, msg, c.sqsClient)
-		return err
-	}
-
-	if _, err := c.sqsClient.SendMessageWithContext(ctx, &responseMsg); err != nil {
-		c.errorHandler.Handle(ctx, err)
-		c.errorEncoder(ctx, err, msg, c.sqsClient)
-		return err
-	}
-	return nil
 }
 
 // ErrorEncoder is responsible for encoding an error to the consumer's reply.
--- a/transport/awssqs/consumer.go
+++ b/transport/awssqs/consumer.go
@@ -12,15 +12,14 @@ import (
 	"github.com/go-kit/kit/transport"
 )
 
-// Consumer wraps an endpoint and provides and provides a handler for sqs msgs
+// Consumer wraps an endpoint and provides a handler for sqs messages.
 type Consumer struct {
-	sqsClient             SQSClient
+	sqsClient             Client
 	e                     endpoint.Endpoint
 	dec                   DecodeRequestFunc
 	enc                   EncodeResponseFunc
 	wantRep               WantReplyFunc
-	queueURL              *string
-	dlQueueURL            *string
+	queueURL              string
 	visibilityTimeout     int64
 	visibilityTimeoutFunc VisibilityTimeoutFunc
 	before                []ConsumerRequestFunc
@@ -33,14 +32,11 @@ type Consumer struct {
 // NewConsumer constructs a new Consumer, which provides a Consume method
 // and message handlers that wrap the provided endpoint.
 func NewConsumer(
-	sqsClient SQSClient,
+	sqsClient Client,
 	e endpoint.Endpoint,
 	dec DecodeRequestFunc,
 	enc EncodeResponseFunc,
-	wantRep WantReplyFunc,
-	queueURL *string,
-	dlQueueURL *string,
-	visibilityTimeout int64,
+	queueURL string,
 	options ...ConsumerOption,
 ) *Consumer {
 	s := &Consumer{
@@ -48,10 +44,9 @@ func NewConsumer(
 		e:                     e,
 		dec:                   dec,
 		enc:                   enc,
-		wantRep:               wantRep,
+		wantRep:               DoNotRespond,
 		queueURL:              queueURL,
-		dlQueueURL:            dlQueueURL,
-		visibilityTimeout:     visibilityTimeout,
+		visibilityTimeout:     int64(30),
 		visibilityTimeoutFunc: DoNotExtendVisibilityTimeout,
 		errorEncoder:          DefaultErrorEncoder,
 		errorHandler:          transport.NewLogErrorHandler(log.NewNopLogger()),
@@ -86,20 +81,23 @@ func ConsumerErrorEncoder(ee ErrorEncoder) ConsumerOption {
 }
 
 // ConsumerVisbilityTimeOutFunc is used to extend the visibility timeout
-// for messages during when processing them. Clients can
-// use this to provide custom visibility timeout extension. By default,
-// visibility timeout are not extend.
+// for messages while the consumer processes them.
+// VisibilityTimeoutFunc will need to check that the provided context is not done.
+// By default, visibility timeout are not extended.
 func ConsumerVisbilityTimeOutFunc(vtFunc VisibilityTimeoutFunc) ConsumerOption {
 	return func(c *Consumer) { c.visibilityTimeoutFunc = vtFunc }
 }
 
-// ConsumerErrorLogger is used to log non-terminal errors. By default, no errors
-// are logged. This is intended as a diagnostic measure. Finer-grained control
-// of error handling, including logging in more detail, should be performed in a
-// custom ConsumerErrorEncoder which has access to the context.
-// Deprecated: Use ConsumerErrorHandler instead.
-func ConsumerErrorLogger(logger log.Logger) ConsumerOption {
-	return func(c *Consumer) { c.errorHandler = transport.NewLogErrorHandler(logger) }
+// ConsumerVisibilityTimeout overrides the default value for the consumer's
+// visibilityTimeout field.
+func ConsumerVisibilityTimeout(visibilityTimeout int64) ConsumerOption {
+	return func(c *Consumer) { c.visibilityTimeout = visibilityTimeout }
+}
+
+// ConsumerWantReplyFunc overrides the default value for the consumer's
+// wantRep field.
+func ConsumerWantReplyFunc(replyFunc WantReplyFunc) ConsumerOption {
+	return func(c *Consumer) { c.wantRep = replyFunc }
 }
 
 // ConsumerErrorHandler is used to handle non-terminal errors. By default, non-terminal errors
@@ -110,16 +108,18 @@ func ConsumerErrorHandler(errorHandler transport.ErrorHandler) ConsumerOption {
 	return func(c *Consumer) { c.errorHandler = errorHandler }
 }
 
-// ConsumerFinalizer is executed at the end of every request from a publisher through SQS.
+// ConsumerFinalizer is executed once all the received SQS messages are done being processed.
 // By default, no finalizer is registered.
 func ConsumerFinalizer(f ...ConsumerFinalizerFunc) ConsumerOption {
 	return func(c *Consumer) { c.finalizer = f }
 }
 
-// Consume calls ReceiveMessageWithContext and handles messages
-// having receiveMsgInput as param allows each user to have his own receive config
+// Consume calls ReceiveMessageWithContext and handles messages having an
+// sqs.ReceiveMessageInput as parameter allows each user to have his own receive configuration.
+// That said, this method overrides the queueURL for the provided ReceiveMessageInput to ensure
+// the messages are retrieved from the consumer's configured queue.
 func (c Consumer) Consume(ctx context.Context, receiveMsgInput *sqs.ReceiveMessageInput) error {
-	receiveMsgInput.QueueUrl = c.queueURL
+	receiveMsgInput.QueueUrl = &c.queueURL
 	out, err := c.sqsClient.ReceiveMessageWithContext(ctx, receiveMsgInput)
 	if err != nil {
 		return err
@@ -127,18 +127,20 @@ func (c Consumer) Consume(ctx context.Context, receiveMsgInput *sqs.ReceiveMessa
 	return c.HandleMessages(ctx, out.Messages)
 }
 
-// HandleMessages handles the consumed messages
+// HandleMessages handles the consumed messages.
 func (c Consumer) HandleMessages(ctx context.Context, msgs []*sqs.Message) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// Copy msgs slice in leftMsgs
+	// Copy received messages slice in leftMsgs slice
+	// leftMsgs will be used by the consumer's visibilityTimeoutFunc to extend the
+	// visibility timeout for the messages that have not been processed yet.
 	leftMsgs := []*sqs.Message{}
 	leftMsgs = append(leftMsgs, msgs...)
 
-	// this func allows us to extend visibility timeout to give use
-	// time to process the messages in leftMsgs
-	go c.visibilityTimeoutFunc(ctx, c.sqsClient, c.queueURL, c.visibilityTimeout, &leftMsgs)
+	visibilityTimeoutCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go c.visibilityTimeoutFunc(visibilityTimeoutCtx, c.sqsClient, c.queueURL, c.visibilityTimeout, &leftMsgs)
 
 	if len(c.finalizer) > 0 {
 		defer func() {
@@ -160,7 +162,7 @@ func (c Consumer) HandleMessages(ctx context.Context, msgs []*sqs.Message) error
 	return nil
 }
 
-// HandleSingleMessage handles a single sqs message
+// HandleSingleMessage handles a single sqs message.
 func (c Consumer) HandleSingleMessage(ctx context.Context, msg *sqs.Message, leftMsgs *[]*sqs.Message) error {
 	req, err := c.dec(ctx, msg)
 	if err != nil {
@@ -182,7 +184,6 @@ func (c Consumer) HandleSingleMessage(ctx context.Context, msg *sqs.Message, lef
 	}
 
 	if !c.wantRep(ctx, msg) {
-		// Message does not expect answer
 		return nil
 	}
 
@@ -200,30 +201,45 @@ func (c Consumer) HandleSingleMessage(ctx context.Context, msg *sqs.Message, lef
 	return nil
 }
 
-// ErrorEncoder is responsible for encoding an error to the consumer reply.
+// ErrorEncoder is responsible for encoding an error to the consumer's reply.
 // Users are encouraged to use custom ErrorEncoders to encode errors to
 // their replies, and will likely want to pass and check for their own error
 // types.
-type ErrorEncoder func(ctx context.Context, err error, req *sqs.Message, sqsClient SQSClient)
+type ErrorEncoder func(ctx context.Context, err error, req *sqs.Message, sqsClient Client)
 
 // ConsumerFinalizerFunc can be used to perform work at the end of a request
 // from a publisher, after the response has been written to the publisher. The
 // principal intended use is for request logging.
-// Can also be used to delete messages once fully proccessed
+// Can also be used to delete messages once fully proccessed.
 type ConsumerFinalizerFunc func(ctx context.Context, msg *[]*sqs.Message)
 
-// DefaultErrorEncoder simply ignores the message. It does not reply
-// nor Ack/Nack the message.
-func DefaultErrorEncoder(context.Context, error, *sqs.Message, SQSClient) {
+// VisibilityTimeoutFunc encapsulates logic to extend messages visibility timeout.
+// this can be used to provide custom visibility timeout extension such as doubling it everytime
+// it gets close to being reached.
+// VisibilityTimeoutFunc will need to check that the provided context is not done and return once it is.
+type VisibilityTimeoutFunc func(context.Context, Client, string, int64, *[]*sqs.Message) error
+
+// WantReplyFunc encapsulates logic to check whether message awaits response or not
+// for example check for a given message attribute value.
+type WantReplyFunc func(context.Context, *sqs.Message) bool
+
+// DefaultErrorEncoder simply ignores the message. It does not reply.
+func DefaultErrorEncoder(context.Context, error, *sqs.Message, Client) {
 }
 
-// DoNotExtendVisibilityTimeout is the default value for visibilityTimeoutFunc
+// DoNotExtendVisibilityTimeout is the default value for the consumer's visibilityTimeoutFunc.
 // It returns no error and does nothing
-func DoNotExtendVisibilityTimeout(context.Context, SQSClient, *string, int64, *[]*sqs.Message) error {
+func DoNotExtendVisibilityTimeout(context.Context, Client, string, int64, *[]*sqs.Message) error {
 	return nil
 }
 
-// EncodeJSONResponse marshals response as json and loads it into input MessageBody
+// DoNotRespond is a WantReplyFunc and is the default value for consumer's wantRep field.
+// It indicates that the message do not expect a response.
+func DoNotRespond(context.Context, *sqs.Message) bool {
+	return false
+}
+
+// EncodeJSONResponse marshals response as json and loads it into an sqs.SendMessageInput MessageBody.
 func EncodeJSONResponse(_ context.Context, input *sqs.SendMessageInput, response interface{}) error {
 	payload, err := json.Marshal(response)
 	if err != nil {
@@ -233,9 +249,9 @@ func EncodeJSONResponse(_ context.Context, input *sqs.SendMessageInput, response
 	return nil
 }
 
-// SQSClient is an interface to make testing possible.
-// It is highly recommended to use *sqs.SQS as the interface implementation.
-type SQSClient interface {
+// Client is consumer contract for the Producer and Consumer.
+// It models methods of the AWS *sqs.SQS type.
+type Client interface {
 	SendMessageWithContext(ctx context.Context, input *sqs.SendMessageInput, opts ...request.Option) (*sqs.SendMessageOutput, error)
 	ReceiveMessageWithContext(ctx context.Context, input *sqs.ReceiveMessageInput, opts ...request.Option) (*sqs.ReceiveMessageOutput, error)
 	ChangeMessageVisibilityWithContext(ctx aws.Context, input *sqs.ChangeMessageVisibilityInput, opts ...request.Option) (*sqs.ChangeMessageVisibilityOutput, error)
--- a/transport/awssqs/consumer.go
+++ b/transport/awssqs/consumer.go
@@ -13,36 +13,19 @@ import (
 	"github.com/go-kit/kit/transport"
 )
 
-// Delete is a type to indicate when the consumed message should be deleted.
-type Delete int
-
-const (
-	// BeforeHandle deletes the message before starting to handle it.
-	BeforeHandle Delete = iota
-	// AfterHandle deletes the message once it has been fully processed.
-	// This is the consumer's default value.
-	AfterHandle
-	// Never does not delete the message.
-	Never
-)
-
 // Consumer wraps an endpoint and provides a handler for SQS messages.
 type Consumer struct {
-	sqsClient             sqsiface.SQSAPI
-	e                     endpoint.Endpoint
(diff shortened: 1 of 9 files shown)


transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/transport"
+)
+
+// Consumer wraps an endpoint and provides and provides a handler for sqs msgs
```
peterbourgon, line 15: Please format all doc comments as full English sentences with punctuation and proper initialism capitalization, e.g.

```suggestion
// Consumer wraps an endpoint and provides a handler for SQS messages.
```

This comment applies to all doc comments in the PR.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return err
+	}
+	input.MessageBody = aws.String(string(payload))
+	return nil
+}
+
+// SQSClient is an interface to make testing possible.
+// It is highly recommended to use *sqs.SQS as the interface implementation.
+type SQSClient interface {
```
peterbourgon, line 238: This name stutters a bit: awssqs.SQSClient. Consider `Client`.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
+package awssqs
```
peterbourgon, line 1: Is there any other `sqs` except `awssqs`? Could this be `sqs`? Or maybe that conflicts with the package name of the SDK?

transport/awssqs/publisher.go
```diff
@@ -0,0 +1,128 @@
...
+	dec              DecodeResponseFunc
+	before           []PublisherRequestFunc
+	after            []PublisherResponseFunc
+	timeout          time.Duration
+}
+
+// NewPublisher constructs a usable Publisher for a single remote method.
+func NewPublisher(
+	sqsClient SQSClient,
+	queueURL *string,
```
peterbourgon, line 29: Why is this a pointer to a string?

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+// and message handlers that wrap the provided endpoint.
+func NewConsumer(
+	sqsClient SQSClient,
+	e endpoint.Endpoint,
+	dec DecodeRequestFunc,
+	enc EncodeResponseFunc,
+	wantRep WantReplyFunc,
+	queueURL *string,
+	dlQueueURL *string,
+	visibilityTimeout int64,
```
peterbourgon, line 43: This is a log of required parameters. Can any of these have sensible default values?

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+	payload, err := json.Marshal(response)
+	if err != nil {
+		return err
+	}
+	input.MessageBody = aws.String(string(payload))
+	return nil
+}
+
+// SQSClient is an interface to make testing possible.
+// It is highly recommended to use *sqs.SQS as the interface implementation.
```
peterbourgon, line 237:
```suggestion
// SQSClient is consumer contract for the Producer and Consumer.
// It models methods of the AWS *sqs.SQS type.
```

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Copy msgs slice in leftMsgs
+	leftMsgs := []*sqs.Message{}
+	leftMsgs = append(leftMsgs, msgs...)
+
+	// this func allows us to extend visibility timeout to give use
+	// time to process the messages in leftMsgs
+	go c.visibilityTimeoutFunc(ctx, c.sqsClient, c.queueURL, c.visibilityTimeout, &leftMsgs)
```
peterbourgon, line 141: Please provide a way to ensure this goroutine is terminated.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// HandleMessages handles the consumed messages
+func (c Consumer) HandleMessages(ctx context.Context, msgs []*sqs.Message) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Copy msgs slice in leftMsgs
+	leftMsgs := []*sqs.Message{}
+	leftMsgs = append(leftMsgs, msgs...)
```
peterbourgon, line 137: This variable is passed to 2 goroutines, but there's no synchronization. How do you prevent data races?

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,242 @@
...
+	return c.HandleMessages(ctx, out.Messages)
+}
+
+// HandleMessages handles the consumed messages
+func (c Consumer) HandleMessages(ctx context.Context, msgs []*sqs.Message) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Copy msgs slice in leftMsgs
+	leftMsgs := []*sqs.Message{}
```
peterbourgon, line 136: I actually don't understand what `leftMsgs` is for. Can you explain?

peterbourgon (commented):
First pass

transport/awssqs/consumer.go
0marq: Indeed this was done to avoid conflicts with the SDK, and to follow the naming for awslambda. I can change it to sqs if you want me to, conflicts with the SDK can be solved by adding an alias when needed.

transport/awssqs/consumer.go
0marq: I agree

transport/awssqs/publisher.go
0marq: This is because the aws sdk uses string pointers. I'll change them to strings and call a pointer to the var when needed

transport/awssqs/consumer.go
0marq: When a message is retrieved from SQS, it is made invisible for a certain period, to avoid that two consumers handle the same message concurrently.
When consuming multiple messages at the same time, or for long processing messages, the SDK allows you to extend this visibility timeout.
Once a consumer is done processing a message it should delete the message from the queue using the ReceiptHandle that can be found in the message. If the visibilityTimeout expires so does the ReceiptHandle, this will lead up to message being visible again in the queue and reprocessed.

leftMsgs is used by the consumer's visibilityTimeoutFunc to extend the visibility timeout for the messages that have not been processed yet. 
After a message has been processed and deleted from the queue, leftMsgs should be updated. Those two actions can be done in the consumer's after functions.

transport/awssqs/consumer.go
0marq: What about adding a sync.Mutex in the consumer and pass it to the functions that manipulate leftMsgs ?

0marq:
I have done all the asked changes, I'll let you run a second pass, and reply to my answers.

transport/awssqs/publisher.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/go-kit/kit/endpoint"
+)
+
+// Publisher wraps an sqs client and queue, and provides a method that
+// implements endpoint.Endpoint.
+type Publisher struct {
+	sqsClient        Client
+	queueURL         string
+	responseQueueURL string
```
vrazdalovschi, line 18: What the sense of this parameter? Is it not a specific use case when you want to provide your "responseQueueURL" attribute? Someone can want to use "responseUrl" and then would be a problem. 
If you want to use your custom messageAttributes, then you should use it in "before" or in encodeMessage, but not in the general approach.

0marq: The Publisher is the one that sets the responseQueue for his messages.
As the SQS doc [response reply pattern implementation](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/working-with-messages.html#implementing-request-response-systems) states :

> Don't create reply queues per message. Instead, create reply queues on startup, per producer, and use a correlation ID message attribute to map replies to requests.

So this field actually makes sense in the Producer.

The Consumer on the other hand will indeed retrieve this value from a message attribute that will be added in the message in a before function.

vrazdalovschi:
@0marq SQS Consumer. Each sqs message has a default flow: read / process / delete. I see that in your implementation the deleting remain on the user part in different ways and it is not obvious how to do better (ConsumerFinalizerFunc or ConsumerResponseFunc). If a message is not deleted, then it will be processed X time, depends on sqs queue configuration. 

Here should be a more obvious way: Add a ConsumerOption with DeleteMessage call which can be used beforeDecode or afterEncode, I can imagine when someone wants to remove and take all the responsibility on the processing message or someone want to delete after successful processing.

transport/awssqs/publisher.go
vrazdalovschi: Ok, I understand your desire for this pattern, but why should we use this "responseQueueURL" as the field of Publisher? I  don't see any uses of this field, so then if you want to add this attribute, you will add in your EncodeRequestFunc or in PublisherBefore option.

0marq: Indeed the field was not used, I just added it as paramter to the Publisher's before function

0marq:
@vrazdalovschi thank you for you remarks, I think I have answered them and taken the appropriate actions. Please tell me what you think of them.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		return err
+	}
+	input.MessageBody = aws.String(string(payload))
+	return nil
+}
+
+// Client is consumer contract for the Producer and Consumer.
+// It models methods of the AWS *sqs.SQS type.
+type Client interface {
```
xyluet, line 300: https://github.com/aws/aws-sdk-go/blob/master/service/sqs/sqsiface/interface.go

SQS already has their API interface

0marq: Thank you for your input. I just made the switch.

xyluet:
SQS (Simple Queue Service), not query.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+	BeforeHandle Delete = iota
+	// AfterHandle deletes the message once it has been fully processed.
+	// This is the consumer's default value.
+	AfterHandle
+	// Never does not delete the message.
+	Never
+)
+
+// Consumer wraps an endpoint and provides a handler for sqs messages.
+type Consumer struct {
```
xyluet, line 30: All pubsub implementations in this repository have their name `Subscriber`. Also on AWS doc said it's a pub/sub messaging, not pub/consumer.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+				c.errorEncoder(ctx, err, msg, c.sqsClient)
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// HandleSingleMessage handles a single sqs message.
+func (c Consumer) HandleSingleMessage(ctx context.Context, msg *sqs.Message, leftMsgs *[]*sqs.Message) error {
```
xyluet, line 212: This function should be private, to remove misuses when developer using this service. Also `HandleMessages`

transport/awssqs/consumer.go
0marq: I found the term consumer more adapted given the fact that you need to send a receive request to retrieve messages from the queue.
Do you have quote and a link to prove your point ?

In their [pub/sub documentation](https://aws.amazon.com/pub-sub-messaging/) they actually oppose pub/sub systems to messaging queues, while in their [messaging queue documentation](https://aws.amazon.com/message-queue/) they use the terms producer/consumer. This would actually mean that Publisher should be replaced by Producer.

If this is ok for you @xyluet, I'll change Publisher by Producer.

xyluet: I'm actually not a maintainer. This will be the first one that using terms producer/consumer in this repo. What do you think @peterbourgon ?

peterbourgon: Let's use whatever the SDK prefers. No preference on Go kit side.

xyluet: Producer/consumer seems ok.

0marq: Done

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+				c.errorHandler.Handle(ctx, err)
+				c.errorEncoder(ctx, err, msg, c.sqsClient)
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// handleSingleMessage handles a single sqs message.
```
xyluet, line 211: sqs -> SQS for all comments since it's a product.

transport/awssqs/request_response_func.go
```diff
@@ -0,0 +1,34 @@
...
+// put it into a request context. In Consumers, RequestFuncs are executed prior
+// to invoking the endpoint.
+// use cases eg. in Consumer : extract message into context, or filter received messages.
+type ConsumerRequestFunc func(context.Context, *[]*sqs.Message) context.Context
+
+// ProducerRequestFunc may take information from a producer request and put it into a
+// request context, or add some informations to SendMessageInput. In Producers,
+// RequestFuncs are executed prior to publishing the message but after encoding.
+// use cases eg. in Producer : enforce some message attributes to SendMessageInput.
+type ProducerRequestFunc func(context.Context, *sqs.SendMessageInput, string) context.Context
```
vrazdalovschi, line 21: string parameter. It is not obvious what is it.

transport/awssqs/request_response_func.go
```diff
@@ -0,0 +1,34 @@
...
+
+// ConsumerResponseFunc may take information from a request context and use it to
+// manipulate a Producer. ConsumerResponseFunc are only executed in
+// consumers, after invoking the endpoint but prior to publishing a reply.
+// use cases eg. : Pipe information from request message to response MessageInput,
+// delete msg from queue or update leftMsgs slice.
+type ConsumerResponseFunc func(context.Context, *sqs.Message, *sqs.SendMessageInput, *[]*sqs.Message, *sync.Mutex) context.Context
```
vrazdalovschi, line 28: Can you add an example of using Mutex? What is the reason to use it?

transport/awssqs/producer.go
```diff
@@ -0,0 +1,127 @@
...
+	before           []ProducerRequestFunc
+	after            []ProducerResponseFunc
+	timeout          time.Duration
+}
+
+// NewProducer constructs a usable Producer for a single remote method.
+func NewProducer(
+	sqsClient sqsiface.SQSAPI,
+	queueURL string,
+	responseQueueURL string,
```
vrazdalovschi, line 31: So, the same question remains. Why do you want to store responseQueueURL in the Producer structure? If someone wants to pass this into message attributes, he will do it in encode or in the beforeProducer. This `responseQueueURL` is not used anywhere.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+// from a producer, after the response has been written to the producer. The
+// principal intended use is for request logging.
+// Can also be used to delete messages once fully proccessed.
+type ConsumerFinalizerFunc func(ctx context.Context, msg *[]*sqs.Message)
+
+// VisibilityTimeoutFunc encapsulates logic to extend messages visibility timeout.
+// this can be used to provide custom visibility timeout extension such as doubling it everytime
+// it gets close to being reached.
+// VisibilityTimeoutFunc will need to check that the provided context is not done and return once it is.
+type VisibilityTimeoutFunc func(context.Context, sqsiface.SQSAPI, string, int64, *[]*sqs.Message, *sync.Mutex) error
```
vrazdalovschi, line 266: string, int64. It is not obvious what to do with them

vrazdalovschi (changes requested):

transport/awssqs/request_response_func.go
peterbourgon: Should absolutely not expose a mutex directly like this.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+		if c.deleteMessage == AfterHandle {
+			if _, err := c.sqsClient.DeleteMessageWithContext(ctx, &sqs.DeleteMessageInput{
+				QueueUrl:      &c.queueURL,
+				ReceiptHandle: msg.ReceiptHandle,
+			}); err != nil {
+				c.errorHandler.Handle(ctx, err)
+				c.errorEncoder(ctx, err, msg, c.sqsClient)
+				return err
+			}
+		}
```
xyluet, line 206:
```suggestion
	for _, f := range c.before {
		ctx = f(ctx, &msgs, c.sqsClient)
	}

	for _, msg := range msgs {
		if c.deleteMessage == BeforeHandle {
			if _, err := c.sqsClient.DeleteMessageWithContext(ctx, &sqs.DeleteMessageInput{
				QueueUrl:      &c.queueURL,
				ReceiptHandle: msg.ReceiptHandle,
			}); err != nil {
				c.errorHandler.Handle(ctx, err)
				c.errorEncoder(ctx, err, msg, c.sqsClient)
				return err
			}
		}

		if err := c.handleSingleMessage(ctx, msg, &leftMsgs); err != nil {
			return err
		}

		if c.deleteMessage == AfterHandle {
			if _, err := c.sqsClient.DeleteMessageWithContext(ctx, &sqs.DeleteMessageInput{
				QueueUrl:      &c.queueURL,
				ReceiptHandle: msg.ReceiptHandle,
			}); err != nil {
				c.errorHandler.Handle(ctx, err)
				c.errorEncoder(ctx, err, msg, c.sqsClient)
				return err
			}
		}
```

It can be done on before or after func, right?

I prefer this as plugin if the developer want to delete the message before or after the message is handled.
For sure before function need to append signature which the SQS API.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+		return nil
+	}
+
+	if err := c.enc(ctx, &responseMsg, response); err != nil {
+		c.errorHandler.Handle(ctx, err)
+		c.errorEncoder(ctx, err, msg, c.sqsClient)
+		return err
+	}
+
+	if _, err := c.sqsClient.SendMessageWithContext(ctx, &responseMsg); err != nil {
```
xyluet, line 242: I will introduce `ResponsePublisher` which it will responsible to publish the response or just run no operation func. How about that?

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,296 @@
...
+			}
+		}()
+	}
+
+	for _, f := range c.before {
+		ctx = f(ctx, &msgs)
+	}
+
+	for _, msg := range msgs {
+		if c.deleteMessage == BeforeHandle {
```
xyluet, line 182: this process delete or after handle message can be done inside before or after function, right?

transport/awssqs/producer.go
xyluet: @0marq I understand. You have current project which need the queue url inside the process.

It can be done via plugin before or after function, and context value.

For example:
```go
type contextKey int

const (
	ContextKeyQueueURL contextKey = iota
)
// set this as before or after function
func SetProducerQueueUrl(url string) ProducureRequestFunc {
  return func(ctx context.Context, anotherParams) context.Context {
    return context.WithValue(ctx, ContextKeyQueueURL, url)
  }
}
```

then you can get the value via `ctx.Value(contextKey).(<yourtype>)`

transport/awssqs/consumer.go
xyluet, line 242: I will introduce ResponsePublisher which it will responsible to publish the response or just run no operation func. How about that?

transport/awssqs/consumer.go
xyluet, line 182: this process delete or after handle message can be done inside before or after function, right?

xyluet:
I've actually made my own implementation and using it on my production. The signature like below:

```go
//main.go

//build subscriber 
subscriber := sqstransport.NewSubscriber(endpoint, dec, enc, opts...)

for {
  out, _ := sqsClient.ReceiveMessage(ctx, input) //handle error
  for _, msg := range out.Messages {
    // Its caller responsibility if want to have many workers because its more easier to scale.
    // Now you are free to extend the visibility timeout, or using hearbeat suggested by AWS.
    subscriber.ServeMessage(&msg)
  }
}
```

What do you think?

transport/awssqs/consumer.go
vrazdalovschi: No, because in before/after func you will do for all messages (for example you have read with 10 messages), so it wouldn't be right. Maybe there would be more "readable" to extract delete to another func?

xyluet: Hmm, i see. Before is before you loop the messages, while after is for each message, weird. What do you think for my comment below? I borrow it from amqp transport.

transport/awssqs/request_response_func.go
0marq: I added this mutex to prevent data races when editing the leftMessages slice. 
This way if your ConsumerResponseFunc needs to edit the leftMessages you should lock it then unlock it once you're done. If your function does not edit the slice, it should leave it unlocked.
Would you rather I lock this mutex before calling the consumer's after functions no matter their content ?

transport/awssqs/producer.go
0marq: The value is actually used when calling before and after

transport/awssqs/consumer.go
0marq: I don't see what that would improve over my implementation. If you think it really does please help me understand.
I have a WantReplyFunc that can be either based on a message attribute or simply return false or true all the time.
Depending on the output of this func, I proceed to the rest of the function which is in charge of encoding and sending a response or not.

I prefere my approach because I find it closer to gokit's approach (in NATS transport for example) where you have a dedicated encoding response function.

transport/awssqs/consumer.go
0marq: I first thought that it would be up to the developer to delete the message inside an after func but @vrazdalovschi thought in https://github.com/go-kit/kit/pull/1018#issuecomment-706682608 that we should be responsible of deleting the message, not the developer given that this is part of the flow in handling an SQS message.

0marq:
> I actually made my own implementation and using it on my production. The signature like below:
> 
> ```go
> //main.go
> 
> //build subscriber 
> subscriber := sqstransport.NewSubscriber(endpoint, dec, enc, opts...)
> 
> for {
>   out, _ := sqsClient.ReceiveMessage(ctx, input) //handle error
>   for _, msg := range out.Messages {
>     // Its caller responsibility if want to have many workers because its more easier to scale.
>     // Now you are free to extend the visibility timeout, or using hearbeat suggested by AWS.
>     subscriber.ServeMessage(&msg)
>   }
> }
> ```
> 
> What do you think?

This would lighten the consumer's responsibilities because we would no longer need to handler visibility timeouts, while allowing multiple message processing.

In the beginning I was thinking to go towards this approach but I ended up building a code more complete but also more complex, my motivation was that the consumer should not only process the messages but also fetch the messages.
I don't know what's go-kit's opinion on this @peterbourgon ?

transport/awssqs/consumer.go
vrazdalovschi: @xyluet So, yes, it can be done in before/after with some default flow (Detele in After), and to provide these Options as helpers. As it's done for `DoNotExtendVisibilityTimeout`
@0marq I meant to provide the more clear way how to delete the messages because it wasn't documented and no examples. I sure, that everyone needs a way to remove the messages before/after their processing. So, to add from the box these before/after consumer functions would be fine.

0marq: Ok I get your point now. But I have an issue. After and before functions do not allow errors to be returned. 
Lets admit I have the following code

```go
// ConsumerDeleteMessagesBefore returns a ConsumerOption that appends a function
// that delete all received messages from queue to the list of consumer's before functions.
func ConsumerDeleteMessagesBefore() ConsumerOption {
	return func(c *Consumer) {
		deleteBefore := func(ctx context.Context, msgs *[]*sqs.Message) context.Context {
			for _, msg := range *msgs {
				if err := deleteMessage(ctx, c.sqsClient, c.queueURL, msg); err != nil {
					c.errorHandler.Handle(ctx, err)
					c.errorEncoder(ctx, err, msg, c.sqsClient)
				}
			}
			return ctx
		}
		c.before = append(c.before, deleteBefore)
	}
}

// ConsumerDeleteMessageAfter returns a ConsumerOption that appends a function
// that delete a message from queue to the list of consumer's after functions.
func ConsumerDeleteMessageAfter() ConsumerOption {
	return func(c *Consumer) {
		deleteAfter := func(ctx context.Context, msg *sqs.Message, _ *sqs.SendMessageInput, leftMsgs *[]*sqs.Message, mux *sync.Mutex) context.Context {
			if err := deleteMessage(ctx, c.sqsClient, c.queueURL, msg); err != nil {
				c.errorHandler.Handle(ctx, err)
				c.errorEncoder(ctx, err, msg, c.sqsClient)
			}
			return ctx
		}
		c.after = append(c.after, deleteAfter)
	}
}

func deleteMessage(ctx context.Context, sqsClient sqsiface.SQSAPI, queueURL string, msg *sqs.Message) error {
	_, err := sqsClient.DeleteMessageWithContext(ctx, &sqs.DeleteMessageInput{
		QueueUrl:      &queueURL,
		ReceiptHandle: msg.ReceiptHandle,
	})
	return err
}
```

Is there anything more I can do if an error happens when deleting the message ?

xyluet: You have error handler and error encoder function? just invoke it inside `deleteMessage`. 

If you want cancellation, you can cancel the context, and your endpoint should select if the context has been cancelled.

0marq: This would require adding a Context.CancelFunc in before and after parameters and calling the function in case of an error right ? I am going to wait for the response of https://github.com/go-kit/kit/pull/1018#issuecomment-707571384 before committing any changes here

0marq:
I understand.
I'll probably go towards @xyluet approach which is a lot less complex on Go kit's side : 
* The user is responsible of calling ReceiveMessage in his main.
* The user will call Go kit's consumer.ServeMessage to serve each received message.

0marq:
Given the new input from @peterbourgon I have simplified the code base : Receiving messages, handling multiple messages and updating visibilityTimeouts are left to the user.
We simply provide a Consumer.ServeMessage and Producer.Endpoint allowing the normal Go kit message processing and publishing flow, with a certain number of built in before/after & encode/decode functions.

transport/awssqs/consumer.go
```diff
@@ -0,0 +1,227 @@
...
+				cancel()
+			}
+			return ctx
+		}
+		c.after = append(c.after, deleteAfter)
+	}
+}
+
+// ServeMessage serves an SQS message.
+func (c Consumer) ServeMessage(ctx context.Context, msg *sqs.Message) error {
```
peterbourgon, line 132: Does this implement an interface expected of something in the AWS SQS package? How should callers use it?

0marq: No not really.
It should be used as depicted by @xyluet in https://github.com/go-kit/kit/pull/1018#issuecomment-707457997

peterbourgon: I think there is some conceptual disconnect here. 

A Go kit service is comprised of 1 or more endpoints, and each endpoint is exposed via 1 or more transports. By convention every Go kit transport package provides a type that exposes a single endpoint with unique DecodeRequest and EncodeResponse functions. But that type should compose into a larger "unit" which represents an entire service. 

For example, transport/http.Server type isn't an sttdlib http.Server itself but an http.Handler, and you're supposed to mount multiple transport/http.Server handlers in a single mux to represent your service. Or, transport/nats.Subscriber isn't it's own client and consumer of the NATS topic, instead it implements nats.MsgHandler so that it can be composed by the caller into a larger consumer that manages the whole service.

Concretely: users of this package shouldn't have to run ServeMessage loops for every transport/awssqs.Consumer they create (i.e. every endpoint in their service). They should create one SQS client/consumer/whatever with 1 or more transport/awssqs.Consumer types, each of which is fed messages by the "outer" component appropriately. I don't know the architecture of the SQS client lib so I don't know if this is natively supported or would have to be provided in this package too.

Does this make sense?

xyluet: It doesn't implement anything. I just borrow from amqp transport.
https://github.com/go-kit/kit/blob/6c170213364cbdae29ce60c4a505ec24c886172f/transport/amqp/subscriber.go#L98

SQS SDK provides almost the same way as RabbitMQ when receiving messages. RabbitMQ is using channel, and we need to for loop the channel and pass a `Delivery`. SQS has the same way, but we need to call `ReceiveMessage` in for loop and getting array of messages from receive message output.

0marq: any update on how you feel about this, knowing that this pattern is already used in amqp transport @peterbourgon ?

peterbourgon: Oh, man, I typed a big comment in reply but I guess it didn't save? :( Will re-type in a bit here, sorry!

peterbourgon: An important difference between AMQP's ServeDelivery and your ServeMessage is that ServeDelivery doesn't "run" — it returns a function that takes a received amqp.Delivery, runs through the request and response flow, and then emits a response to the provided amqp.Channel. Users need to feed the returned function deliveries, but they can do that however they like, they don't necessarily need to manage a goroutine per Subscriber to do so.

So ServeMessage could work that way, maybe. Or maybe there's a better fit. But the package shouldn't require callers to `go consumer.Anything()` to work properly.

xyluet: So ServeMessage should return function like AMQP?
```go 
ServeMessage(<maybe SQS client and/or cancellation context>) func(msg *sqs.Message) 
```

peterbourgon: If that works with the SQS SDK, then sure. Maybe another approach is better. I can't judge.

0marq: I don't really see how this solves the problem, you'd still end up doing something like

``` go
subscriber := sqstransport.NewSubscriber(endpoint, dec, enc, opts...)

for {
  out, _ := sqsClient.ReceiveMessage(ctx, input) //handle error
  for _, msg := range out.Messages {
    consumer.ServeMessage(ctx)(msg)
  }
}
```

or

``` go
subscriber := sqstransport.NewSubscriber(endpoint, dec, enc, opts...)
serveFunc := consumer.ServeMessage(ctx)

for {
  out, _ := sqsClient.ReceiveMessage(ctx, input) //handle error
  for _, msg := range out.Messages {
    serveFunc(msg)
  }
}
```

Is there something I didn't catch ?

The big difference with amqp is that, as far as I know, the SQS SDK does not provide a function that you can call to setup a function that will serve arriving messages automatically.

You need to fetch the messages by your self and handle them on your own. So we have two possibilities :
* We include fetching the messages in go-kit but this would mean including complex code to allow visibility timeout extension and other SQS features that are necessary for some users.
* We let the end user fetch the messages and calling the function that serves the messages.

In both cases the user will need to call `c.ServeMessage` or `c.Consume`, I personally see no other way.

How about you @xyluet ?

xyluet: Just like AMQP, they could just pass `Channel` and call `Consume` inside. 

I prefer let the end user fetch the messages and calling the function that serves a message. We can still use context with closure for other things, e.g. for acking before or after endpoint, etc.

0marq: I just changed the ServeMessage method signature to return a function. Is this what both of you had in mind ?

0marq: Any update guys ?

0marq: Hello !

What are your thoughts about the last changes I committed https://github.com/go-kit/kit/pull/1018/commits/b673cbefc4406a0cbfc7f29de081fc62042316f3 ?

0marq: @xyluet @peterbourgon Any update on this ?

ggranberry:
Is this still being worked on? I'd love to use it for a project

transport/awssqs/consumer.go
JasonTruter: > I think there is some conceptual disconnect here.
> 
> A Go kit service is comprised of 1 or more endpoints, and each endpoint is exposed via 1 or more transports. By convention every Go kit transport package provides a type that exposes a single endpoint with unique DecodeRequest and EncodeResponse functions. But that type should compose into a larger "unit" which represents an entire service.
> 
> For example, transport/http.Server type isn't an sttdlib http.Server itself but an http.Handler, and you're supposed to mount multiple transport/http.Server handlers in a single mux to represent your service. Or, transport/nats.Subscriber isn't it's own client and consumer of the NATS topic, instead it implements nats.MsgHandler so that it can be composed by the caller into a larger consumer that receives multiple message types and dispatches them to the appropriate MsgHandler.
> 
> Concretely: users of this package shouldn't have to run ServeMessage loops for every transport/awssqs.Consumer they create (i.e. every endpoint in their service). They should create one SQS client/consumer/whatever with 1 or more transport/awssqs.Consumer types, each of which is fed messages by the "outer" component appropriately. I don't know the architecture of the SQS client lib so I don't know if this is natively supported or would have to be provided in this package too.
> 
> Does this make sense?

SQS has this concept of long polling and short polling. Long polling would be a long lived connection that lives for a maximum of 20 seconds, allowing the poller to return a response only when there is a message available (cheaper due to less network calls). With short polling, a response is returned immediately but may have an empty response(may require multiple requests to check for a message on the queue).

This is the receive message api endpoint from sqs: 

```ReceiveMessage(input *ReceiveMessageInput) (*ReceiveMessageOutput, error)```

```
// MsgHandler is a callback function that processes messages delivered to
// asynchronous subscribers.
type MsgHandler func(msg *Msg)
```

sagikazarmark:
My two cents: I don't necessarily think that go-kit has to be the central repository of every possible transport implementation. The readme could mention third-party implementations for easy discoverability.

Another thing that I'd require from these submissions is that they are already released in a third-party repository, have a few users in production, etc. Go-kit is released about once a year, so it's not the right place for testing new transports IMHO.

PR #1029, opened 2020-11-03
Dependency update [lightstep-tracer-go/go-opencensus]

### LINKED ISSUE #1000: Dependency update [lightstep-tracer-go/go-opencensus]
Hello, we are using a dependency scanning tool and found an issue in the go.sum. The example application: **addsvc** uses the lightstep-tracer-go and the full go-kit library (go.mod/go.sum) also contains this dependency. 

The issue with lightstep-tracer-go v0.18.1 is containing the version of apache-thrift v0.12.0 with some vulnerabilities:
[[1]](http://mail-archives.apache.org/mod_mbox/thrift-dev/201910.mbox/%3CVI1PR0101MB2142E0EA19F582429C3AEBCBB1920%40VI1PR0101MB2142.eurprd01.prod.exchangelabs.com%3E)
[[2]](https://nvd.nist.gov/vuln/detail/CVE-2019-0205)

I know that this doesn't affect the main code of go-kit library, but the dependency scanning tool doesn't like this vulnerability and shows it with a high severity level. I have checked the last version of **lightstep-tracer-go** v0.20.0 and they have updated the opencensus to v0.22.3 and the issues with the apacha-thrift is fixed (updated to v0.13.0).
 
My proposal is to update the dependency version in go.mod file to have the latest version of this tracer [to v0.20.0]

ChrisHines:
Fixes #1000


go.sum
```diff
@@ -1,11 +1,12 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
-cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
peterbourgon, line 2: Did you manually remove this (and other) lines from go.sum, or was it done automatically somehow?

ChrisHines: Just `go mod tidy`.

ChrisHines:
Also the upgrade button in the go.mod file provided by VS Code. No manual
edits by me though.

go.sum
vrazdalovschi: Yes, go mod tidy

peterbourgon: Strange, I thought `go.sum` was append-only.

peterbourgon:
Sure, although I want to observe that this won't actually fix the problem. Go modules' solving algorithm, MVS, can substitute a given dependency+version with any "compatible" version of that dependency, and "compatibility" under major version 0 is basically "any version".

```
$ tree
.
├── a
│   ├── a.go
│   ├── go.mod
│   └── go.sum
└── b
    ├── go.mod
    ├── go.sum
    └── main.go
```
```
$ cat a/go.mod a/a.go
module a

go 1.15

require (
        github.com/lightstep/lightstep-tracer-go v0.22.0
)

package a

import (
        "github.com/lightstep/lightstep-tracer-go"
)

func Version() string {
        return lightstep.TracerVersionValue
}
```

```
$ cat b/go.mod b/main.go
module b

go 1.15

require (
        a v0.0.1
)

replace (
        a => ../a
        github.com/lightstep/lightstep-tracer-go => github.com/lightstep/lightstep-tracer-go v0.18.1
)

package main

import (
        "a"
)

func main() {
        println(a.Version())
}
```

```
$ cd b ; and go build ; and ./b
0.18.1
```

go.sum
ChrisHines: The checksum database is append only, and go.sum used to be append only, but see https://github.com/golang/go/issues/26381 for when the thought process changed and https://golang.org/ref/mod#go for where it is documented that `go mod tidy` removes unnecessary entries.

ChrisHines:
> Sure, although I want to observe that this won't actually fix the problem. Go modules' solving algorithm, MVS, can substitute a given dependency+version with any "compatible" version of that dependency, and "compatibility" under major version 0 is basically "any version".

> replace (
>         a => ../a
>         github.com/lightstep/lightstep-tracer-go => github.com/lightstep/lightstep-tracer-go v0.18.1
> )

Of course the main module can override anything with a replace in its go.mod. That's what replace is for and outside the scope of anything that we should worry about in the Go kit `go.mod` file.

PR #1030, opened 2020-11-03
Update Go versions for CI

ChrisHines:


ChrisHines:
The builds keep timing out. Go 1.14 succeeded once, but 1.15 has not succeeded at all. I am skeptical that this is a Go version issue, but I'm going to stop rerunning the builds for now.

peterbourgon:
Done in #1054.

PR #1033, opened 2020-11-09
Shutdown manager

hubaxis:
Need to shutdown connection manager when context.Done()

### DIFF
--- a/util/conn/manager.go
+++ b/util/conn/manager.go
@@ -25,7 +25,7 @@ type AfterFunc func(time.Duration) <-chan time.Time
 // the connection is invalidated, and a new connection is established.
 // Connection failures are retried after an exponential backoff.
 type Manager struct {
-	d       Dialer
+	dialer  Dialer
 	network string
 	address string
 	after   AfterFunc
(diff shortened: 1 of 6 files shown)


util/conn/manager_test.go
```diff
@@ -91,6 +120,29 @@ func TestManager(t *testing.T) {
 	}
 }
 
+func TestShoutDown(t *testing.T) {
```
peterbourgon, line 123: Oops.

util/conn/manager.go
```diff
@@ -97,9 +112,17 @@ func (m *Manager) loop() {
 
 	for {
 		select {
+		case <-m.ctx.Done():
+			_ = m.logger.Log("exit from loop")
+			err := conn.Close()
+			if err != nil {
+				_ = m.logger.Log("connection close err", err)
+			}
```
peterbourgon, line 120:
```suggestion
			m.logger.Log("exit from loop")
			if err := conn.Close(); err != nil {
				m.logger.Log("connection close err", err)
			}
```

peterbourgon (changes requested):

Issue #1035, opened 2020-11-16
StdLibAdapter can not extract log line with set prefix correctly


soven:
When using a prefix in stdlib's logger StdLibAdapter extracts the fields from log line incorrectly. 

For example, the program
```go
package main

import (
    "log"
    "os"

    gokitlog "github.com/go-kit/kit/log"
)

func main() {
    logg := log.New(gokitlog.NewStdlibAdapter(gokitlog.NewLogfmtLogger(os.Stdout)),
        "some prefix ", log.LstdFlags)
    logg.Println("log line")
}
```
will give us the following to the stdout:
```
caller="some prefix 2020/11/16 22:33" msg=":04 log line"
```

I have a fix to support the prefix mode by adding a method `Prefix(prefix string, joinPrefixToMsg bool) StdlibAdapterOption`

ChrisHines:
We would be happy to review a PR to fix this issue.

soven:
done

sagikazarmark:
Fixed in #1036

PR #1036, opened 2020-11-16
fix log stdlibadapter when prefixed

soven:


log/stdlib.go
```diff
@@ -49,6 +52,11 @@ func MessageKey(key string) StdlibAdapterOption {
 	return func(a *StdlibAdapter) { a.messageKey = key }
 }
 
+// Prefix sets value of the prefix for the actual log message. By default, there's no prefix.
```
peterbourgon, line 55: What does setting the prefix do?

soven: It allows client code to use the adapter for stdlib logger which is set by the prefix. There is an example in [the issue](https://github.com/go-kit/kit/issues/1035).

peterbourgon: Can you please add some more detail to the doc comment?

log/stdlib.go
```diff
@@ -26,9 +27,11 @@ func (w StdlibWriter) Write(p []byte) (int, error) {
 // messages, and place them under relevant keys.
 type StdlibAdapter struct {
 	Logger
-	timestampKey string
-	fileKey      string
-	messageKey   string
```
whiteboy19787, line 31: I'm not sure what to do this is new to me

log/stdlib_test.go
```diff
@@ -72,6 +72,56 @@ func TestStdLibAdapterExtraction(t *testing.T) {
 	}
```
whiteboy19787, line 72:
```suggestion
	}grab bin
```

log/stdlib.go
soven: sure. done

soven:
looks like something wrong with the CI

log/stdlib.go
```diff
@@ -49,6 +52,21 @@ func MessageKey(key string) StdlibAdapterOption {
...
+// there's no prefix.
+//
+// When you create a stdlib Logger, you can define a log prefix like this:
+//
+//		logger := log.New(w, "some prefix", log.LstdFlags)
+//
+// To correctly extract stdlib logger fields, you should set the same prefix
+// by Prefix stdlib adapter option. Also you can set joinPrefixToMsg if you need
+// to join the prefix to extracted message fields. For example, you can need it
+// when using log.Lmsgprefix flag.
```
peterbourgon, line 65:
```suggestion
// Prefix configures the adapter to parse a prefix from stdlib log events. If
// you provide a non-empty prefix to the stdlib logger, then your should provide
// that same prefix to the adapter via this option.
//
// By default, the prefix isn't included in the msg key. Set joinPrefixToMsg to
// true if you want to include the parsed prefix in the msg.
```

peterbourgon:
Thank you!

PR #1044, opened 2020-12-28
bug fix: etcdv3 not auto register when etcev3 server disconnected

bcw104:
etcdv3 not auto register when etcev3 server disconnected

### DIFF
--- a/sd/etcdv3/client.go
+++ b/sd/etcdv3/client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"time"
+	"fmt"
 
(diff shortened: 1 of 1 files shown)


peterbourgon:
Hi, thanks for the contribution! Can you please add a test which fails without this PR, and passes with it?

bcw104:
@peterbourgon The steps of reproduction the issue are as follows: 
1. Shutdown the etcd server for a while, make sure the service key is expired. 
2. Startup the etcd server, the etcd client will reconnect, but registar can't auto register the service key.

ChenHaoHu:
this pr solve my problem

PR #1049, opened 2021-01-12
only use the eureka instances which status are UP

tiangles:
for some reason, applications may have status like "down", "offline"
In this case, we should not use them as normal applications anymore.


peterbourgon:
Can you please add a test which fails without this PR and passes with it? Thank you!

PR #1054, opened 2021-02-09
Update Go versions for CI (v2)

peterbourgon:
Cleaner version of #1030.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.13', '1.14']
+        go: ['1.14', '1.15']
     env:
       GOFLAGS: -mod=readonly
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.13', '1.14']
+        go: ['1.14', '1.15']
```
adriancole, line 15: curious and feel free to not answer if busy. what would make you decide to extend the range to 1.16? and if so, would you lift the floor also?

peterbourgon: Well, 1.16 wasn't released when I made this change :)

peterbourgon: And yes, the intent is to have this track the currently supported Go versions, which are the current version and the previous version, so technically should be 1.15 and 1.16, now.

PR #1062, opened 2021-02-24
Updates CI to latest go

codefromthecrypt:
Added a one-liner happy to remove if unnecessary


codefromthecrypt:
PS if you are in the mood, I can run the update deps script and see if anything breaks and/or clear up deprecation

PR #1064, opened 2021-02-25
improve code quality

withshubh:
#### Description
This PR fixes a few issues that were affecting the code quality.

#### Summary of changes

- Fix unnecessary calls to Printf
- Fix nil context being passed to function
- Remove empty default in select
- Replace `strings.Index` with `strings.Contains`
- Replace `bytes.Compare` with `bytes.Equal`
- Merge variable declaration with assignment
- Remove unnecessary use of `fmt.Sprintf


sagikazarmark (approved):

sd/etcd/client_test.go
```diff
@@ -135,7 +135,7 @@ func (fw *fakeWatcher) Next(context.Context) (*etcd.Response, error) {
 			return nil, nil
 		case <-fw.err:
 			return nil, errors.New("error from underlying etcd watcher")
-		default:
+
```
peterbourgon, line 138: This results in different behavior, does it not?

sagikazarmark: It doesn't change the observed behavior (it'll still block until either one of the first two cases return something).

TBH I don't really understand why the for loop is necessary.

peterbourgon: Well if you have the default, you need the for, otherwise this is a one-and-done.

But it's still a terrible pattern, as it turns what could be a simple select into a hot loop?? Maybe an artifact from refactoring.

sagikazarmark: Probably, yeah. @withshubh could you please remove the for loop as well?

peterbourgon (changes requested):
Waiting for the fix to sd/etcd

sagikazarmark:
Apparently the user deleted the fork. I resubmitted the PR in #1168

peterbourgon:
Unless you intend to fix the etcd issue too, I'd just close it.

sagikazarmark:
I fixed that too.

PR #1067, opened 2021-03-06
Upgrade etcd to 3.5

sagikazarmark:
As of 3.5 etcd properly uses go modules.


peterbourgon:
Hooray!

sagikazarmark:
Sadly, it's still alpha though, hence the draft PR. But if you are okay with alpha versions, I can mark it as ready.

peterbourgon:
@sagikazarmark Is 3.4 impossible?

sagikazarmark:
@peterbourgon  3.4 won't receive proper go modules support. Furthermore, 3.5 introduces separate client modules which means less dependencies (we don't have to rely on the server stuff which probably has more deps).

peterbourgon:
Issues keep piling up related to this dependency. Can we justifiably use https://github.com/etcd-io/etcd/releases/tag/v3.5.0-beta.3?

sagikazarmark:
TBH I wouldn't recommend that. At least there were quite a few fixes to the release pipeline, so I'd probably wait until it's more stable.

The stable should be close IMHO.

sagikazarmark:
Looks like the packaging issues were sorted out.

sd/etcd/client.go
```diff
@@ -10,7 +10,7 @@ import (
 	"net/http"
 	"time"
 
-	etcd "go.etcd.io/etcd/client"
+	etcd "go.etcd.io/etcd/client/v2"
```
peterbourgon, line 13: Should this be v3?

go.mod
```diff
@@ -42,20 +43,22 @@ require (
...
 	github.com/samuel/go-zookeeper v0.0.0-20190923202752-2cc03de413da
 	github.com/sirupsen/logrus v1.4.2
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	github.com/sony/gobreaker v0.4.1
 	github.com/streadway/amqp v0.0.0-20190827072141-edfb9018d271
 	github.com/streadway/handy v0.0.0-20190108123426-d5acb3125c2a
-	go.etcd.io/etcd v0.0.0-20191023171146-3cf2f69b5738
+	go.etcd.io/etcd/client/pkg/v3 v3.5.0-beta.4
+	go.etcd.io/etcd/client/v2 v2.305.0-beta.4
+	go.etcd.io/etcd/client/v3 v3.5.0-beta.4
```
peterbourgon, line 55: I can't say I understand why all of these are here. Maybe a quick summary?

sagikazarmark: `go.etcd.io/etcd/client/v2`: used by `sd/etcd` (v2 client became a separate module)
`go.etcd.io/etcd/client/v3`: used by `sd/etcdv3`
`go.etcd.io/etcd/client/pkg/v3`: some common client code has been extracted to separate a pkg module (go-kit uses the transport package from there)

etcd decided to slice up the project into smaller modules.

Does that help?

peterbourgon:
#1097

PR #1070, opened 2021-03-19
Fix typo in jwt package

amidam:
* change JWTTokenContextKey to JWTContextKey

    * revise errors as well as comments

### DIFF
--- a/auth/jwt/middleware.go
+++ b/auth/jwt/middleware.go
@@ -12,9 +12,14 @@ import (
 type contextKey string
 
 const (
-	// JWTContextKey holds the key used to store a JWT in the
-	// context.
-	JWTContextKey contextKey = "JWT"
+	// JWTContextKey holds the key used to store a JWT in the context.
+	JWTContextKey contextKey = "JWTToken"
+
+	// JWTTokenContextKey is an alias for JWTContextKey.
+	//
+	// Deprecated: only exists for historical compatibility.
+	// JWTContextKey would be preferable.
+	JWTTokenContextKey = JWTContextKey
 
 	// JWTClaimsContextKey holds the key used to store the JWT Claims in the
 	// context.
--- a/auth/jwt/middleware.go
+++ b/auth/jwt/middleware.go
@@ -17,8 +17,7 @@ const (
 
 	// JWTTokenContextKey is an alias for JWTContextKey.
 	//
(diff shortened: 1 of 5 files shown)


amidam:
fixes #1071

auth/jwt/middleware.go
```diff
@@ -12,9 +12,9 @@ import (
 type contextKey string
 
 const (
-	// JWTTokenContextKey holds the key used to store a JWT Token in the
+	// JWTContextKey holds the key used to store a JWT in the
 	// context.
-	JWTTokenContextKey contextKey = "JWTToken"
+	JWTContextKey contextKey = "JWT"
```
peterbourgon, line 17: This would be a breaking change in two ways: the API identifier itself has changed, and the actual context key has changed. Let's avoid that.

```suggestion
	JWTContextKey contextKey = "JWTToken"

	// JWTTokenContextKey is an alias for JWTContextKey.
	//
	// Deprecated: exists for historical compatibility only, prefer JWTContextKey.
	JWTTokenContextKey contextKey = JWTContextKey
```

peterbourgon (changes requested):

auth/jwt/middleware.go
amidam: I understand. So should I close the issue? or do you have any plan regarding the case in the future?

amidam: I understand. Upon your suggestion, I revert the value to `JWTToken` and add `JWTTokenContextKey` for historical compatibility.

peterbourgon:
Can you please leave the old identifiers in place, and mark them as deprecated?

amidam:
> Can you please leave the old identifiers in place, and mark them as deprecated?

I have done it already. I added a new constant called `JWTContextKey` and marked `JWTTokenContextKey` as a deprecated constant.

amidam:
> > Can you please leave the old identifiers in place, and mark them as deprecated?
> 
> I have done it already. I added a new constant called `JWTContextKey` and marked `JWTTokenContextKey` as a deprecated constant.

Still they both have the same value

peterbourgon (changes requested):
One minor thing then 👍

amidam:
> One minor thing then +1

I have done it.

sagikazarmark (approved):
LGTM

PR #1072, opened 2021-03-22
tracing.opentracing endpoint middleware options

### LINKED ISSUE #804: Add error tag to spans on OT endpoint middleware

alebabai:
# what

- Add `TraceEndpoint` function to not rely on client/server kinds.
- TraceServer and TraceClient should differ only in span kind tag value
- Extend `tracing.opentracing` endpoint middleware API with some options (like it was done for Opencensus):
  - optionally trace business errors
  - add option to set static/dynamic tags
  - add option to set dynamic name based on context and previous name
  - bump opentracing lib (to have `ext.LogError` function)

# why

closes #804

### DIFF
--- a/tracing/opentracing/endpoint.go
+++ b/tracing/opentracing/endpoint.go
@@ -31,7 +31,7 @@ func TraceEndpoint(tracer opentracing.Tracer, operationName string, opts ...Endp
 				}
 			}
 
-			span := opentracing.SpanFromContext(ctx)
+			var span opentracing.Span
 			if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
 				span = tracer.StartSpan(
 					operationName,
--- a/tracing/opentracing/endpoint.go
+++ b/tracing/opentracing/endpoint.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	"github.com/opentracing/opentracing-go"
-	otext "github.com/opentracing/opentracing-go/ext"
+	"github.com/opentracing/opentracing-go/ext"
 
 	"github.com/go-kit/kit/endpoint"
 )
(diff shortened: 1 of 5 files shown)


peterbourgon:
@basvanbeek

alebabai:
Guys, Is it any chance to get it reviewed?

basvanbeek:
Hi @alebabai, thanks for the PR and sorry for the late action. I will review tomorrow.

tracing/opentracing/endpoint.go
```diff
@@ -4,52 +4,98 @@ import (
...
+	}))
+
+	return TraceEndpoint(tracer, operationName, opts...)
+}
+
+// TraceClient returns a Middleware that wraps the `next` Endpoint in an
+// OpenTracing Span called `operationName` with client span.kind tag.
+func TraceClient(tracer opentracing.Tracer, operationName string, opts ...EndpointOption) endpoint.Middleware {
+	opts = append(opts, WithTags(map[string]interface{}{
+		ext.SpanKindRPCServer.Key: ext.SpanKindRPCClient.Value,
```
basvanbeek, line 77:
```suggestion
		ext.SpanKindRPCClient.Key: ext.SpanKindRPCClient.Value,
```

tracing/opentracing/endpoint.go
```diff
@@ -4,52 +4,98 @@ import (
...
 			}
-			defer clientSpan.Finish()
-			otext.SpanKindRPCClient.Set(clientSpan)
-			ctx = opentracing.ContextWithSpan(ctx, clientSpan)
-			return next(ctx, request)
+
+			ctx = opentracing.ContextWithSpan(ctx, span)
+
+			response, err := next(ctx, request)
+			if err := identifyError(response, err, cfg.IgnoreBusinessError); err != nil {
```
basvanbeek, line 54: Would like more symmetry with the OpenCensus implementation and also include the handling of Go kit load balancer retry errors.

basvanbeek: it would also be nice that even if an ignore business error is set to true, we would still see it being tagged with "gokit.business.error", again in-line with the OpenCensus implementation.

basvanbeek (changes requested):
Great effort. Only a few nits mostly with respect to having more symmetry with the OpenCensus middleware.

tracing/opentracing/endpoint.go
alebabai: Fixed, thanks

tracing/opentracing/endpoint.go
alebabai: Yep, you're right.
Implemented it in way as it was done in opencensus, now it support all the stuff.

basvanbeek (approved):
@alebabai thank you so much for hanging in there and having the patience to wait for me to finish the review. The code is now finally inline with functionality as found in the other tracers so I'm grateful for the contribution!

PR #1077, opened 2021-04-12
Create codeql-analysis.yml

jfcg:
Add Github CodeQL security checks for Go.


peterbourgon:
I've looked at these in another context before. IMO they are more noise than signal. Thanks, though.

PR #1082, opened 2021-04-25
upgrade nats-server

AbhilashVijayakumar:
Upgrade NATS Server due to ddos vulnerability :
 * Version 2 prior to 2.2.0
   + 2.0.0 through and including 2.1.9 are vulnerable
 * fixed with nats-io/nats-server commit 423b79440c (2021-03-14)

 refer https://advisories.nats.io/CVE/CVE-2021-3127.txt


ChrisHines:
This seems fine to me, but it should be noted that go-kit only uses nats-server in tests. Also, this change will not propagate to to go-kit dependents until we tag a new release and we should check our go.mod for out of date dependencies when we are preparing a new go-kit release anyway.

sagikazarmark:
Would it make sense to rewrite those tests to use an external nats server instance (similarly to etcd, zookeeper, etc)?

I understand the value of self-contained tests, but this could make the dependency graph smaller.

peterbourgon:
@AbhilashVijayakumar Ping :) Can you fix the tests, please?

peterbourgon:
#1095

Issue #525, opened 2017-05-11
in memory generic metrics aren't shared across `With`


terinjokes:
A common pattern is to use `metrics.*.With` in an endpoint middleware to created derived metrics:

```go
func (s *instrumentingService) Lookup(ctx context.Context) (bool, error) {
	defer func(begin time.Time) {
		s.requestCount.With("method", "lookup").Add(1)
	}(time.Now())

	return s.Service.Lookup(ctx)
}
```

If you're using the generic metrics, there's no way to retrieve these metrics; the in memory metrics aren't shared across calls to `With`.

```go
func TestCounter(t *testing.T) {
	var counter metrics.Counter
	counter = generic.NewCounter("requestCount")

	var svc Service
	svc = NewService()
	svc = instrumentingService{requestCount: counter, Service: svc}

	svc.Lookup(context.Background())

	if count := counter.With("method", "lookup").(*generic.Counter).Value(); count != float64(1) {
		t.Fatalf("expected '%d', got '%d'", float64(1), count)
	}
}

// expected '1', got '0'
```

peterbourgon:
This is by design. Generic metrics aren't really meant to be used outside of a test context.

terinjokes:
@peterbourgon I'm trying to use them in test, I don't see how they're useful there either, as the example shows. I think the design needs to be tweaked just a little bit.

peterbourgon:
I'd happily accept a PR.

peterbourgon:
I am no longer happy to accept a PR here, see #570. Sorry folks :(

PR #1083, opened 2021-04-27
Fixing generic.Counter.With() - closes issue #525

ynodir:
Every `With()` call was returning a new instance which wouldn't work unless they were reassigned back to original variable.

Specifically, calls to `Value()` and `ValueReset()` would be returning original instance's (the one created with `NewCounter()`) values.

Reworked it to return the same instance but with updated `lvs` field.

### DIFF
--- a/metrics/generic/generic_test.go
+++ b/metrics/generic/generic_test.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"math"
 	"math/rand"
+	"reflect"
 	"sync"
 	"testing"
 
@@ -29,6 +30,17 @@ func TestCounter(t *testing.T) {
 	value := counter.Value
 	if err := teststat.TestCounter(counter, value); err != nil {
 		t.Fatal(err)
+	}
+}
+
+func TestIssue525(t *testing.T) {
+	counter := generic.NewCounter("issue525_counter")
+	sameCounter := counter.With("label", "counter").(*generic.Counter)
+	if want, have := counter, sameCounter; want != have {
+		t.Errorf("Instance: want %p, have %p", want, have)
+	}
+	if want, have := []string{"label", "counter"}, sameCounter.LabelValues(); !reflect.DeepEqual(want, have) {
+		t.Errorf("Label values mismatch: want %+v, have %+v", want, have)
 	}
 }
 
--- a/metrics/generic/generic_test.go
+++ b/metrics/generic/generic_test.go
@@ -34,13 +34,20 @@ func TestCounter(t *testing.T) {
 }
 
 func TestIssue525(t *testing.T) {
-	counter := generic.NewCounter("issue525_counter")
-	sameCounter := counter.With("label", "counter").(*generic.Counter)
-	if want, have := counter, sameCounter; want != have {
-		t.Errorf("Instance: want %p, have %p", want, have)
-	}
-	if want, have := []string{"label", "counter"}, sameCounter.LabelValues(); !reflect.DeepEqual(want, have) {
+	counterWithoutLabels := generic.NewCounter("issue525_counter")
+	counterWithLabels := counterWithoutLabels.With("label", "counter").(*generic.Counter)
+	if want, have := []string{"label", "counter"}, counterWithLabels.LabelValues(); !reflect.DeepEqual(want, have) {
 		t.Errorf("Label values mismatch: want %+v, have %+v", want, have)
+	}
+
+	counterWithoutLabels.Add(1)
+	counterWithLabels.Add(1)
+	counterWithLabels.Add(1)
+	if want, have := float64(1), counterWithoutLabels.Value(); want != have {
+		t.Errorf("Wrong counter value: want %f, have %f", want, have)
+	}
+	if want, have := float64(2), counterWithLabels.Value(); want != have {
+		t.Errorf("Wrong counter value: want %f, have %f", want, have)
 	}
 }
 
--- a/metrics/generic/generic.go
+++ b/metrics/generic/generic.go
@@ -32,11 +32,8 @@ func NewCounter(name string) *Counter {
 
 // With implements Counter.
 func (c *Counter) With(labelValues ...string) metrics.Counter {
-	return &Counter{
-		Name: c.Name,
-		bits: atomic.LoadUint64(&c.bits),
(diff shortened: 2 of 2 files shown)


peterbourgon:
```go
c0 := generic.NewCounter("my_counter")
c1 := c0.With("component", "foo")
c1.Add(1) // this results in: my_counter{component=foo} 1
c0.Add(1) // does this result in: my_counter{} 1, or my_counter{component=foo} 2?
```

ynodir:
It'll be 2, of course.
Why'd you expect 1?
With `With()`, you're adding labels to the **same** counter.

Here's a counterexample for you:
```
c0 := generic.NewCounter("my_counter")
c1 := c0.With("component", "foo")
c1.With("component", "foo").Add(1) // do you expect here `c1` to be my_counter{component=foo} 0 or my_counter{component=foo} 1?
```

With what you're implying in your example, `generic.Counter.Value() float64` doesn't make sense since we'd need to have different values for each distinct lvs. It'd need to return an array of `generic.Counter`s, one per each distinct `With()` it was called with plus the original instance.

ynodir:
Alright, @peterbourgon, reworked the PR using `lv.Space` as was done by PR #570 - heavily inspired by it.

There're 2 issues which I'd like your advise on:
1. Using `reflect.DeepEqual()` to compare `lvs`es in `generic.go` - I guess, it's slower than writing our own comparer function - should I replace the `DeepEqual` with our own comparator?
2. Ideally, `lv.Space` should be used on a level above `generic.Counter` as it can contain multiple counters since it accounts for `name` already. But, I guess, we can live with that for now?

ynodir:
@peterbourgon would appreciate any update on this

peterbourgon:
`With` should obviously not mutate the thing it's called on, so it should definitely be `my_counter{} 1`.

 It seems your latest change has that result, do I have it right?

ynodir:
> 
> 
> `With` should obviously not mutate the thing it's called on, so it should definitely be `my_counter{} 1`.
> 
> It seems your latest change has that result, do I have it right?

You're right. It was my misunderstanding of how counters work.
Yes, last commit works as expected.

ynodir:
@peterbourgon if you're ok with the changes, can we merge this, please?

metrics/generic/generic_test.go
```diff
@@ -32,6 +33,24 @@ func TestCounter(t *testing.T) {
 	}
 }
 
+func TestIssue525(t *testing.T) {
+	counterWithoutLabels := generic.NewCounter("issue525_counter")
```
peterbourgon, line 37: Should this not work the same for Gauge and Histogram?

peterbourgon:
Forgive my delay. I've looked at this in more detail. Your expectations about what the generic metrics should do are simply incorrect, and so this PR isn't a fix per se, it's a redefinition and reimplementation of the package. Well, part of it — just the Counter.

I'm not saying you're wrong for wanting them to work this way! It probably makes more sense than the current version. But the current version is what it is, and it's just not meant to be... used like this. If you want to use generic metrics like this, it should be a new package. But I probably wouldn't accept that PR, because (a) I don't see the utility (on this point I could be convinced), and also (b) I don't think package metrics + its subpackages are really in scope [in the long term](https://github.com/go-kit/kit/issue/843).

Sorry for the enormous RTT on this one and the unsatisfying conclusion.

ynodir:
Whoa, didn't expect the PR to be closed without a discussion.

The reason I've created this PR was to be able to use `generic.Counter` for unit-tests seeing as it provides a `Value()` function. I'm open to suggestions as to how to unit-test `metrics.Counter` without needing to write the same mock in every client anew. This is the answer to your utility question - ability to unit-test `metrics.Counter` without writing the same mock every time.
From what you're writing, my understanding is that `generic.Counter` isn't meant to be used in the clients as `Value()` function is, basically, useless without storing all of the instances of `generic.Counter` returned with `With()`. And, I'm sure, you're not expecting production code to do that. So, either a comment to that effect or making it inaccessible from the client code would be nice.
So, can I ask what's the purpose of `generic.Counter` is? The only place I found it being used is in `graphite` package, but even then `With()` is a no-op there (just returns the same instance) and `Value()` isn't even used. I wouldn't say that `generic.Counter` is an in-memory implementation as it doesn't provide an ability to read the values without storing each counter instance in the client code.

metrics/generic/generic_test.go
ynodir: Not sure. I know about `Gauge` and `Histogram` even less than I know about `Counter`s.

peterbourgon:
> The reason I've created this PR was to be able to use generic.Counter for unit-tests seeing as it provides a Value() function. I'm open to suggestions as to how to unit-test metrics.Counter without needing to write the same mock in every client anew.

```go
// Do this
base := generic.NewCounter(...)
other := base.With(...)
other.Add(1)
other.(typecast).Value() // will be 1

// Not this
base := generic.NewCounter(...)
base.With(...).Add(1)
base.(blahblah).Value() // will not be 1
```

In other words, the return value of With must be captured.

> So, can I ask what's the purpose of generic.Counter is? 

The very limited usage I described above, mostly as a way to exercise the metrics interfaces.

ynodir:
> > The reason I've created this PR was to be able to use generic.Counter for unit-tests seeing as it provides a Value() function. I'm open to suggestions as to how to unit-test metrics.Counter without needing to write the same mock in every client anew.
> 
> ```go
> // Do this
> base := generic.NewCounter(...)
> other := base.With(...)
> other.Add(1)
> other.(typecast).Value() // will be 1
> 
> // Not this
> base := generic.NewCounter(...)
> base.With(...).Add(1)
> base.(blahblah).Value() // will not be 1
> ```
> 
> In other words, the return value of With must be captured.

As I said, you shouldn't expect production code to capture the every new counter instance returned with `With()`.

peterbourgon:
> As I said, you shouldn't expect production code to capture every new counter instance returned with With().

That's fine, but the ship has sailed: that's how it works. If you want something else, it needs to come in the form of a new package, not changes to this one.

peterbourgon:
I will observe that there's no technical or structural reason a "fixed" generic.Counter, or a brand-new bettergeneric.Counter, needs to live in the Go kit repository. Go forth and do as thou wilt 😇

PR #1085, opened 2021-04-29
metrics/dogstatsd: Add support for Distribution metrics

pbitty:
Datadog has a [Distribution](https://docs.datadoghq.com/metrics/distributions/) metric type that is not currently supported by go-kit metrics.  This PR adds support for this metric type.

The Dogstatsd [datagram format documentation](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics) does not list Distributions as supporting sampling rates, but Datadog's own dogstatsd client does submit a sampling rate, so I am assuming that it is supported.

References:
* https://github.com/DataDog/datadog-go/blob/v4.6.0/statsd/statsd.go#L570-L579
* https://github.com/DataDog/datadog-go/blob/v4.6.0/statsd/format.go#L139-L141

This is an isolated enough change that I felt opening an issue was not necessary, but if you disagree let me know and I will open one.

### DIFF
--- a/metrics/dogstatsd/dogstatsd.go
+++ b/metrics/dogstatsd/dogstatsd.go
@@ -39,15 +39,16 @@ import (
 // To regularly report metrics to an io.Writer, use the WriteLoop helper method.
 // To send to a DogStatsD server, use the SendLoop helper method.
 type Dogstatsd struct {
-	mtx        sync.RWMutex
-	prefix     string
-	rates      *ratemap.RateMap
-	counters   *lv.Space
-	gauges     map[string]*gaugeNode
-	timings    *lv.Space
-	histograms *lv.Space
-	logger     log.Logger
-	lvs        lv.LabelValues
+	mtx           sync.RWMutex
+	prefix        string
+	rates         *ratemap.RateMap
+	counters      *lv.Space
+	gauges        map[string]*gaugeNode
+	distributions *lv.Space
+	timings       *lv.Space
+	histograms    *lv.Space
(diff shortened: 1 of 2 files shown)


peterbourgon:
The purpose of the Go kit metrics packages is to offer a common interface of Counters, Gauges, and Histograms, which can abstract over different backends. If a backend supports more than those 3 types, that's fine, but why would the Go kit package need to also provide that support?

pbitty:
Perhaps I used the wrong term when I said "metric type".   From go-kit's perspective, this is another implementation of the `metrics.Histogram` interface.  To a package consuming this as a dependency, there would be no difference, but in Datadog one would be able to access features that are only provided by this new "Distribution" implementation.

This is similar to the `prometheus` package, which offers `NewHistogram()` and `NewSummary()` as implementations of `metrics.Histogram`.  The `dogstatsd` package also offers two implementations via `NewHistogram()` and `NewTiming()`.

peterbourgon:
Oh, sorry, I didn't understand it was a Histogram. Let me look closer.

pbitty:
No problem.

I admit, this is basically a copy+paste of the Histogram implementation, with a `d` as the metric type submitted in the payload to dogstastd instead of `h`.  If you would like, I am happy to try reduce some duplication (in this PR or another one).

peterbourgon:
Wait, is it literally identical?

Why... does Dogstat do something different with this exactly identical data?

Yeah, let's have a flag or something in the Histogram constructor? Wild.

Issue #1092, opened 2021-05-25
Support for WebSockets in Server interface


wicknicks:
net/http added support for websockets in v1.12 ([discussion](https://github.com/golang/go/issues/26937)). The change however requires applications to use an additional interface ([Hijacker](https://github.com/golang/go/blob/go1.12/src/net/http/server.go#L179)) to 'hijack' the connection and access the I/O channels. When applications use go-kit, the http.ResponseWriter is replaced by an interceptingWriter here: https://github.com/go-kit/kit/blob/v0.10.0/transport/http/server.go#L99. This only [embeds](https://github.com/go-kit/kit/blob/v0.10.0/transport/http/server.go#L228) the original `http.ResponseWriter`. To open up support for websockets, we needs to embed the Hijacker interface as well.

PR #1093, opened 2021-05-25
Embed the Hijacker interface in intercepting writer

wicknicks:
This PR embeds the Hijacker interface in kit's interceptingWriter. Having an hijacker embedding allows applications to use websockets. looks here for discussion around this: https://github.com/golang/go/issues/26937

Addresses issue #1092.

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"github.com/felixge/httpsnoop"
+
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/transport"
@@ -95,15 +97,8 @@ func ServerFinalizer(f ...ServerFinalizerFunc) ServerOption {
 func (s Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	hj, hijackable := w.(http.Hijacker)
-
 	if len(s.finalizer) > 0 {
-		var iw *interceptingWriter
-		if hijackable {
-			iw = &interceptingWriter{w, hj, http.StatusOK, 0}
-		} else {
-			iw = &interceptingWriter{w, nil, http.StatusOK, 0}
-		}
+		iw := &interceptingWriter{httpsnoop.Wrap(w, httpsnoop.Hooks{}), http.StatusOK, 0}
 		defer func() {
 			ctx = context.WithValue(ctx, ContextKeyResponseHeaders, iw.Header())
 			ctx = context.WithValue(ctx, ContextKeyResponseSize, iw.written)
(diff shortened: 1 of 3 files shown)


transport/http/server.go
```diff
@@ -226,6 +233,7 @@ type Headerer interface {
 
 type interceptingWriter struct {
 	http.ResponseWriter
+	http.Hijacker
```
peterbourgon, line 236: This isn't safe — if this field is nil, `h, ok := w.(http.Hijacker)` will succeed, but calling `h.Hijack()` will panic, as `h` will be nil.

peterbourgon:
http.Hijacker is one of [five common interfaces](https://github.com/felixge/httpsnoop/) which http.ResponseWriters are upgraded-to using the interface upgrade pattern — why just solve for this one?

wicknicks:
hey, @peterbourgon thanks very much for the suggestions. yes, it makes sense to accommodate all five interfaces. such a [switch case block](https://github.com/felixge/httpsnoop/blob/master/wrap_generated_gteq_1.8.go#L73) was what I had in mind for that, but using `httpsnoop` to wrap the writer is a better approach. I'll update this PR shortly. thanks again!

wicknicks:
@peterbourgon made the changes. let me know what you think. thanks in advance!

wicknicks:
btw, I'm not very fluent in golang. does the `interceptingWriter` now include a `Hijack()` method? it seems like it should not. and the only way to do so, is actually have an entire [switch case block](https://github.com/felixge/httpsnoop/blob/master/wrap_generated_gteq_1.8.go#L73) around constructing `interceptingWriter` with the component interfaces.

codyaray (approved):

wicknicks:
@peterbourgon any thoughts on this approach? thanks in advance!

codyaray:
@peterbourgon what do you think? Can this be merged so we can use it? 🙏

transport/http/server.go
```diff
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"github.com/felixge/httpsnoop"
+
```
peterbourgon, line 9: Nit: rm the blank line and re-goimports :)

peterbourgon (approved):
With that change, OK. Not super excited about the additional dependency, but 🤷

wicknicks:
@peterbourgon removed the extra newline. `goimports` didn't bring out any other changes.

wicknicks:
@peterbourgon do you think we can merge this now, and release a new tag? many thanks in advance!

peterbourgon:
This is a big change to a core package. I'll merge it now, so we can get some beta-testing time, so to speak.

peterbourgon:
Oof! What am I saying. Can you please add some tests for this? 😇

peterbourgon:
Happy to re-review with proper test coverage.

PR #1099, opened 2021-06-04
update etcd to 3.5.0-rc.0

sagikazarmark:


ChrisHines (approved):

peterbourgon:
I would prefer not to track rcs and wait until we get the final release. Or is this the last one?

sagikazarmark:
I don't think it is. I don't mind waiting for the latest release, but an rc is better than a beta IMO.

peterbourgon:
Let's hold off on any more bumps until it's properly released.

PR #1100, opened 2021-06-04
Relocate examples to a separate repository

sagikazarmark:
Repurposed this PR to relocate examples to a separate repository.

Should be merged once https://github.com/sagikazarmark/examples is moved to go-kit.

---
This PR is a proof of concept of how a separate examples module would look like. Although submodules in a single repo are discouraged, this module is not supposed to be used as a dependency in other modules, so it should be fine. (The alternative would be moving examples to a separate repo, but I don't think that's a good idea)

The advantage of making the examples package a separate module is that the root module can get rid of quite a few dependencies only used in examples.

Some disadvantages of making this move:

The biggest risk is an ambiguous import path error, but it should be fine as long as people don't import the examples package (I think) which they shouldn't anyway. The alternative is renaming the package to something else.

`go get`ting the go-kit package will no longer download the examples, so any documentation will have to be updated. However, every since go modules, `go get` is no longer the primary way to download the source code for a module, so this shouldn't be a problem either.

Last, but not least: updating dependencies would become somewhat redundant (common dependencies need to be updated in both modules)

In my opinion, the benefits outweigh the costs.

WDYT? @peterbourgon @ChrisHines


peterbourgon:
Sorry, no. But breaking examples out to a separate repo + module would be fine!

sagikazarmark:
Do you want to preserve history (similar to the log module)? IMHO it's not that important for examples, but I can prepare a separate repo with the history preserved if you prefer that way.

ChrisHines:
The benefit of reducing the dependencies of `go-kit/kit` is worth breaking the examples out in some fashion. I'm not particular about making it a nested module or a separate repo, so I'll just second @peterbourgon's vote of a separate repo. I think it is polite to contributors to preserve the history, so I would prefer that approach.

sagikazarmark:
I prepared a repository here: https://github.com/sagikazarmark/kit-examples

I added you both as collaborators

I opened a PR with the necessary changes to make the repo functional: https://github.com/sagikazarmark/kit-examples/pull/1

I'd suggest merging the PR if you are satisfied with the changes and then transferring (and renaming) the repo.

peterbourgon:
Is it possible to transfer a repo in this way? Any docs? Never done it before.

sagikazarmark:
It's actually quite easy to do:

Képernyőfotó 2021-06-06 - 9 43 06

sagikazarmark:
Repurposed this PR to relocate examples to a separate repository.

Should be merged once https://github.com/sagikazarmark/examples is moved to go-kit.

sagikazarmark:
@peterbourgon GitHub doesn't allow me to move the repo: https://github.com/sagikazarmark/examples/pull/1#issuecomment-857281239 :

> Er... something is not right. GitHub says:
> 
> > You don’t have the permission to create public repositories on go-kit
> 
> 🤷

sagikazarmark:
So after reading the documentation, I need permission to create repos which I don't have at the moment.

@peterbourgon if you trust that I'm not going to abuse my new powers, you can give me repo create permission temporarily.

The alternative is what I wanted to do originally: I transfer the repo to you and you can transfer it to the go-kit organization.

peterbourgon:
Apologies for the delay. I should be back up and running at full capacity in about 2 weeks.

You should be able to create repos now, @sagikazarmark — can you confirm?

sagikazarmark:
Yep, thanks @peterbourgon. No worries!

The repo is moved here: https://github.com/go-kit/examples

If you are all happy with the result, we can merge this PR as well.

peterbourgon (approved):

PR #1101, opened 2021-06-04
Use external nats instance for integration tests

sagikazarmark:
This isn't a huge win in terms of dependencies (nats server is still going to be a transitive dependency), but it unifies how integration tests are executed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/hudl/fargo v1.3.0
 	github.com/influxdata/influxdb1-client v0.0.0-20191209144304-8bf82d3c094d
 	github.com/lightstep/lightstep-tracer-go v0.22.0
-	github.com/nats-io/nats-server/v2 v2.2.6
+	github.com/nats-io/nats-server/v2 v2.2.6 // indirect
 	github.com/nats-io/nats.go v1.11.0
 	github.com/oklog/oklog v0.3.2
 	github.com/oklog/run v1.0.0 // indirect
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,11 @@ jobs:
         env:
           eureka.server.responseCacheUpdateIntervalMs: 1000
 
+      nats:
+        image: nats:2.2.6
+        ports:
+          - 4222
+
     steps:
       - name: Set up Go
         uses: actions/setup-go@v2.1.3
@@ -59,6 +64,7 @@ jobs:
           CONSUL_ADDR: localhost:${{ job.services.consul.ports[8500] }}
           ZK_ADDR: localhost:${{ job.services.zk.ports[2181] }}
           EUREKA_ADDR: http://localhost:${{ job.services.eureka.ports[8761] }}/eureka
+          NATS_URL: nats://localhost:${{ job.services.nats.ports[4222] }}
         run: go test -v -race -coverprofile=coverage.coverprofile -covermode=atomic -tags integration ./...
 
       - name: Upload coverage
--- a/docker-compose-integration.yml
+++ b/docker-compose-integration.yml
@@ -24,3 +24,8 @@ services:
       eureka.server.responseCacheUpdateIntervalMs: 1000
     ports:
       - "8761:8761"
+
+  nats:
+    image: nats:2.2.6
+    ports:
+      - "4222:4222"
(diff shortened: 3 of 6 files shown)


ChrisHines:
What are the implications for someone running the tests locally?

sagikazarmark:
They need docker (similarly to other integration tests).

The test execution itself is a little bit trick (again: similarly to the rest of the integration tests): you need to run a specific command to make it work: `NATS_URL=nats://127.0.0.1:4222 go test -tags ./transport/nats` (earlier it was just working with `go test`)

I think we could improve the developer experience overall for all integration tests (eg. have a Makefile with that integration test command), but I would make that a separate PR.

peterbourgon:
This PR converts what were previously unit tests leveraging an in-memory NATS server, to integration tests requiring Docker and a separate process. I don't quite understand the benefit?

sagikazarmark:
I don't think they were unit tests just because they used an in-memory nats server instance. The benefit is not having to depend on nats server from go-kit and to unify how integration tests are written/executed.

peterbourgon:
The definition of a unit test for me is something which doesn't require external processes/interaction. So these weren't integration tests previously.

sagikazarmark:
I don't think that's a good definition for unit tests. If you take a look at the tests themselves, you'll see that they didn't change. Well, technically a cleanup step disappeared, but the important parts remained intact.

If what you said were true, classifying a test as unit or integration test would rely solely on the execution environment and not what the test is about.

For example, the etcd _integration_ tests could be rewritten to launch an in-memory etcd server. (I'm not suggesting to do it). Would that make the tests unit tests?

What if I changed the `newEtcdConn` function to randomly choose between using an in-memory and a docker instance? How would you classify the tests in that case?

IMHO unit testing is about testing units in isolation, whereas these tests make sure that a specific implementation of an interface would work with another software module (the real thing): nats, etcd, zk, etc, hence I call them integration tests regardless of the execution environment.

I admit that the difference is subtle if you consider the in-process argument (you could argue that isolation is achieved by running an in-memory version of the other software module), but I'd like refer back to my earlier example about not being able to classify a test without knowing the execution environment.

---

There was also an argument about not depending on the nats server module in go-kit code to eliminate it from the dependency graph. Looking at the PR again, I'm not so sure about this one anymore. Apparently the nats-go client library also depends on it.

Hopefully the proposed changes in Go 1.17 (can't find the exact reference, but it's about ignoring transitive test dependencies) will help on this front.

---

The unification argument doesn't make sense if we accept that these were unit tests. I don't think that's true.

---

Feel free to close this PR if you made up your mind and don't want to discuss it any further, no hard feelings.

peterbourgon:
> If what you said were true, classifying a test as unit or integration test would rely solely on the execution environment and not what the test is about.

Yes, that's true. Integration means integration with external things: filesystems, processes, time, etc.

> For example, the etcd integration tests could be rewritten to launch an in-memory etcd server. (I'm not suggesting to do it). Would that make the tests unit tests?

Yep.

> What if I changed the newEtcdConn function to randomly choose between using an in-memory and a docker instance? How would you classify the tests in that case?

Flaky 😉 

I don't run Docker on my laptop, and so I don't run integration tests here, either. That means if one of them breaks, I have to wait for CI to tell me. That sucks. If I have any choice in the matter, I'm going to take a unit test over an integration test 100% of the time.

sagikazarmark:
> Yes, that's true. Integration means integration with external things: filesystems, processes, time, etc.

Integration means integration with other software modules regardless of whether it's an external process or not.

> That means if one of them breaks, I have to wait for CI to tell me. That sucks. 

Yeah, that sucks. :)

peterbourgon:
> Integration means integration with other software modules regardless of whether it's an external process or not.

Agree to disagree.

If the benefit here is standardization, then 👎  from me. If there's another reason, I'm all ears.

sagikazarmark:
The other benefit is one less dependency, but let's wait for Go 1.17 and see if it improves the way test dependencies are added to the graph.

go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/hudl/fargo v1.3.0
 	github.com/influxdata/influxdb1-client v0.0.0-20191209144304-8bf82d3c094d
 	github.com/lightstep/lightstep-tracer-go v0.22.0
-	github.com/nats-io/nats-server/v2 v2.2.6
+	github.com/nats-io/nats-server/v2 v2.2.6 // indirect
```
peterbourgon, line 33: Still a dependency, I think?

sagikazarmark: Could be, but not a direct one at least. This is why I decided to leave that argument and wait if Go 1.17 improves the situation. I can't say I fully understand the relevant proposal yet.

ChrisHines:
> There was also an argument about not depending on the nats server module in go-kit code to eliminate it from the dependency graph. Looking at the PR again, I'm not so sure about this one anymore. Apparently the nats-go client library also depends on it.

FWIW, the nats-go client library depends on the nats-server code for its own tests. I don't think it needs it for other reasons.

go.mod
peterbourgon: Gotcha. Just for the record, no real difference in indirect vs. direct from the "does my module depend on this" perspective.

sagikazarmark: Not yet anyway. That's what Go 1.17 is supposed to address, but if nats server is only used in tests in kit it should probably be okay then.

PR #1104, opened 2021-06-09
Avoid unnecessary write-lock contention in nodeFor

martin-sucha:
Most of the time, we don't need to take the write lock as nodes will be
reused.

The access pattern in node is suited for sync.Map as almost all
accesses are reads.


peterbourgon:
Thanks, but no thank you!

PR #967, opened 2020-02-28
Add golangci-lint

sagikazarmark:
This PR adds [golangci-lint](https://golangci.com/) to the project.

I disabled all linters that produced an error for now, if you like the direction, I'll fix violations as well.

For now, I added a Makefile to make using the linter easy both locally and in CI.

Alternatively, we could use a GitHub action in CI: https://github.com/actions-contrib/golangci-lint

PR #1107, opened 2021-06-16
Add linters

sagikazarmark:
I gathered my strength and made the changes requested in #967 as a separate PR.

Before I go ahead and fix all those lint failures, I'd like to agree on the specific rules that we want to turn on. Right now I left everything on default mode.

Note: golint is deprecated. revive seems to be the alternative.

Fixes #967


Makefile
```diff
@@ -0,0 +1,23 @@
...
+bin/staticcheck-${STATICCHECK_VERSION}:
+	@mkdir -p bin
+	curl -L https://github.com/dominikh/go-tools/releases/download/${STATICCHECK_VERSION}/staticcheck_${OS}_amd64.tar.gz | tar -zOxf - staticcheck > ./bin/staticcheck-${STATICCHECK_VERSION} && chmod +x ./bin/staticcheck-${STATICCHECK_VERSION}
+
+.PHONY: lint
+lint: bin/revive bin/staticcheck
+	go vet ./...
+	bin/revive ./...
+	bin/staticcheck ./...
+	gofmt -l -s -e . | grep .go && exit 1
```
ChrisHines, line 23: `gofmt` formatting can—[and has](https://golang.org/doc/go1.10#gofmt), [more than once](https://golang.org/doc/go1.11#gofmt)—[change](https://golang.org/doc/go1.13#gofmt) between versions of Go, how will we handle that when our builds straddle the versions where a change is made?

ChrisHines (commented):
Should linting be a separate build workflow from running tests?

sagikazarmark:
> Should linting be a separate build workflow from running tests?

Yeah, that's a good idea.

Makefile
sagikazarmark: I used the linters Peter listed in his last comment. I don't mind dropping gofmt.

peterbourgon:
Not really down with `revive`, specifically the notion that it's a framework for your own ruleset — that's explicitly what linters should _not_ be! 😉 `golint` is still as effective as it ever was, let's stick with that. I could also be convinced to switch to [mvdan/gofumpt](http://github.com/mvdan/gofumpt) but it will generate a ton of work on first run, I'm sure.

But what is motivating this PR? Is it a specific need, or general housekeeping? If it's general housekeeping, let's maybe wait until after a release, there's a lot of stuff in flight right now, don't you think?

ChrisHines:
I'm not sure how much we even need golint if we're using staticcheck, I suspect there would be a lot of duplicate checks between the two. But I'm not opposed to golint.

What do either of you think of using https://github.com/reviewdog/reviewdog to automate posting linter results as PR comments? I've used it for several years on work projects and find it reduces work for reviewers while simultaneously providing good feedback for contributors.

sagikazarmark:
I guess the first thing is to settle on the list of tools:

- go vet
- staticcheck
- gofmt -> gofumpt?
- ~~revive~~ -> golint?


> But what is motivating this PR? 

I think originally I submitted it about a year ago because of some "nit: do XY because coding style" kind of reviews. Linters can automate that and reduce turnaround time. Obviously, introducing new linters require some housekeeping.

> What do either of you think of using https://github.com/reviewdog/reviewdog 

I haven't used reviewdog before, but I'm familiar with it. I think it's a good idea, but I think GitHub Actions has a similar feature: it can decode the output (we can register decoders) and annotate specific lines. Not sure if it actually comments on the PR though.

Issue #1109, opened 2021-06-18
Update dependencies


sagikazarmark:
Let's update the dependencies of this project to use latest versions.

I'd prefer finishing #1100 first, because it already changes dependencies considerably.

ChrisHines:
#1105 includes a commit where I made a pass at that. Of note, I got some panics it the metrics/cloudwatch2 tests after upgrading. Also I believe there are breaking API changes in the newest versions of those dependencies someone should take a look at.

sagikazarmark:
Thanks, I'll take a look at that.

We should consider upgrading to the v2 SDK for AWS. It's not a giant module anymore for every service.

ChrisHines:
> We should consider upgrading to the v2 SDK for AWS. It's not a giant module anymore for every service.

Yeah, I noticed that, but didn't want to take on the effort of figuring out how to use the new API. I agree it is probably a good idea for someone to do, though.

sagikazarmark:
I'm gonna take a look at it.

sagikazarmark:
PRs sent. I left the "old" AWS SDK in place for now. I couldn't tell the difference between cloudwatch and cloudwatch2. If the only difference is the AWS SDK version, we could just deprecate the old one.

sagikazarmark:
Dependencies are updated. We can run `go get -u` one more time before tagging the final release. I'm closing this issue for now.

PR #1115, opened 2021-06-22
Update protobuf

sagikazarmark:
This PR updates protobuf (with the relevant generators) and the related generated (and version controlled) code.

The code was generated using the following Nix shell setup:

```nix
{ pkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/432fc2d9a67f92e05438dff5fdc2b39d33f77997.tar.gz") {} }:

pkgs.mkShell {
  buildInputs = [
    pkgs.git
    pkgs.go_1_16
    pkgs.protobuf
    pkgs.protoc-gen-go
    pkgs.protoc-gen-go-grpc
  ];
}
```

Related to #1109


ChrisHines (approved):
I'm not a gRPC expert, so I can't really validate all of the changes here. I do find it fascinating that `go mod tidy` doesn't make any changes to `go.sum` in this case, but that reproduces for me locally. Looks good as far as I can tell, though.

sagikazarmark:
Most of the changes are in the code generator, but the new proto library should be backwards compatible with the old one. The rest of the changes only affect tests, not the library itself.

I guess the reason for no go.sum changes is that there are no observed changes in the list of dependencies: the new proto lib  is a dependency of the old one (or vice versa?) and probably something else relies on the old library as well, so it doesn't disappear from `go.sum`. But I agree, the odds for this to happen are pretty low.

PR #1117, opened 2021-06-22
Replace aws lambda structs with stubs

sagikazarmark:
This PR removes the aws-lambda module dependency by replacing some structs with internal stubs.

Since the actual data only travels as JSON, it doesn't make much sense to use the original structs IMO. One less dependency...

Related #1109

### DIFF
--- a/transport/awslambda/handler_test.go
+++ b/transport/awslambda/handler_test.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/aws/aws-lambda-go/events"
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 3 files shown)


transport/awslambda/handler_test.go
```diff
@@ -21,6 +20,17 @@ const (
 	KeyEncMode   key = iota
 )
 
+// Created based on github.com/aws/aws-lambda-go/events.APIGatewayProxyRequest for the purposes of the tests below.
```
ChrisHines, line 23: Consider including the version of the github.com/aws/aws-lambda-go/events package referred to when creating these structs.

sagikazarmark: Good point, done.

ChrisHines (approved):

PR #1119, opened 2021-06-22
Dependency updates

sagikazarmark:
This PR updates various "smaller" dependencies that doesn't need a separate PR.

Related #1109


go.mod
```diff
@@ -22,28 +22,28 @@ require (
 	github.com/go-stack/stack v1.8.0
 	github.com/golang/protobuf v1.5.2
 	github.com/google/btree v1.0.0 // indirect
-	github.com/hashicorp/consul/api v1.3.0
+	github.com/hashicorp/consul/api v1.8.1
```
ChrisHines, line 25: Looks like v1.9.0 just came out.

sagikazarmark: Thanks! Updated

sagikazarmark: Turns out that 1.9.0 requires Go 1.16 (it uses `io.Discard` [added in that version](https://golang.org/doc/go1.16#ioutil)), so I reverted the update for now. If a decision is made that making Go kit 1.16 only is fine, we can upgrade it, but I'd probably wait for Go 1.17 so we can support at least two versions.

ChrisHines: OK, good to know. Agreed on maintaining 1.15 compatibility for now. Last I checked the go-kit policy is to support the current and previous Go release.

ChrisHines (approved):

PR #1124, opened 2021-06-28
Update dependencies

sagikazarmark:
This PR updates some more dependencies (released since the last one or ones that I missed earlier) before the 0.11.0 release.


peterbourgon:
How do we keep "update dependencies" PRs to an absolute minimum? I really really do not want to field an infinite series of "update deps" PRs until my death 😉 As part of the release seems OK, can we stick to that policy?

sagikazarmark:
I think that's fair. AWS SDK for example is released almost every day, we probably shouldn't update it every day. 😉 

We could automate dependency updates using dependabot: it automatically updates dependencies in the form of PRs.

One thing that would help reducing dependency updates is reducing dependencies. 🙂 Particularly the `sd` and `metrics` packages are quite heavy.

But until then, updating dependencies before releases is fine IMO.

peterbourgon (approved):

PR #1125, opened 2021-06-30
Add a log adapter for zerolog.Logger

ozon2:
I am currently using zerolog, and some projects I use are using kitlog, so I made this adapter to make kitlog compatible with zerolog.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/openzipkin/zipkin-go v0.2.5
 	github.com/performancecopilot/speed v3.0.0+incompatible
 	github.com/prometheus/client_golang v1.11.0
+	github.com/rs/zerolog v1.23.0
 	github.com/sirupsen/logrus v1.8.1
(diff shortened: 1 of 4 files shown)


sagikazarmark:
@ozon2 The log package has been extracted to a separate module: https://github.com/go-kit/log

The adapter packages has been dropped. There is no need for them to be in the core. I'd suggest releasing it under your name. We can add a link to the readme pointing to your repository.

ozon2:
@sagikazarmark Thanks for the suggestion. Probably won't do it though, I find it a little overkill to release a single file. Anyway if someone searches for an adapter they might stumble upon this PR.

peterbourgon:
Feel free to comment with a link to a gist with the adapter code! IMO this is the perfect kind of thing to copy/paste.

ozon2:
@peterbourgon Good idea!
https://gist.github.com/ozon2/b2150944976da8a9f2cf1a2a1e8c2047

PR #1128, opened 2021-07-02
fix a broken link to the addsvc example

robbert229:
The documentation included a stale link to where the addsvc example used to live from before the examples were extracted into their own git repo.


sagikazarmark (approved):
Thanks @robbert229 !

Issue #1127, opened 2021-06-30
Abandoned packages


sagikazarmark:
There are a couple packages that seem to be abandoned and outdated (ie. dependencies are old, no go.mod, etc).

These are:

- `github.com/performancecopilot/speed` (see performancecopilot/speed#59)
- `github.com/hudl/fargo` (see hudl/fargo#74)

If there is no answer, I propose we fork them, make the necessary updates and use the forks as drop-in replacements. Before we do that, though, we should consider whether we want to keep these in the core in the first place. It might not worth the hassle if they get moved out of the core. So this issue is a conversation starter.

Related #843

peterbourgon:
I'd much rather drop them altogether.

sagikazarmark:
Works for me, although I already got them both to update to modules and release versions.

bnevis-i:
Additionally, aws-sdk-go@v1.38.68 is introducing transitive vulnerabilities into this package (CVE-2020-9283).  Bringing the whole go.mod up to date in general would be a good idea.

peterbourgon:
We did that like two weeks ago. Ugh. What a waste of time.

bnevis-i:
Actually, I take that back.  aws-sdk doesn't have direct dependency on the broken library. Egg on my face :-(

peterbourgon:
Not at all, best news I've heard today.

sagikazarmark:
@peterbourgon I'd suggest updating these two packages and at the same time deprecating them in the next release.

Are there any other packages that we should deprecate? Should we start planning for #843?

sagikazarmark:
Closing as both libraries have been updated.

PR #1167, opened 2021-07-20
Update fargo library

sagikazarmark:
See #1127

I think it's still worth updating, even if we decide to deprecate it.


peterbourgon:
I understood we'd be doing updates just prior to releases?

sagikazarmark:
Yep, I mostly wanted to use this PR as a reference for the discussion about deprecating/dropping the eureka. The change is trivial, no breaking changes in the library, so I'd say it's worth upgrading along with a deprecation notice.

I don't mind closing this PR and updating dependencies before the next release.

peterbourgon:
OK, as long as it's a one-off :) Just want to reiterate that we shouldn't be processing an unending stream of "update so-and-so dependency" PRs.

peterbourgon (approved):

sagikazarmark:
Let's repurpose this PR then into deprecating eureka (joined with a dependency update). But that would require having a conversation/making a decision about what and how to deprecate.

Since generics are expected to be available in 1.18 and #843 proposes a large refactor once generics are available, maybe we can start planning that. But as @ChrisHines pointed out earlier: making large changes in smaller chunks is better, so if the introduction of generics is targeted for 1.18, maybe the next Go kit version should be released somewhere after Go 1.17, deprecating the packages that we intend to remove/relocate as part of the larger refactor. That would give people enough time to start migrating away (given these packages actively used).

Issue #1026, opened 2020-10-27
High severity issue in github.com/dgrijalva/jwt-go


sagikazarmark:
There is an unpatched, high severity issue in the aforementioned JWT package: https://github.com/dgrijalva/jwt-go/issues/428

Unfortunately, it looks like the author completely abandoned the package.

There is a maintained fork that fixes the issue, but we decided to switch to gopkg.in/square/go-jose.v2

A lot of users seem to prefer this package now.

I might provide a PR later, I just wanted to raise awareness that the current JWT implementation relies on a library with a security risk.

It's not a security issue with the library itself and can easily be fixed by users on their end by using a fork, so public disclosure should probably be ok. Feel free to delete the issue if you disagree @peterbourgon

JimFicarra:
If you're referring to the fork at [github.com/form3tech-oss/jwt-go](https://github.com/form3tech-oss/jwt-go), there's a bug in the code that doesn't properly handle the audience claim.  The [RFC ](https://tools.ietf.org/html/rfc7519#section-4.1.3)specifies the AUD claim can either be a string or an array of strings.  

The code in the fork assumes it's an array of string and any AUD claim with a string (e.g. Azure AD id tokens) will cause a panic.

sagikazarmark:
Thanks for pointing that out, I didn't know that. I guess we are better off with go-jose then.

peterbourgon:
Happy to take a PR which migrates to the other package.

nathanejohnson:
https://github.com/dgrijalva/jwt-go/issues/422

If you look at the bottom, it appears to be fixed in the v4 version, which has been tagged "preview" for a bit but that's what we're using internally for what it's worth.

aldas:
This vulnerability is real problem when you are reusing token signing keys for different applications and have coded explicit rule that `aud` is optional. If `aud` is checked as required field then there is no vulnerability, only a bug (claims are considered invalid).  See https://github.com/dgrijalva/jwt-go/issues/422#issuecomment-749976730

But does it even make sense in security perspective to have a rule - it is **OK** when `aud` field is missing from token and if `aud` is in token we will check if it matches with our value. Even without this vulnerability you have a problem - If attacker gets their hands on **VALID** token without `aud` this check passes.

peterbourgon:
Is there an outcome here? What should we be doing?

JimFicarra:
[...]

PR #1172, opened 2021-07-27
Switch to github.com/golang-jwt/jwt

vovinacci:
Closes https://github.com/go-kit/kit/issues/1026

This PR switches to `github.com/golang-jwt/jwt` to address security vulnerability. More details in the issue and https://github.com/advisories/GHSA-w73w-5m7g-f7qc


peterbourgon:
@sagikazarmark Look good to you?

sagikazarmark (approved):

sagikazarmark:
After taking a more thorough look at the implementation, I'm not _sure_ that there is a breaking change, but I _suspect_ there might be.

Basically, those different types referenced in the public API are all interfaces and I assume the new package didn't break those overnight, so in _theory_ the replaced types could be backwards compatible with the old library.

In practice, though, some of the types used internally (eg. standard claims struct) are not interfaces, so those could break.

A simple test to check if there is a breaking change would be (temporarily) changing back the packages in tests to the old module to see if it works with the new code.

I'm not particularly worried about a breaking change (the library is basically abandoned after all), but if there is a breaking change, it should be released in a minor version, not a patch. Go 1.17 is just around the corner and Chris wants to cut a new release after that (hoping that we can further reduce the amount of dependency downloads).

peterbourgon:
OK let's bench this until the next release.

vovinacci:
@peterbourgon Do you have any preliminary release date in mind?

AliLogic:
Any updates on this?

go.mod
```diff
@@ -12,13 +12,13 @@ require (
...
 	github.com/clbanning/x2j v0.0.0-20191024224557-825249438eec // indirect
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/edsrzf/mmap-go v1.0.0 // indirect
 	github.com/franela/goblin v0.0.0-20200105215937-c9ffbefa60db // indirect
 	github.com/franela/goreq v0.0.0-20171204163338-bcd34c9993f8 // indirect
 	github.com/go-logfmt/logfmt v0.5.0
 	github.com/go-stack/stack v1.8.0
 	github.com/go-zookeeper/zk v1.0.2
+	github.com/golang-jwt/jwt v3.2.1+incompatible
```
sagikazarmark, line 21: Looks like v4 is released with proper Go module support

sagikazarmark (changes requested):
Can you please upgrade the library to v4?

vovinacci:
@sagikazarmark Done

shogo82148:
I recommend to upgrade to Go 1.17 pruned module graphs.

https://golang.org/doc/go1.17#go-command

```
go mod tidy -go=1.17
```

It enables some optimizations, such as [Lazy module loading](https://golang.org/ref/mod#lazy-loading).

eli-darkly:
@shogo82148 If that would require setting the minimum version to 1.17 in go.mod, I would very strongly prefer that they _not_ make that change, at least not yet. Go 1.17 was just released and we can't all immediately update our projects to require it.

sagikazarmark:
> I recommend to upgrade to Go 1.17 pruned module graphs.

This is already in progress on another branch. Not sure which will be merged first yet.

sagikazarmark (approved):

andig:
> @shogo82148 If that would require setting the minimum version to 1.17 in go.mod, I would very strongly prefer that they _not_ make that change, at least not yet. Go 1.17 was just released and we can't all immediately update our projects to require it.

Imho this does not raise the minimum version.

eli-darkly:
@andig:
> Imho this does not raise the minimum version.

Could you say more about why you think it wouldn't? My reading of the linked documentation is that pruned module graphs are only a thing in `go.mod` files that specify `go 1.17`.

shogo82148:
pruned module graphs are just `required` directives.
So, Go 1.16 or earlier parses them as extra dependencies.
This has no negative effects, because these extra dependencies are also in `go.sum` and Go 1.16 loads whole dependencies in `go.sum`.

The `go 1.17` directive enables new language features, and changes the `go` command behavior. https://golang.org/ref/mod#go-mod-file-go
However it doesn't break backward compatibility.

When Go 2.0 comes out in the future, Go 2.0 will not be able to import modules that contain the `go 1.17` directive.

PR #1187, opened 2021-09-28
Use errors.Is for error handling

alexander-melentyev:


transport/http/jsonrpc/client_test.go
```diff
@@ -170,7 +171,7 @@ func TestClientHappyPath(t *testing.T) {
 		}
 
 		b, err := ioutil.ReadAll(r.Body)
-		if err != nil && err != io.EOF {
```
sagikazarmark, line 173: See https://mobile.twitter.com/mvdan_/status/1442492461025222661

metrics/cloudwatch/cloudwatch.go
```diff
@@ -271,10 +271,6 @@ func sum(a []float64) float64 {
 	return v
 }
 
-func last(a []float64) float64 {
-	return a[len(a)-1]
-}
-
```
sagikazarmark, line 277: This should probably be a separate commit/PR.

transport/http/jsonrpc/client_test.go
alexander-melentyev: Added back

metrics/cloudwatch/cloudwatch.go
alexander-melentyev: Done

sagikazarmark (approved):

peterbourgon:
I guess we should stick sd/eureka on the deprecation list...

alexander-melentyev:
Again...

PR #1191, opened 2021-10-01
Add golangci-lint

alexander-melentyev:
Added golangci-lint instead of deprecated metalinter


sagikazarmark:
There is already a conversation about adding linters in #1107

PR #1193, opened 2021-10-01
Refactor cmp

sashamelentyev:


alexander-melentyev:
Data race again...
@sagikazarmark, @peterbourgon, please check

sagikazarmark:
Eureka tests sometimes fail...really, we should just get rid of it already.

peterbourgon:
@alexander-melentyev What problem does this solve?

peterbourgon:
I think we should probably just deprecate or remove sd/eureka altogether, it's not really used any more, as far as I know.

alexander-melentyev:
Nothing. 
Code Is shorter

alexander-melentyev:
@peterbourgon I close this?

peterbourgon:
Yeah probably best.

PR #1194, opened 2021-10-01
Remove eureka

sashamelentyev:


alexander-melentyev:
@peterbourgon I close this?

peterbourgon:
In the abstract, I want to remove eureka, but I don't think just deleting the package is the way to go. I think we will break out `sd/*` to their own repos at some point, and then mark eureka as archived.

PR #1196, opened 2021-10-11
Remove empty func description

sashamelentyev:


peterbourgon:
This is not an empty func description.

PR #1202, opened 2021-11-21
Refactor metrics/cloudwatch

sina-devel:
- Export `option` type for better documentation
- Add documentation for exported option functions
- ~~Use `CloudWatch.apply` method and remove unused `Percentiles` type~~
- Remove unused `CloudWatch.apply` method and `Percentiles` type

### DIFF
--- a/metrics/cloudwatch/cloudwatch.go
+++ b/metrics/cloudwatch/cloudwatch.go
@@ -23,11 +23,6 @@ const (
 	maxValuesInABatch     = 150
 )
 
-type Percentiles []struct {
-	s string
-	f float64
-}
-
 // CloudWatch receives metrics observations and forwards them to CloudWatch.
 // Create a CloudWatch object, use it to create metrics, and pass those metrics as
 // dependencies to the components that will use them.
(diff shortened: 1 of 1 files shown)


metrics/cloudwatch/cloudwatch.go
```diff
@@ -102,8 +104,8 @@ func New(namespace string, svc cloudwatchiface.CloudWatchAPI, options ...option)
 		percentiles:           []float64{0.50, 0.90, 0.95, 0.99},
 	}
 
-	for _, optFunc := range options {
-		optFunc(cw)
+	for _, opt := range options {
+		cw.apply(opt)
 	}
```
peterbourgon, line 109: Can we just remove the apply method altogether?

sagikazarmark: It's actually more safer than a simple function. `apply` cannot be called outside of the package, but with the current `Option` type it's possible to modify the state of `Cloudwatch` after the initialization.

I'd instead change the signature of `Option` to this:

```go
type Option interface{
    apply(*CloudWatch)
}
```

peterbourgon: > I'd instead change the signature of Option to this:

Strong -1. Option interfaces are an antipattern, and especially those that take this form, i.e. an exported interface with an unexported method.

> with the current Option type it's possible to modify the state of Cloudwatch after the initialization.

I understand your point, but I don't think this is a meaningful risk.

sina-devel: I think we can remove the `apply` method, because `cloudwatch.New(namespace, svc, nil)` isn't norm.

sina-devel: fixed

PR #1207, opened 2022-01-02
refactor(all): Remove unused alias and args

sashamelentyev:


peterbourgon:
Thanks, but no thanks.

PR #1212, opened 2022-01-21
feat(http): make interceptingWriter reimplement common interfaces

### LINKED ISSUE #1092: Support for WebSockets in Server interface

Reasno:
successor to #1093 and closes #1092.

### DIFF
--- a/transport/http/intercepting_writer.go
+++ b/transport/http/intercepting_writer.go
@@ -1,0 +1,257 @@
+package http
+
+import (
+	"io"
+	"net/http"
+)
+
+type interceptingWriter struct {
+	http.ResponseWriter
+	code    int
+	written int64
+}
+
+// WriteHeader may not be explicitly called, so care must be taken to
+// initialize w.code to its default value of http.StatusOK.
+func (w *interceptingWriter) WriteHeader(code int) {
+	w.code = code
+	w.ResponseWriter.WriteHeader(code)
+}
+
+func (w *interceptingWriter) Write(p []byte) (int, error) {
+	n, err := w.ResponseWriter.Write(p)
+	w.written += int64(n)
+	return n, err
+}
+
+// reimplementInterfaces returns a wrapped version of the embedded ResponseWriter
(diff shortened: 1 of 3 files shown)


peterbourgon:
Sure! Tests, please? 😇

Reasno:
> Sure! Tests, please? 😇

Test already included in the commit. The diff is collapsed by GitHub for being too large.

peterbourgon:
Oops! You're right! Sorry!

transport/http/intercepting_writer.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// reimplementInterfaces returns a wrapped version of the embedded ResponseWriter
+// and selectively implements the same combination of additional interfaces as
+// the wrapped one. The interfaces it may implement are: http.Hijacker,
+// http.CloseNotifier, http.Pusher, http.Flusher and io.ReaderFrom. The standard
+// library is known to assert the existence of these interfaces and behaves
+// differently. This implementation is derived from
+// https://github.com/felixge/httpsnoop.
+func (w *interceptingWriter) reimplementInterfaces() http.ResponseWriter {
```
peterbourgon, line 34: Can we make this into a constructor? e.g.

```go
func newInterceptingWriter(w http.ResponseWriter) http.ResponseWriter {
    ...
}
```

Issue #1214, opened 2022-02-13
Consul Implementation of Instancer spams Consul and pushes consul agents to 100% CPU


jkratz55:
### What did you do?

Using the GoKit implementation of Instancer for Consul in GoKit 0.11 and up results in Instancer spamming Consul with hundreds of RPC requests/sec on any register/deregister event that occurs after the Instancer is created. This resulted in our Consul agent pods going to 100% CPU and overwhelming our Consul cluster.

Steps to re-create:

1. Start Consul

2. Register a dummy service

`curl --location --request PUT 'http://localhost:8500/v1/agent/service/register' \
--header 'Content-Type: application/json' \
--data-raw '{
    "name": "db-example",
    "id": "db-example",
    "address": "127.0.0.1",
    "port": 5432,
    "tags": [
        "database"
    ]
}'`

3. Open a terminal and run the following command to get insights into what the Consul agent is doing

`consul monitor -log-level=trace`

4. Create Instancer for db-example service (see sample code below) and run app/code

5. Deregister the service

`curl --location --request PUT 'http://localhost:8500/v1/agent/service/deregister/db-example'`

Observe the behavior from consul monitor, the agent is getting slammed with requests and it never stops until the Instancer is stopped or application exits.

Simple example application.

```
package main

import (
	"fmt"
	"github.com/go-kit/kit/log"
	"github.com/go-kit/kit/sd"
	"github.com/hashicorp/consul/api"
	"time"

	"github.com/go-kit/kit/sd/consul"
)

func main() {

	apiClient, err := api.NewClient(api.DefaultConfig())
	if err != nil {
		panic(err)
	}
	client := consul.NewClient(apiClient)

	instancer := consul.NewInstancer(client, log.NewNopLogger(), "db-example", []string{}, true)

	events := make(chan sd.Event, 1)
	go func() {
		instancer.Register(events)
	}()

	go func() {
		for event := range events {
			fmt.Println(event.Instances)
		}
	}()

	time.Sleep(1 * time.Hour)
}
```

### What did you expect?

Consul agent should not be getting spammed with hundreds of requests a second.

### What happened instead?

Instancer spammed Consul with way too many requests and drove the CPU utilization to 100%.

`7.0.0.1:53702 latency=9.375µs
2022-02-12T18:06:20.710-0500 [DEBUG] agent.http: Request finished: method=GET url=/v1/health/service/db-example?index=155&passing=1 from=127.0.0.1:53702 latency=12.208µs
[...]

PR #1215, opened 2022-02-13
Fixed instancer loop logic that was causing Consul to get spammed

jkratz55:
Address issue #1214 - Instancer spamming Consul with requests

The instancer loop was never updating the lastIndex but passing it to getInstances which caused Instancer to spam Consul repeatedly and nearly killed our Consul agent pods and time instances were added or removed from Consul for that service.

### DIFF
--- a/sd/consul/instancer.go
+++ b/sd/consul/instancer.go
@@ -65,9 +65,9 @@ func (s *Instancer) loop(lastIndex uint64) {
 		instances []string
 		err       error
 		d         time.Duration = 10 * time.Millisecond
+		index     uint64
 	)
 	for {
-		index := lastIndex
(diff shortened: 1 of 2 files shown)


peterbourgon:
Cool! Can you add a test which fails on the previous version and passes on this branch?

jkratz55:
Hey @peterbourgon, I added a test case to the PR which fails on the old version and passes with the fix. Since the logic was in a loop that ran forever, the unit tests seem odd but seem to work. I'm just pumping out an error on a channel when the mock client detects the same index value was passed twice in a row and assumes every lookup Consul returns a new index since it isn't doing blocking queries in the mock.

peterbourgon:
Thanks!

PR #1232, opened 2022-06-02
chore: update all dependencies, and add dependabot

sashamelentyev:


peterbourgon:
Thanks, but I don't want to add Dependabot.

peterbourgon:
IMO there's no reason to update deps without specific reason.

Issue #1243, opened 2022-08-02
Upgrade to prometheus/client_golang 1.12.2


rootulp:
### What would you like?

https://github.com/go-kit/kit/blob/50c38262a8c71be2532f45100e622025bc5ffedf/go.mod#L23

I'd like to leverage new features in prometheus/client_golang [1.12.2](https://github.com/prometheus/client_golang/releases/tag/v1.12.2). Are there any objections to upgrading this dependency?

ChrisHines:
How do you want to leverage the new features? Do you want to use them in Go kit code or your own code?

rootulp:
In a consuming application ([celestiaorg/celestia-core](https://github.com/celestiaorg/celestia-core)) I wanted to use `counter.Inc()` which I thought was only available in 1.12.2 ([here](https://github.com/prometheus/client_golang/blob/v1.12.2/prometheus/counter.go#L40)) but it also appears available in 1.11.1 ([here](https://github.com/prometheus/client_golang/blob/v1.11.1/prometheus/counter.go#L40)). So I investigated this codebase a little more and learned that my feature request would imply adding `Inc()` to 
https://github.com/go-kit/kit/blob/50c38262a8c71be2532f45100e622025bc5ffedf/metrics/metrics.go#L5-L8

and this method may not be applicable to counters for other packages in https://github.com/go-kit/kit/tree/50c38262a8c71be2532f45100e622025bc5ffedf/metrics

Do you have thoughts here? It's likely that this issue can be closed as not planned

peterbourgon:
Feel free to add a PR that adds Inc to Counter, but it's not obvious to me that 

```go
c.Inc()  // is better than
c.Add(1) // in any meaningful sense
```

peterbourgon:
Said another way,

> I'd like to leverage new features in prometheus/client_golang [1.12.2](https://github.com/prometheus/client_golang/releases/tag/v1.12.2). Are there any objections to upgrading this dependency?

You don't need to upgrade the dependency to use the new features. So I guess I'll close this for now.

PR #1244, opened 2022-08-02
Upgrade to prometheus/client_golang 1.12.2

rootulp:
Resolves https://github.com/go-kit/kit/issues/1243


peterbourgon:
Your downstream application is not prevented from using this version of client_golang if this patch is not applied. So I don't see the rationale; am I missing something?

Issue #1081, opened 2021-04-22
add  GOFIBER (fiber) the  best performant web framework to kit


gitsupi:

peterbourgon:
No.

yzia2000:
Oh just curious, are there any issues fiber as to why it is not added here?

peterbourgon:
Two main reasons. 

Fiber uses fasthttp, which doesn't implement HTTP/2 at all, provides an incomplete implementation of HTTP/1, and (IMO) has very low code and documentation quality. The fasthttp package is a niche optimization for specialized use cases, and inappropriate as a default choice.

Two, Fiber is a transliteration of a Javascript framework, and values a "Javascript-like" developer experience very highly. That preference seems to outweigh following Go language idioms and best practices. The most common examples that come to mind are misuse of [the empty interface](https://pkg.go.dev/github.com/gofiber/fiber/v2#App.Use) to simulate a dynamic type system, and misuse of [variadic params](https://pkg.go.dev/github.com/gofiber/fiber/v2#New) to express optional semantics. Fiber encourages bad habits.

yzia2000:
Thanks for the comprehensive explanation!

PR #1246, opened 2022-08-04
Suggestion of adding fiber.

KyleWiering:
If it is harmful - would be good to notate that here.


peterbourgon:
https://github.com/go-kit/kit/issues/1081#issuecomment-1131736350

Issue #1249, opened 2022-08-25
Security vuln CVE-2019-19794 with transitive dependency github.com/miekg/dns@v1.0.14


francogeller:
### What would you like?

Latest **version 0.12.0** have a transitive dependency with **github.com/miekg/dns@v1.0.14** because it is a dependency of the following direct dependencies:
- github.com/hashicorp/serf@v0.9.5
- github.com/hashicorp/consul/api@v1.10.1

Any dependency with **github.com/miekg/dns@v1.0.14** must be fully deprecated due to known vulnerability [CVE-2019-19794](https://github.com/advisories/GHSA-44r7-7p62-q3fr) with this module version.

To definitely deprecate **github.com/miekg/dns@v1.0.14** we must:
- Update github.com/hashicorp/serf@v0.9.5 up to to _github.com/hashicorp/serf@v0.10.0_ (current latest)
- Update github.com/hashicorp/consul/api@v1.10.1 up to _github.com/hashicorp/consul/api@v1.14.0_ (current latest)

```bash
$ go get -u github.com/hashicorp/serf@latest
$ go get -u github.com/hashicorp/consul/api@latest
$ go mod tidy
```

After this:
All transitive dependencies with **github.com/miekg/dns@v1.0.14** are completely removed.
All tests continues to complete successfully.

francogeller:
PR #1250 is a proposal to fix the vulnerability

PR #1250, opened 2022-08-25
Update github.com/hashicorp/serf and github.com/hashicorp/consul/api to fix CVE-2019-19794

francogeller:
This PR update github.com/hashicorp/serf and github.com/hashicorp/consul/api in order to fully deprecate github.com/miekg/dns@v1.0.14 due to [CVE-2019-19794](https://github.com/advisories/GHSA-44r7-7p62-q3fr) security vuln.
Issue #1249


peterbourgon:
For the record, version bumps like this aren't necessary in intermediary Go projects, because the version is determined by the downstream consumer. I'm happy to merge simple PRs like this one, but there is no security issue here, really.

francogeller:
Hello, thank you very much for considering the PR. Particularly in our projects, the change would be convenient because the applications we have that search for code vulnerabilities mark kit as a vulnerable dependency because it has a transitive dependency with a vulnerable version of dns. 
Maybe this will help others to comply with their security policies, if they have them. But I agree with you that it is not a security vulnerability, instead we can take it as an improvement with integrations.

peterbourgon:
> the change would be convenient because the applications we have that search for code vulnerabilities mark kit as a vulnerable dependency because it has a transitive dependency with a vulnerable version of dns.

Any tool which does this is incorrect 🤷

ChrisHines:
It's also worth noting that changes like this don't propagate to a wide audience until Go kit tags a new release. Someone would have to run `go get github.com/go-kit/kit/...@master` or something similar. The `...@latest` version will continue to get the newest tagged release, as will a basic `go get` when adding Go kit to a go.mod for the first time.

PR #1256, opened 2023-03-01
Remove dead link to code generator from README

heymatthew:
When browsing the code genorators listed in the README, it looks like one is a dead link. I had a search around Leigg's repos and it looks like he's just deleted it.

I suggest removing the dead link to https://github.com/chaseSpace/kit.


heymatthew:
💖

PR #1260, opened 2023-04-30
go.mod: Upgrade go min version to 1.18

lvjp:
This will permit to use generics in future merge requests


peterbourgon:
Can we actually just make this 1.20 and 1.19?

alwismt:
What about upgrade to least 1.19

peterbourgon:
Why do we think these failures are a function of the Go version?

PR #1261, opened 2023-05-23
chore: update golang-jwt/jwt and other modules

andreaciri:
This PR updates all go modules to the latest version.

The main changes are related to golang-jwt/jwt v5, for which I followed the [migration guide](https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md#migration-guide-v500).

The only breaking change I'm introducing in go kit is the rename of `StandardClaimsFactory()` into `RegisteredClaimsFactory()` in order to keep consistency with the underlying naming.


auth/jwt/middleware.go
```diff
@@ -79,10 +79,10 @@ func MapClaimsFactory() jwt.Claims {
...
 }
 
-// StandardClaimsFactory is a ClaimsFactory that returns
-// an empty jwt.StandardClaims.
-func StandardClaimsFactory() jwt.Claims {
-	return &jwt.StandardClaims{}
+// RegisteredClaimsFactory is a ClaimsFactory that returns
+// an empty jwt.RegisteredClaims.
+func RegisteredClaimsFactory() jwt.Claims {
+	return &jwt.RegisteredClaims{}
```
andreaciri, line 85: In case naming consistency is not so critical, we can consider to keep `StandardClaimsFactory` so to avoid a breaking change in go kit.
From my understanding `jwt.RegisteredClaims{}` is mostly backwards compatible with `jwt.StandardClaims{}`.

Any feedback is appreciated

peterbourgon:
A major version bump signals backwards-incompatible changes, essentially an entirely different module than the previous major version. If you want to use jwt/v5 then I'd like to see it either as a new Go kit "jwt2" package, or a pretty comprehensive test suite to ensure that "mostly backwards compatible" actually means "fully backwards compatible" in terms of the API that Go kit offers to _its_ consumers.

auth/jwt/middleware.go
peterbourgon: Changing public API identifiers like this should probably be avoided if at all possible.

fabiorubim740:
> This PR updates all go modules to the latest version.
> 
> The main changes are related to golang-jwt/jwt v5, for which I followed the [migration guide](https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md#migration-guide-v500).
> 
> The only breaking change I'm introducing in go kit is the rename of `StandardClaimsFactory()` into `RegisteredClaimsFactory()` in order to keep consistency with the underlying naming.

Issue #1253, opened 2022-12-08
Fix security vulnerability CVE-2022-41717 exposed by Go gRPC dependency


DonSealy:
### What would you like?

It appears that kit/transport/grpc is vulnerable to the attack outlined in [CVE-2022-41717](https://github.com/golang/go/issues/56350) due to the package dependency on the Go gRPC implementation. Can go-kit be updated to leverage fixes for this vulnerability? It is fixed in Go minor releases 1.18.9 and 1.19.4.

peterbourgon:
Go kit expresses a dependency on the module google.golang.org/grpc at version v1.40.0. The latest version of that module is currently v1.51.0. These two versions share a major version number, and so Go modules treats them as compatible during version resolution, and applications that (correctly) assert that later version will receive it.

Happy to review and approve a PR that bumps the grpc module version as suggested, assuming tests pass and etc. But that's a nice-to-have, if it doesn't happen it's not any kind of vulnerability. See e.g. #1250.

PR #1264, opened 2023-06-08
update google/grpc to 1.55

DonSealy:
Bumps minor version of google.golang.org/grpc from v1.40 to v1.55 to prevent code security scanners from picking up [CVE-2022-41717](https://github.com/advisories/GHSA-xrjj-mj9h-534m), which is a transitive dependency and fixed in v1.52 and later.

A few tests needed minor adjustments, as [v1.52.0 introduced a change](https://github.com/grpc/grpc-go/releases/tag/v1.52.0) that prevents the grpc.Dial() method from receiving an empty target string. I also addressed a deprecated method being used (WithInsecure) and updated it so it won't break later.


DonSealy:
Related issue: [1253](https://github.com/go-kit/kit/issues/1253)

PR #1281, opened 2023-12-17
feat(jsonrpc): add method to retrieve request ID from context

longit644:
This  enhancement introduces a method to retrieve the request ID from the context. This is particularly useful for customizing the `ErrorEncoder` for the jsonrpc transport.


longit644:
Hi @peterbourgon, could you please take a look at this? Thank you.

PR #1294, opened 2024-07-10
Fixed EncodeJSONResponse body writing 

### LINKED ISSUE #1291: http.EncodeJSONResponse body writing is not compatible with net/http
### What did you do?

1. Implemented http.StatusCoder() which returns 304 (Not Modified) 
2. Observed an error when http.EncodeJSONResponse is called: `http: request method or response status code does not allow body`

It seems that body is not written only on 204: https://github.com/go-kit/kit/blob/master/transport/http/server.go#L177

However, stdlib defines "no body allowed" as 100-199, 204, 304: https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/net/http/transfer.go;l=460

### What did you expect?

No error. 

### What happened instead?

Got an error.

kerma:
EncodeJSONResponse is not writing body for status codes 100-199, 204, 304 according to https://datatracker.ietf.org/doc/html/rfc7230#section-3.3

Fixes #1291

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -159,7 +159,8 @@ func NopRequestDecoder(ctx context.Context, r *http.Request) (interface{}, error
 // JSON object to the ResponseWriter. Many JSON-over-HTTP services can use it as
 // a sensible default. If the response implements Headerer, the provided headers
 // will be applied to the response. If the response implements StatusCoder, the
-// provided StatusCode will be used instead of 200.
+// provided StatusCode will be used instead of 200. If the StatusCode is between 100-199,
+// or equal to 204 or 304, the response body will not be written.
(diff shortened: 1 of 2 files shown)


peterbourgon:
Eh, that RFC is more of a suggestion than a requirement :)

kerma:
@peterbourgon I'm a bit confused. I reported #1291 a while ago. In the response there you hinted that this should be fixed, as the EncodeJSONResponse is broken when the user implements a response with a 304 StatusCode.

Should I include just the 304 for that fix, or why you closed this one exactly?

peterbourgon:
Huh, me too apparently.

kerma:
@peterbourgon anything for me to do here? I guess the CI / Build issues are unrelated to this PR.

PR #1295, opened 2024-07-19
Add test for cloudwatch2 counter that demonstrates

bernata:
potential slice bug with label-values.
Label-values are stored in a Counter as a string slice and copyied over to a new Counter when With(lvs ...string) is invoked. The slice bug can present when the same Counter is copied twice with With() and each invocation does not resize the capacity of the slice.

### DIFF
--- a/metrics/cloudwatch2/cloudwatch2_test.go
+++ b/metrics/cloudwatch2/cloudwatch2_test.go
@@ -2,6 +2,7 @@ package cloudwatch2
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


bernata:
Hi @peterbourgon - thank you for contributing gokit; and all your contributions on advocating gokit and its surroundings [go, elegant design]. 
I wanted to get your thoughts on this test which I believe demonstrates a slice bug in some of the metrics Counter structures when the slice is not resized and two Counters refer to the same backing memory. Always using a new slice i think is the most straightforward fix. When you have a moment, it would be great to get your thoughts on this. Thank you again.

PR #1297, opened 2024-08-26
Bump libs

romanglovahs:


peterbourgon:
Thanks, but no reason to do this! 🙇

romanglovahs:
image
@peterbourgon that is for security reason please check attachment

peterbourgon:
I understand the security stuff, but there's no reason for Go kit itself to update its `go.mod` or whatever for this reason.

PR #1301, opened 2025-01-13
feat: adding a Summary type in metrics package

ncostamagna:


peterbourgon:
Thanks for the thought and appreciate the PR, but eliding the Summary type is by design.

ncostamagna:
Got it, thanks!

